PR #15, opened 2018-08-17
fall back to mock data when API is not available

jharting:


karelhala:
Looking good! However I would change it to be used only when in dev mode, either with help of proxy from dev server or some env settings. But when bundling for live not include it there.

jharting:
@karelhala right now we do not have any production API so this change is primarily for the live version at https://access.redhat.com/insights/platform/inventory to show something. Once we have an actual API I can change that

PR #50, opened 2019-01-29
Remove inventory CI URL

dLobatog:
It's probably not a good idea to keep the CI inventory URL out in the open, so that it may not be abused


karelhala:
@dLobatog spandx URL shouldn't be there at all. Can you please remove it?

PR #59, opened 2019-02-19
Call refresh data when pagination changes

karelhala:
Is related to https://github.com/RedHatInsights/insights-frontend-components/pull/310


karelhala:
@jiridostal it is related to inventory component it is a bug and filters should not be there if no data are displayed.

PR #70, opened 2019-03-18
Update routes to cloud config

karelhala:
No visual changes.


src/entry.js
```diff
@@ -6,13 +6,11 @@ import { init } from './store';
...
 
-// exposes webpack variable RELEASE
-/*global RELEASE:true*/
-/*eslint no-undef: "error"*/
+const pathName = window.location.pathname.split('/');
 
 ReactDOM.render(
     <Provider store={init(logger).getStore()}>
-        <Router basename={ `/${RELEASE}/platform/inventory` }>
+        <Router basename={`${pathName[1] === 'beta' ? pathName[2] : pathName[1]}/inventory`}>
```
karelhala: @jharting this is how you can load inventory on multiple endpoints. It will use basename as either `rhcs/inventory` or `insights/inventory`.

PR #76, opened 2019-03-25
Improve build to release automatically to beta

karelhala:
* on `master` it will automatically release to `ci-beta` and `qa-beta`
* on `master-stable` it will automatically release to `ci-stable` and `qa-stable`
* on any other branch listed as build branch it will automatically push to this branch as a way how to promote specific env

### DIFF
--- a/.travis/custom_release.sh
+++ b/.travis/custom_release.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+set -x
+
+if [[ "${TRAVIS_BRANCH}" = "master" ]]; then
+    .travis/release.sh "ci-beta"
+    .travis/release.sh "qa-beta"
+    
+elif [[ "${TRAVIS_BRANCH}" = "master-stable" ]]; then
+    .travis/release.sh "ci-stable"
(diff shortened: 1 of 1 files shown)


.travis/custom_release.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [[ "${TRAVIS_BRANCH}" = "master" ]]; then
+    .travis/release.sh "ci-beta"
+    .travis/release.sh "qa-beta"
+    
+elif [[ "${TRAVIS_BRANCH}" = "master-stable" ]]; then
+    .travis/release.sh "ci-stable"
+    .travis/release.sh "qa-stable"
+
+# If current dev branch is deployment branch, push to build repo
```
iphands: Shouldnt ci/qa be removed from this conditional.
It might be confusing as hell to have two branches potentially pushing to these envs.

I would consider removing the whole stanza.

karelhala: I wanted to allow specific pushes, to ci and qa. If that would be needed. But I can remove that if it would cause confusion.

PR #96, opened 2019-05-08
New FEC version to adapt compliance to FF < 60

dLobatog:
This includes https://github.com/RedHatInsights/insights-frontend-components/pull/427


karelhala:
Thanks for openening this issue let's close it because we moved to use the new monorepo and this is already included in there.

PR #98, opened 2019-05-24
Clean dependencies and use new package to build repo

karelhala:
This uses new package for common build config. Notice how most of webpack configs are gone and you just have to pass root folder of your app so we know where these files are located.


ryelo:
This is awesome! We should update the frontend-starter-app to do something similar.

package.json
```diff
@@ -50,60 +56,31 @@
...
     "stylelint-config-recommended-scss": "^3.2.0",
     "stylelint-scss": "^3.4.0",
     "webpack": "^4.25.1",
     "webpack-bundle-analyzer": "^3.3.2",
     "webpack-cli": "^3.1.2",
-    "webpack-hot-client": "^4.1.1",
-    "webpack-serve": "1.0.2",
-    "webpack-visualizer-plugin": "^0.1.11",
-    "whatwg-fetch": "^3.0.0",
-    "write-file-webpack-plugin": "^4.4.1"
```
AllenBW: i believe we need the "webpack-dev-server": "3.4.1" dep here

karelhala: This dependency comes from https://github.com/RedHatInsights/frontend-components/blob/master/packages/config/package.json#L37

PR #123, opened 2019-08-13
Vulnerability exports different store and stylesheet

jiridostal:


package.json
```diff
@@ -11,7 +11,7 @@
     "@redhat-cloud-services/frontend-components-inventory-compliance": "0.0.22",
     "@redhat-cloud-services/frontend-components-inventory-general-info": "0.0.9",
     "@redhat-cloud-services/frontend-components-inventory-insights": "0.1.12",
-    "@redhat-cloud-services/frontend-components-inventory-vulnerabilities": "0.2.7",
+    "@redhat-cloud-services/frontend-components-inventory-vulnerabilities": "1.12.1",
```
karelhala: Can you update it to latest `1.13.2` so we are using the newest version?

jiridostal: Updated!

PR #356, opened 2020-03-19
Apply stale filter from URL

karelhala:


src/actions.js
```diff
@@ -17,6 +17,14 @@ export const deleteEntity = (systems, displayName) => ({
     }
 });
 
+export const setFilter = (filterValue, filterKey) => ({
+    type: SET_INVENTORY_FILER,
```
rvsia: Filer> filter

karelhala: Nice catch!

src/Routes.js
```diff
@@ -35,6 +36,7 @@ function checkPaths(technology, app) {
 
 export const Routes = ({ childProps: { history } }) => {
     const pathName = window.location.pathname.split('/');
+    const { status } = queryString.parse(location.search);
```
Hyperkid123: did you consider using brosers `URLSearchParams` instead of the library?

karelhala: Good point! This library looks slightly easier to use, but better to use browser build in features than custom libraries.

PR #413, opened 2020-04-08
Whats new is old, lets do #80 again, rename Insights to Advisor

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-5794

oof, reorders deps, missed that


AllenBW:
Ahhh i remember #80

PR #417, opened 2020-04-09
Update prod-beta to newest version

karelhala:


AllenBW:
oughta bring in #419 if yah can 🤞

PR #421, opened 2020-04-13
Update inventory-compliance to latest

dLobatog:
Currently we have updated all of our envs (:warning: *except prod-stable*) to be usable with the version of the API post 0.0.33.  Can you cherry-pick this everywhere but prod-stable or should I submit separate PRs?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2274,33 +2274,19 @@
       }
     },
     "@redhat-cloud-services/frontend-components-inventory-compliance": {
-      "version": "0.0.33",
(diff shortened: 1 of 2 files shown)


karelhala:
@dLobatog there's also https://github.com/RedHatInsights/insights-inventory-frontend/pull/410 from dependabot, it updates the version to `1.0.5` which one would be prefered?

dLobatog:
@karelhala I would say 1.0.5 is preferred - at this point just go with latest, everywhere but QA-stable and prod-stable.

PR #813, opened 2020-10-07
Limit table actions for users with permissions

rvsia:
part of https://issues.redhat.com/browse/RHCLOUD-9754

**Before**

image

**After**

image

### DIFF
--- a/src/routes/InventoryTable.js
+++ b/src/routes/InventoryTable.js
@@ -14,6 +14,7 @@ import { useStore } from 'react-redux';
 import DeleteModal from '../components/DeleteModal';
 import TextInputModal from '@redhat-cloud-services/frontend-components-inventory-general-info/TextInputModal';
 import flatMap from 'lodash/flatMap';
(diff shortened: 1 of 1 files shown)


src/routes/InventoryTable.js
```diff
@@ -172,29 +181,29 @@ const Inventory = ({
...
-                                                    displayName
-                                                }));
-                                                handleModalToggle(() => true);
-                                            }
-                                        }, {
-                                            title: 'Edit',
-                                            onClick: (_event, _index, data) => {
-                                                activateSystem(() => data);
-                                                onEditOpen(() => true);
+                                    hasCheckbox={false}
```
karelhala, line 186: It would probably be more straight forward to do `hasCheckbox={canPerformActions}` which would indicate that if user has access to actions show the checkbox.

karelhala (approved):
Looking good! Just one minor nitpick.

karelhala (approved):

PR #817, opened 2020-10-07
Move permissions to redux

rvsia:
closes https://issues.redhat.com/browse/RHCLOUD-9754

**Description**

If not on prod, the `PermissionLoader` component will load up `writePermissions` to redux. `InventoryTable` and `InventoryDetail` are not shown until the permissions are loaded (to prevent pop-in). Also, `writePermissions` are passed to `InventoryDetail` > `SystemCard` to also prevent pop-in and to not load permission multiple times.

**TODO**
- [x] test
- [x] generic-info update https://github.com/RedHatInsights/frontend-components/pull/769

### DIFF
--- a/src/components/PermissionsLoader.js
+++ b/src/components/PermissionsLoader.js
@@ -1,0 +1,24 @@
+import { useEffect } from 'react';
+import { useDispatch } from 'react-redux';
+
+import useInventoryWritePermissions from '../hooks/useInventoryWritePermissions';
+import { ACTION_TYPES } from '../constants';
+
+const PermissionLoader = () => {
+    const { isLoading, hasAccess } = useInventoryWritePermissions();
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        dispatch({ type: `${ACTION_TYPES.LOAD_WRITE_PERMISSIONS}_PENDING` });
+    }, []);
+
+    useEffect(() => {
+        if (!isLoading) {
+            dispatch({ type: `${ACTION_TYPES.LOAD_WRITE_PERMISSIONS}_FULFILLED`, payload: { writePermissions: hasAccess } });
+        }
+    }, [isLoading]);
+
+    return null;
+};
+
+export default PermissionLoader;
--- a/src/components/inventory/GeneralInfo.js
+++ b/src/components/inventory/GeneralInfo.js
@@ -1,9 +1,13 @@
 import React from 'react';
-import { useStore } from 'react-redux';
+import { useStore, useSelector } from 'react-redux';
 import GeneralInformation from '@redhat-cloud-services/frontend-components-inventory-general-info/cjs';
 
 const GeneralInformationTab = () => {
-    return <GeneralInformation store={useStore()} />;
+    const writePermissions  = useSelector(
+        ({ permissionsReducer: { writePermissions } }) => writePermissions
+    );
+
+    return <GeneralInformation store={useStore()} writePermissions={writePermissions} />;
 };
 
 export default GeneralInformationTab;
--- a/src/components/inventory/GeneralInfo.js
+++ b/src/components/inventory/GeneralInfo.js
@@ -3,8 +3,8 @@ import { useStore, useSelector } from 'react-redux';
 import GeneralInformation from '@redhat-cloud-services/frontend-components-inventory-general-info/cjs';
 
 const GeneralInformationTab = () => {
-    const writePermissions  = useSelector(
-        ({ permissionsReducer: { writePermissions } }) => writePermissions
+    const writePermissions = useSelector(
+        ({ permissionsReducer }) => permissionsReducer?.writePermissions
     );
 
     return <GeneralInformation store={useStore()} writePermissions={writePermissions} />;
--- a/src/routes/InventoryDetail.js
+++ b/src/routes/InventoryDetail.js
@@ -24,7 +24,8 @@ const Inventory = ({ entity, currentApp, clearNotifications, loadEntity }) => {
     const store = useStore();
     const { InventoryDetail, AppInfo, DetailWrapper } = ConnectedInventory;
     const { loading, writePermissions } = useSelector(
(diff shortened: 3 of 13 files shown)


src/components/PermissionsLoader.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from 'react';
+import { useDispatch } from 'react-redux';
+
+import useInventoryWritePermissions from '../hooks/useInventoryWritePermissions';
+import { ACTION_TYPES } from '../constants';
+
+const PermissionLoader = () => {
```
karelhala, line 7: Interesting way of using hook component to mutate redux!

src/components/inventory/GeneralInfo.js
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
-import { useStore } from 'react-redux';
+import { useStore, useSelector } from 'react-redux';
 import GeneralInformation from '@redhat-cloud-services/frontend-components-inventory-general-info/cjs';
 
 const GeneralInformationTab = () => {
-    return <GeneralInformation store={useStore()} />;
+    const writePermissions  = useSelector(
+        ({ permissionsReducer: { writePermissions } }) => writePermissions
+    );
```
karelhala, line 8: It might be better to use safe accessor here
```suggestion
  const writePermissions = useSelector(
    ({ permissionsReducer }) => permissionsReducer?.writePermissions
  );
```

src/routes/InventoryDetail.js
```diff
@@ -16,15 +16,17 @@ import '@redhat-cloud-services/frontend-components-inventory-insights/index.css'
...
 
 const Inventory = ({ entity, currentApp, clearNotifications, loadEntity }) => {
     const [ConnectedInventory, setInventory] = useState({});
     const store = useStore();
     const { InventoryDetail, AppInfo, DetailWrapper } = ConnectedInventory;
-    const canPerformActions = useInventoryWritePermissions();
+    const { loading, writePermissions } = useSelector(
+        ({ permissionsReducer: { loading, writePermissions } }) => ({ loading, writePermissions }),
+        shallowEqual
+    );
```
karelhala, line 29: Same as above
```suggestion
    const { loading, writePermissions } = useSelector(
        ({ permissionsReducer }) => ({ permissionsReducer?.loading, permissionsReducer?.writePermissions }),
        shallowEqual
    );
```

src/routes/InventoryTable.js
```diff
@@ -85,7 +84,10 @@ const Inventory = ({
     const [filters, onSetfilters] = useState([]);
     const [ediOpen, onEditOpen] = useState(false);
     const [globalFilter, setGlobalFilter] = useState();
-    const canPerformActions = useInventoryWritePermissions();
+    const { loading, writePermissions } = useSelector(
+        ({ permissionsReducer: { loading, writePermissions } }) => ({ loading, writePermissions }),
+        shallowEqual
+    );
```
karelhala, line 90: And here
```suggestion
    const { loading, writePermissions } = useSelector(
        ({ permissionsReducer }) => ({ permissionsReducer?.loading, permissionsReducer?.writePermissions }),
        shallowEqual
    );
```

src/store/permissions/reducer.js
```diff
@@ -0,0 +1,39 @@
+import { applyReducerHash } from '@redhat-cloud-services/frontend-components-utilities/files/ReducerRegistry';
+import { ACTION_TYPES } from '../../constants';
+
+const loadWritePermissionsPending = (state) => ({
+    ...state,
+    loadingFailed: false,
+    loading: !insights.chrome.isProd,
```
karelhala, line 7: Shouldn't this be always true?

src/store/permissions/reducer.js
```diff
@@ -0,0 +1,39 @@
...
+    writePermissions: insights.chrome.isProd
```
karelhala, line 8: I would set this to `undefined` while loading.

src/store/permissions/reducer.js
```diff
@@ -0,0 +1,39 @@
...
+
+const loadWritePermissionsFailed = (state) => ({
+    ...state,
+    loading: false,
+    loadingFailed: true
+});
+
+const defaultPermissionState = {
+    loading: !insights.chrome.isProd,
+    writePermissions: insights.chrome.isProd,
```
karelhala, line 26: I would change these default values to
```suggestion
    loading: false,
    writePermissions: undefined,
```

karelhala (commented):
Just a few nitpicks, but this is looking hella sweet!

src/store/permissions/reducer.js
rvsia: This is just a safe check. Pending should not never happen when users are on prod anyway.

src/store/permissions/reducer.js
rvsia: Actually no, because when you are on prod, there is no loading happening (because we don't have to do it - it would be just wasting of users's resources - on prod they have always write permissions now)

src/store/permissions/reducer.js
rvsia: ^ basically the same answer

karelhala (approved):

karelhala:
Let's not forget to remove all those `isProd` on Oct 16 when RBAC goes live in inventory :smile:

src/components/PermissionsLoader.js
rvsia: Yeah normally I would add this code to some top level component as a mount event, but I didn't want to duplicate the `usePermissions` hook API code.

And actually like this approach, as it's really just a 'component plugin' :smile: I am using it a lot in Sources UI.

rvsia:
@karelhala Sorry, I had to revert your commit suggestions as we missed to put keys in a object :sweat_smile: 

`() => ({ permissions?.loading }}`
>
`() => ({ loading: permissions?.loading }}`

PR #845, opened 2020-10-19
Enable RBAC everywhere

karelhala:
Since we are enabling RBAC for profuction no need to have RBAC gated behind `isProd` flag from chrome.

### DIFF
--- a/src/store/permissions/reducer.test.js
+++ b/src/store/permissions/reducer.test.js
@@ -11,16 +11,6 @@ describe('permissionsReducer', () => {
         });
     });
 
-    it('pending write permission - on prod', () => {
-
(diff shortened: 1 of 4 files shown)


src/store/permissions/reducer.test.js
```diff
@@ -12,17 +12,13 @@ describe('permissionsReducer', () => {
...
 
         expect(permissionsReducer({}, { type: `${ACTION_TYPES.LOAD_WRITE_PERMISSIONS}_PENDING` }))
         .toEqual({
             loading: false,
             loadingFailed: false,
             writePermissions: true
         });
-
-        insights.chrome.isProd = tmp;
     });
```
rvsia, line 22: I think this whole test can be :skull:

rvsia (approved):

PR #860, opened 2020-10-23
Enable delete button if user has execute rights

karelhala:


src/routes/InventoryTable.js
```diff
@@ -165,6 +165,8 @@ const Inventory = ({
 
     const calculateSelected = () => selected ? selected.size : 0;
 
+    console.log(writePermissions, 'WTF!');
```
rvsia, line 168: console.log

rvsia (approved):

PR #916, opened 2020-12-07
Update to webpack 5

karelhala:
### Update to WP5 and use new inv component

Since we want to move to SPA we have to update each app to start using webpack 5 and scalprum component for inventory. This PR does this thing and also changes how inventory table is tested. Some dependencies for details are beta because they were bloating the build. They should be fine in stable as all of them released stable version, but for the initial draft let's go with the beta versions and update once we prove that nothing breaks.


src/AppEntry.js
```diff
@@ -0,0 +1,29 @@
...
+InventoryApp.propTypes = {
+    useLogger: PropTypes.bool
+};
+
+InventoryApp.defaultProps = {
+    useLogger: false
+};
+
+export default InventoryApp;
+
```
rvsia, line 29: empty line

src/AppEntry.js
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+import { BrowserRouter as Router } from 'react-router-dom';
+import { Provider } from 'react-redux';
+import { init } from './store';
+import App from './App';
+import { getBaseName } from '@redhat-cloud-services/frontend-components-utilities/files/esm/helpers';
+import logger from 'redux-logger';
+
+const InventoryApp = ({ useLogger }) => {
+    return (
```
rvsia, line 11: we can omit return statement

src/components/inventory/Vulnerability.js
```diff
@@ -1,10 +1,22 @@
...
+        setComponent(() => React.lazy((async () => {
+            const Vulnerabilities = await import('@redhat-cloud-services/frontend-components-inventory-vulnerabilities/dist/esm');
+            const { SystemCvesStore } = Vulnerabilities;
+            getRegistry().register({ SystemCvesStore });
+            return Vulnerabilities;
+        })));
+    }, []);
+
+    return <Suspense>
+        <Component customItnlProvider entity={ { id: useParams().inventoryId } }/>
```
rvsia, line 18: Won't Fragment be angry that he gets some unknown props?

rvsia (approved):

PR #1080, opened 2021-02-24
Fix registration of reducers

karelhala:


src/Routes.js
```diff
@@ -21,6 +22,8 @@ function checkPaths(technology, app) {
 }
 
 export const Routes = () => {
+    const { getRegistry } = useContext(RegistryContext);
+    getRegistry().register(reducers);
```
rvsia, line 26: Couldn't this possibly register reducers on each render?

karelhala: Good point, I've wrapped it in `useEffect` that checks for `getRegistry` as it is mutated once app fully renders.

rvsia (approved):

PR #1096, opened 2021-03-05
[DO not merge] Async component example.

Hyperkid123:
This is an example PR how to shared additional federated modules.

### Requirements
- define module (component) that should be shared
- expose the module via `@redhat-cloud-services/frontend-components-config/federated-modules` or via `ModuleFederationPlugin`
- If not using the shared config, also make sure that the module is present in the `fed-mods.json` build file

I've also added additional updates to properly serve files in beta env


karelhala:
@Hyperkid123 let's close this one, so we don't accidentally merge it.

PR #1116, opened 2021-03-09
Hotifx build

Hyperkid123:
The patchman dependencies were causing issues with the html webpack plugin. I had to add a direct dependency, dedupe dependencies, and alias to fix the HTML build.

Long term fix is here: https://github.com/RedHatInsights/patchman-ui/pull/457
We also need a new version of patch inventory to be installed


karelhala (approved):

Hyperkid123:
@karelhala just testing clean install if it works. If not we will have to add dedupe before the build command

rvsia (approved):

PR #1261, opened 2021-05-10
Add start:proxy task

karelhala:


README.md
```diff
@@ -24,6 +24,14 @@ SPANDX_CONFIG="$(pwd)/insights-inventory-frontend/config/spandx.config.js" bash
 2. ```npm run start```
     - starts webpack bundler and serves the files with webpack dev server
 
+### Running the app without insights-proxy
+
+You don't have to run insights proxy to serve your application, you can use built in `webpack-proxy` by running
+```
+npm run start:proxy
```
rvsia, line 31: I think you are missing the command in package.json, also there should be both variants for beta and non betas.

karelhala: Good point! Instead of adding yet another script with process env I've added section to readme with description how to run each start/build command with beta

rvsia (approved):

PR #1262, opened 2021-05-10
Migrate advisor tab to use fed modules

karelhala:


AllenBW (approved):
🧼 👩‍🎨 🎉 

woah, thats it? NEAT!

rvsia (approved):
@karelhala Do we any sync inv-app process? This can be easily broken if the inventory gets ahead. 😿  Maybe if the fed module fail, then load the old component...

karelhala:
@rvsia I had similiar thought but that would be a hacky way around this. @AllenBW already told me that the advisor UI has shipped the code on their side all the way to prod-stable. So no need to worry about it.

PR #1266, opened 2021-05-10
Migrate general info tab to federated modules

karelhala:


rvsia (approved):
🙏

rvsia:
@karelhala  The tests failed because of different timestamps. However, there are also many useLayoutEffect warnings. I think we should get rid of them. The log is gigantic.

PR #1275, opened 2021-05-19
ros tab to use fed modules

kgaikwad:
ROS side PR - https://github.com/RedHatInsights/ros-frontend/pull/40

+ adding @karelhala into loop for review.

### DIFF
--- a/src/components/inventory/Ros.js
+++ b/src/components/inventory/Ros.js
@@ -1,0 +1,21 @@
+import React, { useContext } from 'react';
+import { useRouteMatch } from 'react-router-dom';
+import { RegistryContext } from '../../store';
+import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
+import fallback from '../SpinnerFallback';
+
+const RosTab = () => {
+    const { params } = useRouteMatch('/:inventoryId');
+    const { getRegistry } = useContext(RegistryContext);
+
+    return <AsyncComponent
+        appName="ros"
+        module="./SystemDetail"
+        getRegistry={getRegistry}
+        fallback={fallback}
+        customItnlProvider
+        entity={ { id: params.inventoryId } }
+    />;
+};
+
+export default RosTab;
--- a/src/components/inventory/Ros.js
+++ b/src/components/inventory/Ros.js
@@ -11,10 +11,9 @@ const RosTab = () => {
     return <AsyncComponent
         appName="ros"
         module="./SystemDetail"
-        getRegistry={getRegistry}
+        getRegistry={ getRegistry }
         fallback={fallback}
-        customItnlProvider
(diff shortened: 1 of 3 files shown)


src/store/reducers.js
```diff
@@ -52,7 +53,13 @@ function entityLoaded(state, { payload: { entitlements } } = { payload: {} }) {
                 title: 'Patch',
                 name: 'patch',
                 component: PatchTab
+            },
+            isEntitled(entitlements && entitlements.insights) && {
```
karelhala, line 57: Is the `ROS` enabled for all users? If not we might want to add the cookie check as well here.

src/components/inventory/Ros.js
```diff
@@ -0,0 +1,21 @@
...
+    const { params } = useRouteMatch('/:inventoryId');
+    const { getRegistry } = useContext(RegistryContext);
+
+    return <AsyncComponent
+        appName="ros"
+        module="./SystemDetail"
+        getRegistry={getRegistry}
+        fallback={fallback}
+        customItnlProvider
+        entity={ { id: params.inventoryId } }
```
karelhala, line 17: If you are going to use the solution with `useSelector` you can change the `entity` prop to `inventoryId={ params.inventoryId }`

Also no need for `customItnlProvider` as ROS is not using currently any intl service, right?

src/store/reducers.js
kgaikwad: Ros is not enabled for all users. Added cookie check.

src/components/inventory/Ros.js
kgaikwad: Done!

src/store/reducers.js
```diff
@@ -32,6 +33,10 @@ function entitiesLoaded(state, { payload }) {
 }
 
 function entityLoaded(state, { payload: { entitlements } } = { payload: {} }) {
+
+    const hasRosCookie = insights.chrome.visibilityFunctions.isProd() ?
+        insights.chrome.visibilityFunctions.hasCookie('cs_ros_beta_enable') : true;
```
karelhala, line 38: This one should also pass in the value
```suggestion
        insights.chrome.visibilityFunctions.hasCookie('cs_ros_beta_enable', '1') : true;
```

kgaikwad: Thank you! Good catch :+1:

karelhala (approved):

src/store/reducers.js
kgaikwad: @karelhala,
I forgot to mention one thing earlier that ROS is not enabled by default like few other services and designed for RHEL instances from cloud (AWS/azure). 
Along with cookie check, do I need to add extra conditions here? Or it is ok to show Tab with empty recommendations table.

karelhala (approved):

PR #1326, opened 2021-06-30
Add inventory component

karelhala:


karelhala:
@rvsia yup, pretty much just moving files over and making sure everything works as it should.

PR #1333, opened 2021-07-12
remove cookie check for ROS tab

kgaikwad:
With [JIRA card](https://issues.redhat.com/browse/RHIROS-201), removing ros specific cookie check from respective repositories.

+adding @karelhala into loop.

### DIFF
--- a/src/store/reducers.js
+++ b/src/store/reducers.js
@@ -29,7 +29,6 @@ function entitiesLoaded(state, { payload }) {
 }
 
 function entityLoaded(state) {
-
     return {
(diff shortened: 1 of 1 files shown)


src/store/reducers.js
```diff
@@ -30,9 +30,6 @@ function entitiesLoaded(state, { payload }) {
 
 function entityLoaded(state) {
```
rvsia, line 32: Can we remove the empty line?

rvsia (approved):

PR #1336, opened 2021-07-14
Update prod beta

rvsia:


rvsia:
@Hyperkid123 I tried it with proxy - the application is working fine, but there is still the same error for ROS tab. I guess proxy does not work properly when messing up with federated modules. Or if not, could you please take a look?

Hyperkid123:
@rvsia I will take a look.

Hyperkid123:
@rvsia the reason why it's not working in prod beta is because the ROS app is not up to date and it does not expose the correct module:
- prod-beta: https://github.com/RedHatInsights/ros-frontend/blob/prod-beta/config/prod.webpack.config.js
- ci-beta: https://github.com/RedHatInsights/ros-frontend/blob/main/config/prod.webpack.config.js

The `./SystemsDetail` does not exist in prod beta, therefore it cannot be rendered. Simple. Ros has to update in prod beta in order to work.

rvsia:
@Hyperkid123 Oh, I was confused by that the Inv Frontend also exposes .SystemDetail page. :D So, we should update ROS first and not to update this.

kgaikwad:
> @rvsia the reason why it's not working in prod beta is because the ROS app is not up to date and it does not expose the correct module:
> 
>     * prod-beta: https://github.com/RedHatInsights/ros-frontend/blob/prod-beta/config/prod.webpack.config.js
> 
>     * ci-beta: https://github.com/RedHatInsights/ros-frontend/blob/main/config/prod.webpack.config.js
> 
> 
> The `./SystemsDetail` does not exist in prod beta, therefore it cannot be rendered. Simple. Ros has to update in prod beta in order to work.

Thank you @Hyperkid123 for pointing this. 
My mistake! I didn't notice and realize that webpack config is missing for prod env.
I will make the appropriate changes in ros-frontend.

karelhala:
@rvsia the ros PR has been merged, let's merge this PR

PR #1354, opened 2021-08-03
Show ROS tab only on beta environments

karelhala:


rvsia (approved):

src/store/reducers.js
```diff
@@ -50,7 +50,7 @@ function entityLoaded(state) {
                 name: 'patch',
                 component: PatchTab
             },
-            {
+            insights?.chrome?.isBeta() && {
```
kgaikwad, line 53:
```suggestion
(!insights?.chrome?.isProd() || insights?.chrome?.isProd() && insights?.chrome?.isBeta()) && {
```


@karelhala - ROS service is not yet in GA but ROS tab is visible under  inventory detail page in public Prod. 
ROS should be visible everywhere except in public prod (i.e under GA)

karelhala: Sure thing!

kgaikwad (approved):
LGTM 👌👍

kgaikwad:
Thank you @karelhala for taking care of this!

PR #1388, opened 2021-08-31
PoC for inventory frontend's new build pipeline

BlakeHolifield:
* Adds a `pr_check.sh` to run with each commit to an open PR
* Adds a `build_deploy.sh` to run with each new commit to the main / master branch
* Adds a `Dockerfile` to mount the nginx config

Note: This won't work yet and will need entries for its `build.yml` in app-interface. Those will come after some validation from the plat-ex team.

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -exv
+
+IMAGE="quay.io/cloudservices/insights-inventory-frontend"
+IMAGE_TAG=$(git rev-parse --short=7 HEAD)
+
+if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" ]]; then
+    echo "QUAY_USER and QUAY_TOKEN must be set"
+    exit 1
+fi
+
+
+AUTH_CONF_DIR="$(pwd)/.podman"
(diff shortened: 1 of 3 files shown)


build_deploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -exv
+
+IMAGE="quay.io/cloudservices/insights-inventory-frontend"
```
karelhala, line 5: You can read the app name from `package.json#insights.appname` this way we could come up with a clever idea to generate this on the fly and apps won't have to change anything in this file.

pr_check.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+export APP_NAME="host-inventory" # name of app-sre "application" folder this component lives in
+export IMAGE="quay.io/cloudservices/insights-inventory-frontend"
```
karelhala, line 7: Same as above, we can read the name from the package.

karelhala (commented):
Looking good!

karelhala (approved):

PR #1391, opened 2021-09-10
Use only the latest request to reload state

rvsia:
When changing filters during loading, filters are not being updated - because useEffect hooks are disabled when `loaded` is false.

I removed this condition and make a new condition in reducer so only the latest request will rewrite the store. This means that there can be multiple requests at the same time, but only the latest will change the state.

**Before** (notice the different total)

Kapture 2021-09-10 at 11 01 03

**After**

Kapture 2021-09-10 at 10 59 50

cc @leSamo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "insights-inventory-frontend",
       "version": "0.0.1",
       "dependencies": {
         "@patternfly/react-core": "4.135.0",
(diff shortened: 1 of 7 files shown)


leSamo:
When I try to apply filter while another filter is being applied, it looks like the first filter is applied, then second, but table is not cleared between that and I have 20 items in the table even if my per page is only 10.
inventory

karelhala (approved):

rvsia:
@leSamo good catch! I believe that the reducer is merging items.. so I guess I need to clear them, not just return a current state.

rvsia:
@leSamo @karelhala There is also an enhancer for the reducer that mutates the state when entities are loaded - so I added the check there too and it seems working ok to me. Can you please double check it, @leSamo ?

PR #1405, opened 2021-10-05
Fixed reset filter

arivepr:
Addresses issue with  'Clear filters' deleting default filters from table along with user-added filters.

Screen Shot 2021-10-05 at 4 51 42 PM

_after hitting Reset filters.._

Screen Shot 2021-10-05 at 4 52 00 PM


arivepr:
Don't have app sre permissions, can't see why that build is failing.

Hyperkid123:
> **Don't have app sre permissions, can't see why the "ci.ext.devshift.net" build is failing.**

hmm great this will break the deployment...

Hyperkid123:
It is failing because some file is missing
```
21:50:04 + ./pr_check.sh
21:50:04 /tmp/jenkins4518323946609197312.sh: line 2: ./pr_check.sh: No such file or directory
```

PR #1416, opened 2021-10-19
feat(InventoryTable): RHICOMPL-2377 - Add and call getTags for custom tags list

bastilian:
A `getTags` prop would allow apps to provide a function to be called in order to query a custom API providing the list of tags.

(See https://github.com/RedHatInsights/compliance-frontend/pull/1415/files#diff-d41389c1e49b96059af8df5c978c229a8cc521008bd0ac86a8f71a43ae620473R203 and below for an example)

### DIFF
--- a/doc/custom_fetch.md
+++ b/doc/custom_fetch.md
@@ -7,16 +7,26 @@
       - [results](#results)
       - [total](#total)
       - [loaded](#loaded)
-  - [Example](#example)
+    - [Example](#example)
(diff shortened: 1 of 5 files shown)


karelhala (approved):
Looking good, can you update [doc/custom_fetch.md](https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/doc/custom_fetch.md) as well so others know about it?

kruai (approved):

PR #1432, opened 2021-11-02
Update dependencies

rvsia:
Getting `TypeError: Cannot read properties of undefined (reading 'init')`

Screenshot 2021-11-02 at 15 47 01

When running locally cc @Hyperkid123 @karelhala is that something changed in scalprum? 🤔 

**EDIT: caused by NPM link**

**EDIT2: unlinked and it still does not work 🤔 **

**EDIT3: wrong path, I was trying to use ansible instead of isnights**


karelhala (approved):

karelhala:
@rvsia looks like lock file is out of sync with package.json

PR #1433, opened 2021-11-03
Add frontend template

bsquizz:


kruai (approved):
If we need this to unblock the HBI API, I'm all for it

PR #1435, opened 2021-11-04
Resource tab optimization

arivepr:
**At long last**

Aims to hide the tab for whatever system is **not** on azure or aws. We look at the 'cloud_provider' field for confirmation.

Screen Shot 2021-11-04 at 2 21 55 PM
(no cloud_provider field provided)

Screen Shot 2021-11-04 at 2 04 41 PM
(cloud_provider field set to 'aws')


src/components/InventoryDetail/InventoryDetail.js
```diff
@@ -41,7 +42,16 @@ const InventoryDetail = ({
...
+            const action = loadEntity(currId, { hasItems: true }, { showTags });
+            dispatch(action);
+            action.payload.then(() => {
+                getEntitySystemProfile(currId).then(result => {
+                    const cloudProviderObj = (typeof result.results[0].system_profile.cloud_provider !== 'undefined')
+                        && result.results[0].system_profile.cloud_provider;
+                    dispatch(setRosTabVisibility((cloudProviderObj === 'aws' || cloudProviderObj === 'azure')));
+                    return result;
+                });
+            });
```
karelhala, line 54: I'd change the flow of action completely, instead of fetching the data and dispatching specific action for ROS, I'd disable the ROS tab by default. And I'd add new reducer that reacts to `ACTION_TYPES.LOAD_SYSTEM_PROFILE` and enables the ROS tab. `resourceOptTabVisibility` should be enough. This way we won't add additional logic to the shared InventoryDetail and at the same time make all custom tab hiding just for inventory app.

Approach in this PR (waiting for `getEntitySystemProfile`) could mean that ROS tab would not be visible even in ROS application itself, which is unwanted.

src/store/inventory-actions.js
```diff
@@ -193,6 +194,11 @@ export const deleteEntity = (systems, displayName) => ({
     }
 });
 
+export const setRosTabVisibility = (shouldShowRosTab) => ({
```
karelhala, line 197: I don't think you need this action, it should be enough to wait for the `ACTION_TYPES.LOAD_SYSTEM_PROFILE`. If such action is not fired when user is on `ROS` tab by default, we should fire the `systemProfile` action directly when `InventoryDetail` route is rendered.

src/store/reducers.js
```diff
@@ -105,11 +124,7 @@ function entityDeleted(state, { meta }) {
     };
 }
 
-function onEntitiesLoaded(state, { payload, meta }) {
-    if (meta?.lastDateRequest < state?.lastDateRequest) {
-        return state;
-    }
```
karelhala, line 111: @rvsia this was added to prevent race conditions, right? I don't think we can remove this.

src/store/reducers.js
```diff
@@ -160,7 +175,8 @@ export const tableReducer = applyReducerHash(
 
 export const entitesDetailReducer = () => applyReducerHash(
     {
-        [INVENTORY_ACTION_TYPES.LOAD_ENTITY_FULFILLED]: entityLoaded
+        [INVENTORY_ACTION_TYPES.LOAD_ENTITY_FULFILLED]: entityLoaded,
+        [SET_ROS_TAB_VISBILITY]: resourceOptTabVisibility
```
karelhala, line 179: You can use `INVENTORY_ACTION_TYPES.LOAD_SYSTEM_PROFILE`
```suggestion
        [INVENTORY_ACTION_TYPES.LOAD_SYSTEM_PROFILE]: resourceOptTabVisibility
```

src/components/InventoryDetail/InventoryDetail.js
Hyperkid123: > I'd disable the ROS tab by default.

@karelhala  That is what happens.

> And I'd add a new reducer that reacts to ACTION_TYPES.LOAD_SYSTEM_PROFILE and enables the ROS tab.
Using the action type `ACTION_TYPES.LOAD_SYSTEM_PROFILE` can result in more race conditions when switching between detail pages :disappointed:. Ideally, we would use the `thunk` middleware to chain the actions directly through redux though. Optimal way would be if the data was provided in the entity detail but that is not happening.

karelhala: The ROS tab should be disabled only on full inventory view. We can wait for the `ACTION_TYPES.LOAD_SYSTEM_PROFILE` (no need to fire a new one as it should already be fired) to load and enable the ROS tab in inventory app redux store and enable the tab if criterium is met. The problem I have with the approach in this PR is that we're adding extra logic to shared component just to disable the tab in inventory app.

Hyperkid123: In that case, shouldn't his be handled by the ROS tab itself then?

karelhala: It's partially handleded by the ROS tab, they currently show empty state. But they don't want to show the tab at all for some users (I was against it, but the PM specifically required it)

Hyperkid123: Well, I would personally move the responsibility to the team. As I said, provide a component API to specifically state which tabs to show. They will have full control and we have one thing less to worry about. I assume we will get requests for other tabs with similar behaviour in the future. I think it may come in handy when you navigate through the URL. This can cause some wire behaviour when you navigate directly through the URL, and your tab is missing for some reason and pops up later.

arivepr:
Test still failing, not sure how to address them. Functionality is working after these changes. Will squash commits once I know what else to change :P

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -14,11 +14,14 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
     const history = useHistory();
     const dispatch = useDispatch();
     const searchParams = new URLSearchParams(search);
-    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps);
+    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps.filter(({ isVisible }) => isVisible !== false));
```
karelhala, line 17: I would add another guard in here
```suggestion
    const items = useSelector(({ entityDetails }) => (entityDetails?.activeApps || []).filter(({ isVisible }) => isVisible !== false));
```

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -14,11 +14,14 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
...
     const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp);
+    const disabledApps = useSelector(({ systemProfileStore }) => systemProfileStore?.disabledApps);
     const defaultApp = activeApp?.appName || appList?.find(({ pageId, name }) => items?.[0]?.name === (
         pageId || name))?.name || items?.[0]?.name;
-    const applications = appList || items;
+    let applications = appList || items;
+    const [activeTabs, setActiveTabs] = useState(applications);
```
karelhala, line 23: This can be simplified to directly assign to default state
```suggestion
    const [activeTabs, setActiveTabs] = useState(appList || items);
```

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -29,6 +32,16 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
         }
     }, []);
 
+    useEffect(() => {
+        const filteredResult = disabledApps && disabledApps.length && applications.filter(app => app.name !== disabledApps[0]);
```
karelhala, line 36: This can be simplified with safe accessor, using `includes` and fallbacking to original value
```suggestion
        // filters out disabled apps, if no disabledApps is used it will use the full applications list
        const filteredResult = applications.filter(app => !disabledApps?.includes(app.name));
        setActiveTabs(filteredResult)
```

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -50,7 +63,7 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
                     isFilled
                     className="ins-c-inventory-detail__app-tabs"
                 >
-                    { applications.map((item, key) => (
+                    { activeTabs && activeTabs.length && activeTabs.map((item, key) => (
```
karelhala, line 66: This can be simplified with safe accessor
```suggestion
                    { activeTabs?.map((item, key) => (
```

src/store/action-types.js
```diff
@@ -65,3 +65,4 @@ export const CLEAR_FILTERS = 'CLEAR_FILTERS';
 export const TOGGLE_TAG_MODAL = 'TOGGLE_TAG_MODAL';
 export const CONFIG_CHANGED = 'CONFIG_CHANGED';
 export const TOGGLE_DRAWER = 'TOGGLE_INVENTORY_DRAWER';
+export const SET_ROS_TAB_VISBILITY = 'SET_ROS_TAB_VISBILITY';
```
karelhala, line 68: I don't think we need this const

src/store/inventory-actions.js
```diff
@@ -193,6 +193,11 @@ export const deleteEntity = (systems, displayName) => ({
     }
 });
 
+// export const setRosTabVisibility = (shouldShowRosTab) => ({
```
karelhala, line 196: No need for commented out code

src/store/systemProfileStore.js
```diff
@@ -45,8 +45,14 @@ export function calculateInterfaces(interfaces) {
 
 export function onSystemProfile(state, { payload: { results } }) {
     const systemProfile = (results && results[0] && results[0].system_profile) || {};
+    const cloudProviderObj = (results && results[0] && (typeof results[0].system_profile.cloud_provider !== 'undefined'))
+                        && results[0].system_profile.cloud_provider;
     return {
         ...state,
+        disabledApps: [
+            ...(cloudProviderObj === 'aws' || cloudProviderObj === 'test cloud provider') ? [] : ['ros']
+            // ...(cloudProviderObj === 'aws' || cloudProviderObj === 'azure') ? ['ros'] : []
```
karelhala, line 54: I think we actually want this code to present and the one above not, right?

karelhala (commented):
Looking hella good! Just a few minor adjustments and we are golden!

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -14,11 +14,15 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
     const history = useHistory();
     const dispatch = useDispatch();
     const searchParams = new URLSearchParams(search);
-    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps);
+    // eslint-disable-next-line max-len
+    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || []).filter(({ isVisible }) => isVisible !== false);
```
karelhala, line 18: Might be a better to wrap after `filter`
```suggestion
    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || [])
    .filter(({ isVisible }) => isVisible !== false);
```

Hyperkid123 (approved):

PR #1456, opened 2021-12-02
Early December Release

arivepr:
Pushing/Mergin for release, specifically for the fix on the resource optimization tab.


arivepr:
If title is a problem for being too vague, more than happy to change to any suggested names for this release pr :)

Hyperkid123:
@arivepr what about the rest of the envs? Do we only target prod beta?

arivepr:
@Hyperkid123 The feature that I need released is only available on beta, so the rest of the environments I leave to your prerogative!

(was told to open up pr against beta for now)

Hyperkid123 (approved):

PR #1466, opened 2022-01-13
Enable ROS tab in production.

Hyperkid123:
This PR can be merged once ROS is deployed to prod-stable env: https://github.com/RedHatInsights/cloud-services-config/pull/1012

cc @PreetiW

### DIFF
--- a/src/store/reducers.js
+++ b/src/store/reducers.js
@@ -67,7 +67,7 @@ function entityLoaded(state) {
                 name: 'patch',
                 component: PatchTab
             },
-            (!insights.chrome.isProd || (insights.chrome.isProd && insights?.chrome?.isBeta())) && {
+            {
                 title: 'Resource Optimization',
                 name: 'ros',
                 isVisible: false,


PreetiW (commented):
Thank you @Hyperkid123 for the PR and heads up 💯

src/store/reducers.js
```diff
@@ -67,7 +67,7 @@ function entityLoaded(state) {
                 name: 'patch',
                 component: PatchTab
             },
-            (!insights.chrome.isProd || (insights.chrome.isProd && insights?.chrome?.isBeta())) && {
+            {
                 title: 'Resource Optimization',
                 name: 'ros',
                 isVisible: false,
```
PreetiW, line 73: @Hyperkid123 just curious if we also need to remove `isVisible`? just observed that its not part of other services so got curious.

Hyperkid123: The visibility of the ROS tab is also based on additional inventory data. Some systems might not support the ROS features therefore the tab is hidden. And it is hidden by default because we first have to get some general system data from API. If the condition is met, the flag is flupped.

PreetiW: ohh make sense, I got it 💯 
Thanks for clarifying @Hyperkid123 👍🏻

karelhala (approved):

PR #1470, opened 2022-01-14
[WIP] Make inventory table sortable by OS

ryandillinfelton:
Resolves [this Jira](https://issues.redhat.com/browse/ESSNTL-1441) feature request.

Simply makes the table orderable/sortable by the OS column the same way that `hostname` and `updated` are now.

sortable_by_os_not_selected

Here is an example of the generated query

Right now this change is blocked on [ESSNTL-1879](https://issues.redhat.com/browse/ESSNTL-1879). There is a bug in the `/hosts/{host_id_list}` endpoint that makes it return a `400` when ordered by `operating_system`.

### DIFF
--- a/src/store/entities.js
+++ b/src/store/entities.js
@@ -26,6 +26,11 @@ export const defaultState = {
     tagsLoaded: false,
     allTagsLoaded: false,
     invConfig: {},
+    //TODO: change back
+    // sortBy: {
+    //     key: 'updated',
+    //     direction: 'desc'
+    // }
     sortBy: {
         key: 'updated',
         direction: 'desc'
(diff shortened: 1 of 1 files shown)


src/api/api.js
```diff
@@ -70,6 +70,17 @@ export const filtersReducer = (acc, filter = {}) => ({
...
+function processOrderBy(orderBy) {
+    //TODO: remove logging
+    console.log("porcessing orderBy:\n orderBy: "+orderBy);
+    if (orderBy === "system_profile") {
+        orderBy = "operating_system";
+    }
+
+
+    return orderBy;
+}
```
ryandillinfelton, line 82: I'm not a fan of this solution because it limits us down the line from being able to have more `system_profile` columns. This likely will be a problem down the line, but there are also other design choices that are limiting us in the same way and fixing them all would be out of scope for this PR (IMO).

If there are other low-cost solutions that are better I'd love to hear thoughts, this is still an unfamiliar code base.

ryandillinfelton: New version removes this.

fpjrh:
@karelhala can you let me know what you need to complete this review?
Thank you!

karelhala (approved):

PR #1479, opened 2022-01-26
Sticky headers in table

gabipodolnikova:
Table headers in all apps should be "sticky"


src/components/InventoryTable/EntityTable.js
```diff
@@ -142,6 +142,7 @@ const EntityTable = ({
                         ...actions && rows?.length > 0 && { actions }
                     } }
                     { ...tableProps }
+                    isStickyHeader
```
karelhala, line 145: Can you please.move it before `tableProps`? That way consumers can rewrite it if they don't like the sticky header.

karelhala (commented):
Simple and straight forward! I like these changes.

karelhala (approved):

PR #1488, opened 2022-02-01
Skeleton fixes

gabipodolnikova:


src/components/InventoryTable/EntityTable.js
```diff
@@ -147,7 +147,11 @@ const EntityTable = ({
                     <TableHeader />
                     <TableBody />
                 </PfTable> :
-                <SkeletonTable colSize={ columns.current?.length || 3 } rowSize={ 15 } />
+                <SkeletonTable
+                    colSize={ columns.current?.length || 3 }
+                    rowSize={ 15 }
+                    variant={variant ? variant : tableProps.variant}
```
karelhala, line 153: You can use short cutting here
```suggestion
                    variant={variant ?? tableProps.variant}
```

karelhala (commented):
Apart from the simplified shortcutting it looks good!

PR #1491, opened 2022-02-10
EESNTL-2112: apply APP glogabl-filter

mkholjuraev:
Fixes: https://issues.redhat.com/browse/ESSNTL-2122.

It is easy to test. Just run the PR locally and apply 'Ansible Automation Platform' global filter.

PR #1497, opened 2022-02-21
Add mssql global filter option

karelhala:
## New MSSQL filter is available

Chrome service recently added new filter to global filter https://github.com/RedHatInsights/insights-chrome/pull/1776 so let's pass it down the foodchain. This PR is similiar to https://github.com/RedHatInsights/insights-inventory-frontend/pull/1491

### DIFF
--- a/src/routes/InventoryTable.js
+++ b/src/routes/InventoryTable.js
@@ -150,6 +150,8 @@ const Inventory = ({
                         ...workloads?.SAP?.isSelected && { sap_system: true },
                         ...workloads && workloads['Ansible Automation Platform']?.isSelected
                             && { ansible: 'not_nil' },
+                        ...workloads?.['Microsoft SQL']?.isSelected
+                            && { mssql: 'not_nil' },
                         ...SID?.length > 0 && { sap_sids: SID }
                     }
                 }


johnsonm325 (approved):
Looks straight-forward

src/routes/InventoryTable.js
```diff
@@ -148,8 +148,10 @@ const Inventory = ({
                     system_profile: {
                         ...globalFilter?.filter?.system_profile,
                         ...workloads?.SAP?.isSelected && { sap_system: true },
-                        ...workloads && workloads['Ansible Automation Platform']?.isSelected
+                        ...workloads?.['Ansible Automation Platform']?.isSelected
                             && { ansible: { controller_version: true } },
+                        ...workloads?.['Microsoft SQL']?.isSelected
+                            && { mssql: { version: 'not_nil' } },
```
johnsonm325, line 154: According to Asa, it's a good idea to update this with `&& { mssql: 'not_nil' },` to prep for the future. Currently there is only one child, but that may change and this is a good way to prepare. See this exchage: https://ansible.slack.com/archives/C02T90PLJ5V/p1646040794757729

johnsonm325 (changes requested):
This change should be made after a rebase to take in Asa's change: https://github.com/RedHatInsights/insights-inventory-frontend/pull/1502

src/routes/InventoryTable.js
karelhala: Is this deployed? We should update the AAP filter as well. I saw a PR in chrome to update the AAP filter, but was not sure if it was fully deployed for MSSQL as well.

johnsonm325: #1502 is merged, so I guess the answer is that the AAP filter is already updated. If I understood Asa correctly, updating mssql here would prepare us for whenever mssql got another child. Once you rebase this, the AAP filter will update though.

Let me know if I'm misunderstanding your question.

karelhala: Sweet! Updated the code, looks even cleaner. Tested it locally and it filters properly, so everything is smooth.

johnsonm325 (approved):

PR #1502, opened 2022-02-23
Update AAP filter usage

kruai:
The current AAP filter does not work, because not all AAP hosts have `ansible.controller_version` populated.
On the API side, we're adding the ability to filter on whether a parent field has any child field with a value. This PR should not be merged until the API change is ready.

[This PR](https://github.com/RedHatInsights/insights-chrome/pull/1783) updates the global context filter with this change in insights-chrome.


kruai:
@bastilian @mkholjuraev @karelhala This is now ready for review! The API PR got merged and is now live in Stage.

ryandillinfelton (approved):

PR #1505, opened 2022-03-01
 Disable cache for rbac inventory read permissions 

MichalSajdik:


MichalSajdik:
https://issues.redhat.com/browse/VULN-2123

karelhala:
After reading the linked ticked I don't think it's the right place to do this kind of work. We are using [RBACHook](https://github.com/RedHatInsights/frontend-components/blob/master/packages/utils/src/RBACHook/RBACHook.js) to calculate currently active permissions. We can turn the hard refresh trough that place, turning off the RBAC this way can lead to a few problems. For instance every app that turns these off will have to treat the not accessible pages.

karelhala:
Frontend components now include an option to disable cache [@redhat-cloud-services/frontend-components-utilities/v/3.2.9](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-utilities/v/3.2.9), you can just change the `usePermissions` in `useInventoryWritePermissions.js` and pass `true` as last parameter and it should query the RBAC everytime.

leSamo:
Hi @karelhala I think the JIRA ticket doesn't explain fully what we are trying to do. We already check the permission for Inventory on our Vulnerability side (with cache bypass; as per your first comment) before we render the Inventory component. Inventory doesn't bypass cache so after changing permissions without refreshing, all Vulnerability components have up to date permissions, but Inventory table does not (as it's caching). There are two solutions:
- we check permission on our side and disable checking permissions inside Inventory (this PR)
- we add `true` as the last param so Inventory does not cache at all - less optimal

karelhala:
@leSamo the problem of checking rbac on vuln side of things bring a lot of other problems. For instance we have unified look and feel if user has no access to inventory. If Vuln would have to bypass this by disabling RBAC they would have to draw similiar components if user has no access to it.

Other thing is that if you disable it in Vuln it won't take effect in every other app. You would fix this in Vuln (with extra hours of implementing unauth pages) but if user were to navigate to compliance for instance they will still see unauth components. You'd basically have to go to each app that uses inventory, write similiar unauth screens and checks and keep them up to date. It's better to do that on inventory side of things, instead on each app.

Moreover if in future inventory team decides to add more granular permission checks vuln team would have to implement it as well and that would cause unwanted behavior.

I would understand this if you wanted to use custom unauth components, but that would be more of a discussion between UXers and PMs (we did this in the past and basically said inventory unauth should look the same in each app). Or if you wanted to show inventory even if the user doesn't have inventory rights, and this can bring potential security risk - admin would have removed view inventory, but kept some rules that show them just in vulnerability wouldn't know about it and user could see inventory in vulnerability app.

If you want I can dig up the email thread from the past where we were discussing just these problems.

MichalSajdik:
Disabling cache may not be performance friendly, but it can resolve issues in cost effective way, so I pushed new implementation supporting this approach.

karelhala (approved):

PR #1515, opened 2022-03-09
fix(ESSNTL-2194) Use compact table

AsToNlele:
Request was to change the table to compact version. I've centered the content as well.

In my opinion the name should be a bit smaller maybe md or even sm. Open for changes.

Before:
image

After:
image

### DIFF
--- a/src/components/InventoryTable/EntityTable.js
+++ b/src/components/InventoryTable/EntityTable.js
@@ -7,7 +7,8 @@ import {
     Table as PfTable,
     TableBody,
     TableHeader,
-    TableGridBreakpoint
+    TableGridBreakpoint,
+    TableVariant
 } from '@patternfly/react-table';
 import { mergeArraysByKey } from '@redhat-cloud-services/frontend-components-utilities/helpers/helpers';
 import { SkeletonTable } from '@redhat-cloud-services/frontend-components/SkeletonTable';
@@ -199,6 +200,7 @@ EntityTable.defaultProps = {
     hasCheckbox: true,
     showActions: false,
     rows: [],
+    variant: TableVariant.compact,
     onExpandClick: () => undefined,
     tableProps: {}
 };
--- a/src/components/InventoryTable/InventoryList.scss
+++ b/src/components/InventoryTable/InventoryList.scss
@@ -68,6 +68,9 @@
 }
 
 .ins-c-entity-table.pf-c-table {
+    td {
+        vertical-align: middle;
+    }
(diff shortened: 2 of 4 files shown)


src/components/InventoryTable/InventoryList.js
```diff
@@ -43,7 +44,7 @@ const ContextInventoryList = ({ showHealth, onRefreshData, ignoreRefresh, ...pro
 
     return (<Grid gutter="sm" className="ins-inventory-list">
         <GridItem span={ 12 }>
-            <InventoryEntityTable { ...props } onRefreshData={onRefreshData} />
+            <InventoryEntityTable { ...props } variant={TableVariant.compact} onRefreshData={onRefreshData} />
```
bastilian, line 47: This might be better as a `variant` prop of the `InventoryTable` with the TableVariant.compact set as the default.
There might be cases where an application explicitly uses a specific variant and setting it here would overwrite that.

bastilian (changes requested):

src/components/InventoryTable/EntityTable.js
```diff
@@ -199,6 +200,7 @@ EntityTable.defaultProps = {
     hasCheckbox: true,
     showActions: false,
     rows: [],
+    variant: TableVariant.compact,
```
AsToNlele, line 203: Is it fine this way?

bastilian: If you look at line 181, there is already a `variant` prop defined in the `tableProps`. Changing the default for the `tableProps` below should do the trick.

AsToNlele: Unfortunately changing the ``tableProps`` defaults does nothing, because the default props are overwritten and not merged when the object is not empty.

Example:
When parent sets ``tableProps`` with ``canSelectAll``, default prop ``variant`` is not applied
image

But when parent doesnt set ``tableProps``, default prop ``variant`` is applied
image

I've found a fix on line 150 by merging the tableProps with the defaultProps:
from
```{...tableProps}```
to
```{ ...Object.assign({}, EntityTable.defaultProps.tableProps, tableProps) }```

but it doesn't look that good.

Should I leave it as it is currently or is there a better way?

bastilian: You are right. Well, let's leave it as is, this works.

bastilian (approved):
Works for me! Thank you @AsToNlele!

PR #1521, opened 2022-03-11
Remove "RHEL machine id" from collection card

kruai:
This PR addresses [ESSNTL-1996](https://issues.redhat.com/browse/ESSNTL-1996).
Basically, "RHEL machine id" maps to the "rhel_machine_id" field on host records; that field no longer exists, and is not returned by the API, so this removes it from the UI.


ryandillinfelton (approved):
LGTM

PR #1522, opened 2022-03-16
chore: handle undefined registeredWithFilter

mkholjuraev:
Intended to fix `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'filter')` error in Drift app.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -153,8 +153,8 @@ export async function getEntities(items, {
 
         return data;
     } else if (!hasItems) {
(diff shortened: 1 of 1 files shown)


bastilian:
@mkholjuraev Do we know why the `filters` is undefined?

johnsonm325 (approved):
Tested this with local drift-frontend and the change allows the systems table to properly render. Great job!

mkholjuraev:
@bastilian while only in Invetory application, it was never undefined. as registeredWithFilter comes as a  hard coded constant. But I am not sure why this is not the case in Drift.

PR #1530, opened 2022-04-04
Check if permissions are being loaded

rvsia:
We have to check if permissions are being loaded to show/hide the spinner

### DIFF
--- a/src/Utilities/Wrapper.js
+++ b/src/Utilities/Wrapper.js
@@ -4,13 +4,13 @@ import { usePermissions } from '@redhat-cloud-services/frontend-components-utili
 import { Spinner } from '@patternfly/react-core';
 
 const RenderWrapper = ({ cmp: Component, hideLoader, isRbacEnabled, inventoryRef, store, ...props }) => {
-    const { hasAccess } = usePermissions('inventory', [
+    const { hasAccess, isLoading } = usePermissions('inventory', [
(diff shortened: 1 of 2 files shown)


rvsia:
@kruai Hello, can you please assign someone to this PR? This issue fixes an issue that ROS (and probably all other teams) encounters - for a short period of time when the table mounts, the "No Permissions" page is shown instead of the spinner.

kruai (approved):
@rvsia This looks good to me, so I can approve... if you'd feel more comfortable having the Interact team validate as well, they can be contacted via the #team-interact channel on Ansible Slack.

This wouldn't happen to also fix the issue where the table straight-up doesn't appear for these apps ([link here](https://ansible.slack.com/archives/C02T90PLJ5V/p1648821707521579)), would it?

rvsia:
@kruai

I feel this PR is so simple that there is no need for more people to check it.

And yes - the table, at least in ROS is still broken. Something changed in the recent versions of the inventory app - I ran an older version today locally and it was OK, after `npm install` it broke.

kruai:
OK... if that's the case, I may not have validated thoroughly enough when I approved Dependabot's PRs. I'll go back through them and see if I missed something...

rvsia:
@kruai In my previous experience, I had a lot of issues with the auto generated client - it shifts params etc.

PR #1532, opened 2022-04-04
ESSNTL-2257

adonispuente:
With this change, when clicking 'Reset Filters' button in the inventory table, the filter chips go back to the default filters of **status fresh** and **stale**, and the URL reflects that.

To see changes navigate to inventory page, 
Add a filter (or remove one)
Press the reset filter button
The default filters are restored, and the URL will reflect the parameters

### DIFF
--- a/src/components/InventoryTable/EntityTableToolbar.js
+++ b/src/components/InventoryTable/EntityTableToolbar.js
@@ -186,6 +186,7 @@ const EntityTableToolbar = ({
      * @param {*} refresh refresh callback function.
      */
     const onSetFilter = (value, filterKey, refresh) => {
+        console.log(filters)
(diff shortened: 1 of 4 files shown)


mkholjuraev (commented):
While testing your fix locally, I faced an issue in the screen video below. On the first reset filter event, it works as expected and resets the filter to stale and fresh. However, if I again filter by some other filter and click on Reset filter button I get an error. Let me know if I am missing something.

https://user-images.githubusercontent.com/59481011/162180607-c1521f5a-db50-47d7-a8ca-04ea06e5fffc.mp4

johnsonm325 (approved):
Just fix the linting error and we're good to go.

PR #1533, opened 2022-04-05
fix(API): ESSNTL-1392 - Include hostnameOrId to filter tags

bastilian:
This fix the issue with the tags in the dropdown for the filter not being filtered properly when searching with the "Name" filter in the `InventoryTable`.


gabipodolnikova (approved):
LGTM!

PR #1535, opened 2022-04-06
Bump @redhat-cloud-services/frontend-components to v3.8.4

kruai:
This is to fix the skeletonTable component, which was fixed here: https://github.com/RedHatInsights/frontend-components/pull/1498


kruai:
As part of this, I've updated 4 snapshots that I believe were incorrectly updated in #1527.

kruai:
It's unclear to me why one test is failing in the PR build, when the test passes locally...

ryandillinfelton (approved):

PR #1573, opened 2022-06-03
Fix for ESSNTL-2787

adonispuente:
Reset Filters button is now always present. When adding an osFilter , if you then remove another filter behind it, it does not change the url to display operating system three times with arguments 0,1,2. 

To replicate first fix, 
- Navigate to inventory
- deselect all filters 
- Reset Filter button is still present, where before it was not

To replicate second fix
- Navigate to inventory
- Choose osFilter , 8.0 , 9.0 , whichever you like
- remove fresh and stale filters
- You'll see the url is valid, and filters are working properly
- Remove all filters, you'll see that reset filter button is still present, and if you select it you're back to default filters

### DIFF
--- a/src/components/InventoryTable/EntityTableToolbar.js
+++ b/src/components/InventoryTable/EntityTableToolbar.js
@@ -8,7 +8,6 @@ import { PrimaryToolbar } from '@redhat-cloud-services/frontend-components/Prima
 import { fetchAllTags, clearFilters, toggleTagModal, setFilter } from '../../store/actions';
 import { defaultFilters } from '../../Utilities/constants';
 import debounce from 'lodash/debounce';
-import flatMap from 'lodash/flatMap';
 import {
     TagsModal,
     TEXT_FILTER,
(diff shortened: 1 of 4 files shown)


src/routes/InventoryTable.js
```diff
@@ -107,18 +107,26 @@ const Inventory = ({
             options.filters = Object.entries(defaultFilters).map(([key, val]) => ({ [key]: val }));
         }
 
+        let results = [];
+        options?.filters.forEach((item)=>{
```
bastilian, line 111: Why not `.filter` the filters (and avoid `let` and filling it)?

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -342,7 +327,7 @@ const EntityTableToolbar = ({
                     }))
                 }
             }}
-            { ...isFilterSelected() && hasAccess && { activeFiltersConfig: constructFilters() } }
```
bastilian, line 345: What was the reasoning for removing this? ...I'm very suspicious of this change. :D

adonispuente: My thinking is that now we're setting activeFiltersConfig regardless of if a filter is selected or not. activeFiltersConfig is only applied if a filter is selected. We want the 'reset filter' button to always display regardless of if a filter is selected. So now we only need to check if the user has access.

adonispuente: My thinking is that now we're setting activeFiltersConfig regardless of if a filter is selected or not. activeFiltersConfig is only applied if a filter is selected. We want the 'reset filter' button to always display regardless of if a filter is selected. So now we only need to check if the user has access.

src/routes/InventoryTable.js
adonispuente: Honestly I just thought it looked better. With the way the data is formatted I can do something like
        let results = options?.filters.filter((item)=> item.osFilter);
but then in the operatingSystem conditional on line 129, itd have to be 
results2[0]?.osFilter 
and I thought that looked wonky

Fewwy (approved):
Tested it, great work
LGTM

PR #1583, opened 2022-06-15
fix(InventoryTable): Mount the component to allow state updates

bastilian:
This fixes this issue:

Screenshot 2022-06-15 at 12 41 37

However, I am not 100% sure what was/is happening. The `loading` was for RBAC permissions and the RenderWrapper is also requesting RBAC permissions. So, there must be a connection there, but I don't know how or what.


adonispuente (approved):
worked great while testing, LGTM

karelhala (approved):
Kinda strange the `loading` is causing this error. We should investigate why is it doing this strange behavior. But let's merge it to fix the issue and we can investigate later.

PR #1593, opened 2022-06-30
feat(RBAC): Use RBAC hook with context

bastilian:
This PR is intended to improve the loading behaviour of Inventory components and routes (using the RBAC hook).

It changes the components using the `usePermissions` hook to use a new iteration of the hook, which uses a context[0].
It also cleans up some things using the redux store and removes the need for the permissions loader and reducer.

[0] **Required**: https://github.com/RedHatInsights/frontend-components/pull/1576


bastilian:
This has been cherry picked into https://github.com/RedHatInsights/insights-inventory-frontend/pull/1603

PR #1600, opened 2022-07-14
Fix hardcoding changes from 2787

adonispuente:
In PR 2782, I updated the wrong InventoryTable (there are 3). This PR reverts the hard coding of ShowDeleteButton in an inventory table that gets imported, and hard codes it in the one just for this repo. The reset Filter Button is to always be present as per UI/UX. In inventory the Reset Filter button should always be present with this PR.

 To see if it affects another Repo, you can run something like advisor or vulnerability via` npm run start:proxy:beta -- --port=8004` and run inventory with something like `LOCAL_API=vulnerability:8004~https npm run start:proxy`.

### DIFF
--- a/src/components/InventoryTable/InventoryTable.js
+++ b/src/components/InventoryTable/InventoryTable.js
@@ -201,7 +201,7 @@ const InventoryTable = forwardRef(({ // eslint-disable-line react/display-name
                     paginationProps={paginationProps}
                     loaded={loaded}
                     showTagModal={showTagModal}
-                    activeFiltersConfig={{ deleteTitle: 'Reset filters', ...props.activeFiltersConfig, showDeleteButton: true }}
+                    activeFiltersConfig={{ deleteTitle: 'Reset filters', ...props.activeFiltersConfig }}
                 >
                     { children }
                 </EntityTableToolbar>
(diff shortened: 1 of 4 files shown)


src/components/InventoryTable/InventoryTable.js
```diff
@@ -201,7 +201,7 @@ const InventoryTable = forwardRef(({ // eslint-disable-line react/display-name
                     paginationProps={paginationProps}
                     loaded={loaded}
                     showTagModal={showTagModal}
-                    activeFiltersConfig={{ deleteTitle: 'Reset filters', ...props.activeFiltersConfig, showDeleteButton: true }}
+                    activeFiltersConfig={{ deleteTitle: 'Reset filters', ...props.activeFiltersConfig }}
```
bastilian, line 204: Let's add a test for when the InventoryTable has a `activeFiltersConfig` prop set that it pass it along to the EntityTableToolbar.

adonispuente: Just added

bastilian (approved):
Awesome! Thank you @adonispuente!

bastilian:
@adonispuente Though, please before merging squash and update the commit msg. ;)

PR #1603, opened 2022-07-20
fix(App): ESSNTL-2943 - Store registry in state

bastilian:
The registry has been redefined on calls to `register`, which caused a re-renders, which in turn could cause issues if the store wasn't proper.

https://github.com/RedHatInsights/frontend-components/pull/1576 has been merged and packages have been bumped to include the release with the changes.


gkarat (commented):
Looks good. I didn't run locally, but the changes look equal to the previous version with an exception to that `registry` is set only once on the `InventoryApp` render.

However, what was making the component re-render in the previous version? I see you mentioned the `registry` variable that was modified and caused the re-render, however, what caused the `registry` to be updated then? Is it the `init` function that changed its reference or smth?

bastilian:
@gkarat Thanks for taking a look! Yes, the fix is relatively small, however, we should test it thoroughly locally. 

Before this change calls to `getRegistry().register` like this one:https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/routes/InventoryDetail.js#L64 Would mutate what is assigned to `registry` and passed via the context. This would cause parts to re-render, which could in some cases when they don't rerender in a particular order cause the store to get messed up and raise errors like `entityDetails is undefined` and such.

gkarat:
@bastilian, I understand that there are re-renders of the inventory table caused by mutation of one of the `registry` deps. Still, is it desired that the inventory table no longer listens to the registry change? Wouldn't it cause any issues that you set the registry only once? I have a little knowledge about the registry and how it works across the chroming. But just want to make a double check with a bit abstract questions :)

Also, I ran the PR locally, and made just simple operations on the table... so far, so good

bastilian:
@gkarat It is not that the InventoryTable doesn't listen to changes or `register` calls anymore, it still works all like it should. Without keeping the registry in a state, the calls to register would create a new instance of the `ReducerRegistry` and (try to) reinitialise the state, in some cases improperly. Now the registry is the same instance and calls to `register` will use the same store to add/replace the reducers.

mkholjuraev (commented):
When I tried running the PR locally, I noticed one regression that needs to be taken into attention.
1. When one or more systems are selected and the user clicks on a system, the page does not get directed to the details page. Instead, the page gets refreshed.

https://user-images.githubusercontent.com/59481011/180747057-e8dad9ad-9e30-4662-8502-03d959c5f5a3.mp4

bastilian:
@mkholjuraev It seems to work as expected here: 

https://user-images.githubusercontent.com/7757/180958347-ed30fff0-1e4e-410a-8714-4b5ddb216c6e.mp4

Did you try to reinstall packages?

mkholjuraev:
@bastilian I had not tried reinstalling deps as the PR does not have changes. Let me try this...

mkholjuraev:
I have tried reinstalling deps, but I am still getting the issue intermittenly

https://user-images.githubusercontent.com/59481011/180995645-8aadd510-4bc5-4e37-a704-7fd4bf9ceadb.mp4

bastilian:
@mkholjuraev I was able now to reproduce it, but I need to do some digging to see what's going on. There doesn't appear to be any obvious error that would trigger the redirect.

src/Utilities/Wrapper.js
```diff
@@ -1,25 +1,22 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { usePermissions } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
-import { Spinner } from '@patternfly/react-core';
+import { usePermissionsWithContext } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
 
 const RenderWrapper = ({ cmp: Component, hideLoader, isRbacEnabled, inventoryRef, store, ...props }) => {
```
gkarat, line 5: Can you remove `hideLoader` from here and the propTypes since it is nowhere used anymore?

src/Utilities/Wrapper.js
```diff
@@ -1,25 +1,22 @@
...
-    const { hasAccess, isLoading } = usePermissions('inventory', [
+    const { hasAccess } = usePermissionsWithContext([
         'inventory:*:*',
         'inventory:*:read',
         'inventory:hosts:read'
-    ], true);
+    ]);
     return (
-        (isLoading && !hideLoader) ?
```
gkarat, line 13: I have a very superficial understanding of the code, but I immediately had a question if this is desired to remove the spinner and the loading state from the component logic...

gkarat (approved):
another quick look. overall, looking good

adonispuente (approved):
There are still some comments that need to be addressed but this PR is working as intended, great job! LGTM \o/

src/Utilities/Wrapper.js
bastilian: It would not make much sense here, as it would not be rendered since, the loading state for the permissions is now implemented in the `RBACProvider`. https://github.com/RedHatInsights/frontend-components/blob/88bc536dc7284618d96eb81d20ac4a3441d66ec4/packages/components/src/RBACProvider/RBACProvider.tsx#L50

src/Utilities/Wrapper.js
bastilian: Thanks for catching that! ...Odd that eslint did not notice it...

PR #1608, opened 2022-07-27
fix(InsightsDisconnected): update the disconnected systems verification logic

mkholjuraev:
Fixes: https://issues.redhat.com/browse/ESSNTL-2702

To test:

1. visit the inventory app
2. click on the filters dropdown
3. click on the 'Data collector' filter and select the 'insights-client not connected' option
4. verify that all hosts in the list have a disconnected icon next to the last seen date
5. click on any disconnected systems to open the details page
6. verify that the disconnected icon exists on the detail header as well and that the distance between the last seen date and the icon is smaller.  

Last seen info in the details page before:
image

Last seen info in the details page after:
image


src/Utilities/sharedFunctions.js
```diff
@@ -0,0 +1,19 @@
...
+    date.setDate(date.getDate() - numOfWeeks * 7);
+
+    return date;
+};
+
+export const verifyDisconnectedSystem = (perReporterStaleness) => {
+    if (perReporterStaleness && !perReporterStaleness.puptoo) {
+        return true;
+    } else if (perReporterStaleness && perReporterStaleness.puptoo) {
+        const stalenessDate = new Date(perReporterStaleness.puptoo?.stale_timestamp);
```
kruai, line 11: Is it necessary to use `?.` here? Just wondering since you confirmed the presence of `perReporterStaleness.puptoo` in the conditional

kruai, line 11: Is it necessary to use `?.` here? Just wondering since you confirmed the presence of `perReporterStaleness.puptoo` in the conditional

kruai (approved):
Just had one question, but looks good to me!

src/Utilities/sharedFunctions.js
mkholjuraev: Yup, you are correct, I will update the PR. Thanks for the review.

PR #1615, opened 2022-08-04
fix(AppEntry): Calling init with undefined causes an error in non-dev

bastilian:
When passing `undefined` in non-dev environments, it would raise an error, as it expects a middleware function.


mkholjuraev (approved):
LGTM! thanks @bastilian.

PR #1616, opened 2022-08-04
fix(Routes): Use proper react-router Redirect

bastilian:
The custom `checkPaths` appears to have stopped working correctly and cause an unintended redirect to the root on details pages under `/beta`.

Using the `Redirect` still preserves the redirect to the root when no matching route is found, but does not cause unintended redirects on `/beta` details pages.


mkholjuraev (approved):
Works great resolves refresh issue on beta! @bastilian thank you!

PR #1617, opened 2022-08-08
Use named exports when using fallback in General info cards

karelhala:
### Description

There's an error when importing default exports in general info card when running in fed modules. This is caused by race condition where Webpack is trying to figure out dependencies for default exports at the same time and thinks there are circular dependencies. This can be fixed by using named exports in inventory and whenever some app wants to use these details in fed modules they will use `default` exports. The problem can currently be observed on `stage` environment in `edge` application. Once this PR is released there's no need for an Edge application to update anything.


src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -10,13 +10,13 @@ import { SortByDirection } from '@patternfly/react-table';
...
 import { systemProfile } from '../../../store/actions';
 import InfoTable from '../InfoTable';
-import OperatingSystemCard from '../OperatingSystemCard';
-import SystemCard from '../SystemCard';
-import BiosCard from '../BiosCard';
-import InfrastructureCard from '../InfrastructureCard';
-import ConfigurationCard from '../ConfigurationCard';
-import SystemStatusCard from '../SystemStatusCard';
-import DataCollectorsCard from '../DataCollectorsCard/DataCollectorsCard';
+import { OperatingSystemCard } from '../OperatingSystemCard';
```
bastilian, line 13: If it is required to use the named exports, let's add a comment here to indicated that and why.

I can easily see myself going over this in the future and using the default exports again. :D

bastilian (approved):
LGTM! Thank you @karelhala!

arivepr (approved):
Hope I didn't do an oopsie by updating the branch! 😅 
T'was complaining. 

Either way, once the build passes I'll merge in!

PR #1618, opened 2022-08-10
fix(ESSNTL-3183): fix insights-client status in General informaion tab

mkholjuraev:
Resolves:

1. [ESSNTL-3183](https://issues.redhat.com/browse/ESSNTL-3183) Makes insights-client status consistent with UX guidelines. Status becomes 'Stale' when the reporter staleness is set, 'N/A' when there is no insights-client id. Otherwise always 'Active'.
2. [ESSNTL-3204](https://issues.redhat.com/browse/ESSNTL-3204) updates condition of showing InsightsPrompt component to use per_reporter_staleness


adonispuente:
When testing the PR and looking at ticket [ESSNTL-3183](https://issues.redhat.com/browse/ESSNTL-3183) it says that if you have the filters  'Stale' and 'insights-clients' and then click on the first system, the Data collector portion should show 'Stale'. I found in testing that i

https://user-images.githubusercontent.com/60629070/184169874-064a63f0-c5ba-4fa3-9cf1-58ff98e4bf08.mp4

t still shows 'Active'

mkholjuraev:
@adonispuente thank you for the review. It seems to me that I did not understand the description. I will work on it.

mkholjuraev:
Now, this must show insights-client as stale when you open a stale host.

src/Utilities/sharedFunctions.js
```diff
@@ -4,16 +4,29 @@ export const subtractWeeks = (numOfWeeks, date = new Date()) => {
...
-            return true;
-        }
+export const verifyCollectorStaleness = (reporterStaleness) =>{
+    const stalenessDate = new Date(reporterStaleness.stale_timestamp);
+    const currentDateTime = new Date();
+
+    const twoWeeksPeriod = subtractWeeks(2);
+    const oneWeeksPeriod = subtractWeeks(1);
+
+    if (stalenessDate > currentDateTime) {
```
adonispuente, line 14: This is more of a question/nitpick. Do you think its more readable if you make the conditionals something like this?

```suggestion
}; if (stalenessDate > currentDateTime) {
        return 'Fresh';
    } else if ((oneWeeksPeriod < stalenessDate) && (stalenessDate < currentDateTime)) {
        return 'Stale';
    }
    else if ((twoWeeksPeriod < stalenessDate) && (stalenessDate < oneWeeksPeriod)) {
        return 'Stale warning';
    } else {
        return 'Culled';
    }
```

adonispuente (commented):
The first ticket seems to be resolved now but im having trouble testing the second. For ESSNTL-3204, i cant find a system with insights-clients not connected in order to test on insights-qa, did you use a different account? If not, how can I test this?

src/Utilities/sharedFunctions.js
mkholjuraev: You can find an insights-client not connected system by filtering with the 'Insights not connected' data collector. It is among the filters in the inventory table toolbar. I admit that this if/else block is not optimal. But this condition can not be achieved with other ways in my opinion (switch...).

adonispuente: Oh the if/else chain I posted didnt change it, it just adds parentheses inside the conditional.

adonispuente (approved):
PR works as intended, LGTM!

PR #1619, opened 2022-08-10
fix(InventoryList): ESSNTL-2639 - Tag Filter UI

AsToNlele:
[ESSNTL-2639](https://issues.redhat.com/browse/ESSNTL-2639)

Before:
image

After:
image

### DIFF
--- a/src/components/InventoryTable/InventoryList.scss
+++ b/src/components/InventoryTable/InventoryList.scss
@@ -22,7 +22,10 @@
 
     .ins-c-chip-filters .pf-c-chip-group:not(:last-of-type) { margin-right: var(--pf-global--spacer--xs); }
 
-    .ins-c-tagfilter { width: initial; }
+    .ins-c-tagfilter { 
+        width: initial;
+        padding: 0;
+    }
 }
 
 .ins-c-table__empty {


src/components/InventoryTable/InventoryList.scss
```diff
@@ -22,7 +22,10 @@
 
     .ins-c-chip-filters .pf-c-chip-group:not(:last-of-type) { margin-right: var(--pf-global--spacer--xs); }
 
-    .ins-c-tagfilter { width: initial; }
+    .ins-c-tagfilter { 
+        width: initial;
+        padding: 0;
```
bastilian, line 27: While this does fix the issue, the question still remains, why it is only the tag filter that is blown up. Where does it get wider, higher and/or more padding?

bastilian: It is also suspicious that on other applications the tag filter, which should be the exact same component isn't blown up there. This is compliance for example:

Screenshot 2022-08-10 at 14 51 26

AsToNlele: I did some digging and found [this](https://github.com/RedHatInsights/insights-advisor-frontend/blob/6a1d4d7a8b075e4b3d0d0faa9cb0595374f891f5/src/App.scss#L50) in advisor. So that's why it works in Advisor. Should I make a PR to the Frontend components or what is the best way to handle this?

bastilian: @AsToNlele Sorry, for the delayed response. Yes, i think we should rather look into fixing this in the frontend-components and remove the `padding: 0;` from each app. Advisor is not the only app that does this, compliance, malware and possible other apps do this as well (see https://github.com/search?q=org%3ARedHatInsights+ins-c-tagfilter&type=code).

We should find out why the tags filter even expands, compared to other dropdown filters. We should start with the tag filter here: https://github.com/RedHatInsights/frontend-components/tree/0c0a27a1ff201d1bfa129a58d2d1c8300c566605/packages/components/src/FilterHooks

If you need help setting up a dev environment with frontend components, let me know and I can help you set it up. You might also want to run an application that applies the padding, like compliance, but as a start it should be enough to link the frontend components in the inventory, remove the padding here, track down where in the FEC the padding/expanding is coming from and than remove the padding in other apps when we bump the FEC packages to the versions with a fixed tagfilter here in the inventory.

johnsonm325: @AsToNlele @bastilian Here's where the problem seems to stem from. There is an extra `TextInput` component inside of the dropdown and that adds extra padding. We can simply fix it by adding a custom css property there to remove all padding and it should work. https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/src/ConditionalFilter/GroupFilter.tsx#L305-L316

AsToNlele: @johnsonm325 oh thanks for looking into this. I was trying to debug this by running the ``frontend-components`` locally and then npm linking it to the inventory. Unfortunately I wasn't able to make it running as I had a "permissions" error. Are you able to connect local ``frontend-components`` with inventory?

PR #1621, opened 2022-08-17
ESSNTL-2806: Get list of operating systems from API & OS version filter bug fixes

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-2806.

The PR changes the behavior of the Inventory table OS filter: instead of statically pre-defined versions, the table will request the Inventory API (GET /system_profile/operating_system) to obtain the list of OS versions available for an account.

### Screenshots

Successful state (request fulfilled):

image

Error state (request failed):

image

### Notes

- [x] Make the OS filter render versions applicable for a user account
- [x] Check that OS filter modification changes the URL search parameters
- [x] Repair affected tests
- [x] How to handle the error state (the operating systems were not fetched successfully?)
- [x] Check other places where OS versions are used: see https://github.com/RedHatInsights/insights-inventory-frontend/pull/1621#issuecomment-1219651178

### Regressions check:

- [x] Selecting version checkbox adds a new chip to the toolbar and updates the URL search parameters accordingly.
- [x] Removing a chip removes the selection from the filter toolbar and also modified URL accordingly.
- [x] Deselecting a checkbox removes the associated chip and updates URL.
- [x] Navigating to the page with predefined `operating_system` URL search parameter correctly applies the OS filter.


src/components/InventoryTable/EntityTableToolbar.test.js
```diff
@@ -207,10 +229,14 @@ describe('EntityTableToolbar', () => {
         });
 
         describe('delete filter', () => {
+            beforeEach(() => {
+                mockTags.onGet(/\/api\/inventory\/v1.*/).reply(200, { results: [] });
+                mockSystemProfile.onGet().reply(200, { results: [] });
+            });
+
             it('should dispatch action on delete filter', async () => {
```
gkarat, line 237: This is the only test failing currently because it can't find a mock for the /operating_system request:
```
Could not find mock for: 

    {

      "method": "get",

      "url": "/api/inventory/v1/system_profile/operating_system"

    }
```
Still don't know why all other tests pass with the same mocks configuration.

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -22,15 +22,15 @@ export const groupOSVersions = (osVersions) => {
...
+export const buildOSFilterConfig = (config = {}, systems) => ({
     ...config,
     label: 'Operating System',
     value: 'os-filter',
     type: 'group',
     filterValues: {
         selected: config.value,
         onChange: config.onChange,
-        groups: groupOSVersions(operatingSystems).map((item) => ({
+        groups: groupOSVersions(systems).map((item) => ({
```
mkholjuraev, line 33: Nitpick: I would opt for renaming back to 'operatingSystems'

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -83,10 +83,16 @@ const EntityTableToolbar = ({
     const allTagsLoaded = useSelector(({ entities: { allTagsLoaded } }) => allTagsLoaded);
     const allTags = useSelector(({ entities: { allTags } }) => allTags);
     const additionalTagsCount = useSelector(({ entities: { additionalTagsCount } }) => additionalTagsCount);
+    const operatingSystems = useSelector(({ entities: { operatingSystems } }) => operatingSystems);
+    const operatingSystemsLoaded = useSelector(({ entities: { operatingSystemsLoaded } }) => operatingSystemsLoaded);
```
mkholjuraev, line 87: Nitpick: Let's make it apparent that this is a boolean by renaming it to 'areOperatingSystemsLoaded'. if you feel this is a long,

src/components/filters/useOperatingSystemFilter.js
```diff
@@ -9,14 +9,20 @@ export const operatingSystemFilterReducer = (_state, { type, payload }) => ({
     }
 });
 
-export const useOperatingSystemFilter = ([state, dispatch] = [operatingSystemFilterState]) => {
+export const useOperatingSystemFilter = (
+    operatingSystems,
+    operatingSystemsLoaded,
+    [state, dispatch] = [operatingSystemFilterState]
+) => {
+    // FIXME: is this state value needed and used at all?
```
mkholjuraev, line 17: I also think that this can be removed.

mkholjuraev (commented):
A few nitpicks, I have bumped an issue with filter chips. When I filter by 8.0, 8.1 and then by 9.1, the chip for 9.1 is missing.

https://user-images.githubusercontent.com/59481011/185588928-db1124a7-4043-4cce-9239-11eb572846f6.mp4

src/components/InventoryTable/EntityTableToolbar.js
gkarat: @mkholjuraev, perhaps we could leave it as it is? in my opinion, the adjective `loaded` is already pointing to the true or false nature of this variable

src/components/filters/useOperatingSystemFilter.js
gkarat: why? I still see this state variable unused and wondering whether this is needed at all

gkarat:
@mkholjuraev, thank you for review. I fixed your bug in https://github.com/RedHatInsights/insights-inventory-frontend/pull/1621/commits/88a0004a81be43b625692b3db1760e27757bd427. This was funny one: the chips were missing because the static list of OS that remained in code didn't have 9.1 version yet and thus the value was filtered out. I completely removed this list and now I implemented the check in `buildOSChip` by verifying value with the OS list from API.

src/components/filters/useOperatingSystemFilter.js
mkholjuraev: I believe that this state variable is used as a default value holder. But this is not needed if we are always getting `operatingSystemFilter` from the state prop (I think we are getting this always).

mkholjuraev (changes requested):
It looks like there is a regression. Now all selected OS versions are reflected in the chips, but when I deselect, some of them are staying in the chips.
image

src/components/filters/useOperatingSystemFilter.js
bastilian: This state unused (in Inventory). I believe it was meant as a default for when the hook would be used outside the Inventory without redux. There doesn't appear to be any other app using it though. We should double check though.

gkarat: Can we leave this FIXME comment, so that we don't forget later, after this PR, make a check? :-)

gkarat:
@mkholjuraev, could you please check if this bug has disappeared? I fixed it in the last commits. I also managed to fix the bug when you opened the URL with pre-defined search query params like `operating_system=8.1`, but no filters were applied.

I will revisit the PR code, and probably simplify some places, where it is possible. But the pull request is more or less ready for review.

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -9,43 +10,76 @@ export const toGroupSelectionValue = (osVersions) => osVersions.reduce((acc, ver
...
-        type: OS_CHIP,
-        chips: operatingSystems.filter(({ value: osValue }) => value.some((osFilterValue) => osFilterValue === osValue))
-        .map(({ label, ...props }) => ({ name: label, ...props }))
-    }] : []);};
+export const buildOSChip = (operatingSystemValue = {}, operatingSystems) => {
+    const osGroups = Object.values(operatingSystemValue);
+    const minors = osGroups.flatMap((group) =>
+        Object.entries(group)
+        .filter((entry) => entry[1] === true)
+        .map((entry) => entry[0])
```
karelhala, line 70: You can use destruct operator here
```suggestion
        .map(([ entry ]) => entry)
```

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -9,43 +10,76 @@ export const toGroupSelectionValue = (osVersions) => osVersions.reduce((acc, ver
...
-        category: 'Operating System',
-        type: OS_CHIP,
-        chips: operatingSystems.filter(({ value: osValue }) => value.some((osFilterValue) => osFilterValue === osValue))
-        .map(({ label, ...props }) => ({ name: label, ...props }))
-    }] : []);};
+export const buildOSChip = (operatingSystemValue = {}, operatingSystems) => {
+    const osGroups = Object.values(operatingSystemValue);
+    const minors = osGroups.flatMap((group) =>
+        Object.entries(group)
+        .filter((entry) => entry[1] === true)
```
karelhala, line 69: I would use desctruct and access just the value
```suggestion
        .filter(([, val]) => val === true)
```

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -9,43 +10,76 @@ export const toGroupSelectionValue = (osVersions) => osVersions.reduce((acc, ver
...
-    const value = Object.values(operatingSystemValue || {}).flatMap((majorOsVersion) => Object.keys(majorOsVersion));
-    return (value?.length > 0 ? [{
-        category: 'Operating System',
-        type: OS_CHIP,
-        chips: operatingSystems.filter(({ value: osValue }) => value.some((osFilterValue) => osFilterValue === osValue))
-        .map(({ label, ...props }) => ({ name: label, ...props }))
-    }] : []);};
+export const buildOSChip = (operatingSystemValue = {}, operatingSystems) => {
+    const osGroups = Object.values(operatingSystemValue);
+    const minors = osGroups.flatMap((group) =>
```
karelhala, line 67: No need to assign to a value, you can directly use `Object.values(operatingSystemValue)`
```suggestion
    const minors = Object.values(operatingSystemValue).flatMap((group) =>
```

src/api/api.js
```diff
@@ -235,3 +235,7 @@ export function getAllTags(search, { filters, pagination, ...options } = { pagin
         }
     );
 }
+
+export function getOperatingSystems() {
+    return fetch('/api/inventory/v1/system_profile/operating_system').then(response => response.json());
```
karelhala, line 240: Please don't use `fetch` we have an JS client for this available https://github.com/RedHatInsights/javascript-clients/blob/master/packages/host-inventory/doc/classes/systemprofileapi.md#apisystemprofilegetoperatingsystem
```suggestion
    return systemProfile.apiSystemProfileGetOperatingSystem()
```

You can (and probably should) also pass other applied filters to filter further down the list of OS. You'll probably have to import `SystemProfileApi` from inventory javascript client and add `const systemProfile = new SystemProfileApi(undefined, INVENTORY_API_BASE, instance);

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -194,6 +200,10 @@ const EntityTableToolbar = ({
 
     const shouldReload = page && perPage && filters && (!hasItems || items);
 
+    useEffect(() => {
+        dispatch(fetchOperatingSystems());
```
karelhala, line 204: Not needed for now, but I'd like to see a followup PR with adding applied filters when fetching OSs, that way users won't see an OS which is not available for current filter

src/store/entities.js
```diff
@@ -264,6 +265,17 @@ export default {
...
     [ACTION_TYPES.ALL_TAGS_REJECTED]: loadingRejected,
+    [ACTION_TYPES.OPERATING_SYSTEMS_PENDING]: (state) => ({ ...state, operatingSystemsLoaded: false }),
+    [ACTION_TYPES.OPERATING_SYSTEMS_FULFILLED]: (state, { payload: { results } }) => ({
+        ...state,
+        operatingSystems: results.map(entry => {
+            const { name, major, minor } = entry.value;
+            const versionStringified = major.toString() + '.' + minor.toString();
+            return { label: name + ' ' + versionStringified, value: versionStringified };
+        }),
+        operatingSystemsLoaded: true
```
karelhala, line 276: Can we move this into a specific function?

mkholjuraev (approved):
Works well! @gkarat thank you.

src/store/entities.js
```diff
@@ -264,6 +265,17 @@ export default {
     [ACTION_TYPES.LOAD_TAGS_PENDING]: showTagsPending,
     [ACTION_TYPES.LOAD_TAGS_FULFILLED]: showTags,
     [ACTION_TYPES.ALL_TAGS_REJECTED]: loadingRejected,
+    [ACTION_TYPES.OPERATING_SYSTEMS_PENDING]: (state) => ({ ...state, operatingSystemsLoaded: false }),
+    [ACTION_TYPES.OPERATING_SYSTEMS_FULFILLED]: (state, { payload: { results } }) => ({
+        ...state,
+        operatingSystems: results.map(entry => {
+            const { name, major, minor } = entry.value;
+            const versionStringified = major.toString() + '.' + minor.toString();
```
karelhala, line 273: No need to call `toString`, we can use string literals
```suggestion
            const versionStringified = `${major}.${minor}`;
```

src/components/InventoryTable/EntityTableToolbar.js
gkarat: @karelhala, could you please elaborate more on that (maybe, provide some examples)? This change is a main goal of this PR, and it was requested by UX to always show the list of OS versions available for the account. Another note from them also was that the list shouldn't change with other filters being applied.

src/api/api.js
gkarat: @karelhala, my first commits were containing the code that you have just suggested. However, the `systemProfile` client was introduced in later versions of `javascript-clients`, and that meant I also needed to upgrade the package in the Inventory app. However, after doing that, we ran into multiple issues caused by the breaking changes in the API, please see https://github.com/RedHatInsights/javascript-clients/issues/166. The number of bugs appeared caused by the upgrade of this client seemed to us inappropriate to solve within this PR, and we decided to take the golden path and implement the request by this one liner.

To your second comment: we are asked not to modify the list of OS versions. That means, there is no need to provide parameters to the call to `system_profile/operating_system`.

src/components/InventoryTable/EntityTableToolbar.js
karelhala: If the list shouldn't change with other filters applied I think we are good to go with this. The API should return just the OS versions user can apply.

jiridostal: @katierik Just tagging you for awareness

gkarat: Yes, exactly :) operating_system should respond only with the versions that are available for the current account.

src/api/api.js
karelhala: @gkarat understandable, let's open another PR where we update the version of the client to stay updated and we can use the version here as well. It would be really bad if half of the request were done with `fetch` and the other with client itself.

gkarat: Sounds reasonable, thank you

karelhala: Actually opened a PR to use the latest JS client version https://github.com/RedHatInsights/insights-inventory-frontend/pull/1628

gkarat:
The PR breaks other applications such as RHEL Advisor. Need to take a look and adjust the code so that it doesn't break consumers

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -331,7 +331,10 @@ const EntityTableToolbar = ({
                         ...filter,
                         filterValues: {
                             placeholder: filter?.filterValues?.placeholder || `Filter by ${filter?.label?.toLowerCase()}`,
-                            ...filter?.filterValues,
+                            /* TODO: remove the OS check once RHEL Advisor has buildOSFilterConfig function call updated */
```
gkarat, line 334: @karelhala, @bastilian, I would like to know your opinion on this issue. The thing is, RHEL Advisor calls `buildOSFilterConfig` and overrides filter configurations including the OS filter. However, the function, in order to polyfill available OS filter values, used the deleted `operatingSystems` constant. This leads to the situation when Advisor generates the OS filter config with 0 available values.

The fix in this commit should detect this situation and apply the values from the API call (that Inventory made, not Advisor). This is a kludge, but we can revisit this scenario later on. This only happens to RHEL Advisor (other apps do not override config using `buildOSFilterConfig`.

karelhala: @gkarat I would go the other way around and align Advisor to the new API before going further with this PR. That way we don't have to keep any outstanding TODOs. However if that would have proven to be too complex we can go this way and align Advisor later (later meaning the sprint right after this gets merged)

gkarat:
https://github.com/RedHatInsights/insights-inventory-frontend/pull/1621/commits/116d418a24008fef03fb1436aa4aee4583f7105a and https://github.com/RedHatInsights/insights-inventory-frontend/pull/1621/commits/116d418a24008fef03fb1436aa4aee4583f7105a help to fix another issue that appeared to RHEL Advisor Systems page.

gkarat:
@karelhala, thanks a lot for your help upgrading javscript-clients. Could you please take a look once again at this PR? I consider this as finished. I also made a manual test in compliance, drift, advisor, remediations, patchman, sed and vulnerability apps, it looks like working as expected (except some of the apps which override the OS versions with static values -- to be fixed soon).

cc: @RedHatInsights/team-interact

src/components/InventoryTable/EntityTableToolbar.js
gkarat: @karelhala, sounds reasonable. Let me try to take a look at the RHEL Advisor's code and see if we can easily rework this. I will link the PR to this one once I open.

karelhala (approved):
Looking good!

src/components/InventoryTable/EntityTableToolbar.js
gkarat: @karelhala, OK, this turned out to be possible. PTAL https://github.com/RedHatInsights/insights-advisor-frontend/pull/1027

gkarat:
https://github.com/RedHatInsights/insights-advisor-frontend/pull/1027 has to be merged first.

gkarat:
^ Squashed some of the commits, but there are no new changes.

PR #1631, opened 2022-09-08
fix(ESSNTL-3336): fix os column sorting

mkholjuraev:
Resolves: [ESSNTL-3336](https://issues.redhat.com/browse/ESSNTL-3336)

Fixes the OS columns sorting by making the column key and sort key consistent system_profile. In case, they are different, sorting does not work and we need to provide operating.

### DIFF
--- a/src/components/InventoryTable/helpers.js
+++ b/src/components/InventoryTable/helpers.js
@@ -31,22 +31,36 @@ export const createRows = (rows = [], columns = [], { actions, expandable, noSys
         }];
     }
 
-    return flatten(rows.map((oneItem, key) => ([{
-        ...oneItem,
-        ...oneItem.children && expandable && { isOpen: !!oneItem.isOpen },
-        cells: buildCells(oneItem, columns, extra),
-        actionProps: {
-            'data-ouia-component-id': `${oneItem.id}-actions-kebab`
-        }
-    }, oneItem.children && expandable && {
-        cells: [
-            {
-                title: typeof oneItem.children === 'function' ? oneItem.children() : oneItem.children
-            }
-        ],
-        parent: key * 2,
-        fullWidth: true
-    }]))).filter(Boolean);
+    return flatten(rows.map((oneItem, key) => {
+        //makes the row column key and sort key consistent, otherwise sorting does not work
+        const rowItem = {
+            ...oneItem,
+            // eslint-disable-next-line camelcase
+            operating_system: oneItem.system_profile?.operating_system
+        };
+
+        return (
+            [
+                {
+                    ...rowItem,
+                    ...rowItem.children && expandable && { isOpen: !!rowItem.isOpen },
+                    cells: buildCells(rowItem, columns, extra),
+                    actionProps: {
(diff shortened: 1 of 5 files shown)


src/components/InventoryTable/helpers.js
```diff
@@ -31,22 +31,36 @@ export const createRows = (rows = [], columns = [], { actions, expandable, noSys
...
-        ],
-        parent: key * 2,
-        fullWidth: true
-    }]))).filter(Boolean);
+    return flatten(rows.map((oneItem, key) => {
+        //makes the row column key and sort key consistent, otherwise sorting does not work
+        const rowItem = {
+            ...oneItem,
+            // eslint-disable-next-line camelcase
+            operating_system: oneItem.system_profile?.operating_system
```
gkarat, line 39: Is it ok that here can be an undefined value? Will it be treated as "Not available" in the table? Just checking

src/components/InventoryTable/helpers.js
```diff
@@ -31,22 +31,36 @@ export const createRows = (rows = [], columns = [], { actions, expandable, noSys
...
-    }, oneItem.children && expandable && {
-        cells: [
-            {
-                title: typeof oneItem.children === 'function' ? oneItem.children() : oneItem.children
-            }
-        ],
-        parent: key * 2,
-        fullWidth: true
-    }]))).filter(Boolean);
+    return flatten(rows.map((oneItem, key) => {
```
gkarat, line 34: I think you can use `flatMap` here

gkarat (approved):
Tried locally, everything seems to be fine 👍🏼 good job

src/components/InventoryTable/helpers.js
mkholjuraev: Yes, I have verified it.

mkholjuraev: I believe flatmap is just syntactical sugar. I am not a fan of touching what is working. 

_It is identical to a [map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) followed by a [flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) of depth 1 (arr.map(...args).flat())_  see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

mkholjuraev:
@gkarat thank you for the review!

src/components/InventoryTable/helpers.js
mkholjuraev: I believe flatmap is just syntactical sugar. I am not a fan of touching what is working :). 

_It is identical to a [map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) followed by a [flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) of depth 1 (arr.map(...args).flat())_  see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

PR #1632, opened 2022-09-12
Revert "fix(ESSNTL-3336): fix os column sorting"

mkholjuraev:
This reverts commit ff9be46e3a7f31289eeb1d31e99fb52c93ffc292. Also fixes the bug by handing this specific sort key by getting its index explicitly in the codebase. Please verify that it works by running locally.

### DIFF
--- a/src/components/InventoryTable/EntityTable.js
+++ b/src/components/InventoryTable/EntityTable.js
@@ -132,6 +132,7 @@ const EntityTable = ({
                     onSort={ (event, index, direction) => {
                         onSortChange(
                             event,
+                            cells?.[index - Boolean(hasCheckbox) - Boolean(expandable)]?.sortKey ||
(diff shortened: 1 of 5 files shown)


gkarat (commented):
I had a problem sorting the column in Advisor. Both inventory and advisor were deployed locally. To reproduce:

- Go to https://stage.foo.redhat.com:1337/beta/insights/advisor/systems 
- Sort by OS column

Screenshot from 2022-09-12 16-14-30
Screenshot from 2022-09-12 16-14-24

gkarat (approved):
But my feeling is that this issue comes from Advisor itself (also broken on prod). Would be good if you could confirm it.

Otherwise the code looks OK. Also, for test commit please use `test` prefix, not `chore`.

mkholjuraev:
@gkarat thanks for the review. I have found the culprit line on Advisor. It has also some key modification that needs to be adjusted. I will create a fix for it. Merging this one.

PR #1634, opened 2022-09-14
Update display name, once we fire the update function

karelhala:
### Description

There is a problem, when updating display name or ansible name - for a brief period of time the UI still shows old values. This PR bypasses this error by setting the value directly to the store without calling the server again. It treats the response as idealistic response, meaning it assumes everything will run smoothly. If there is an error while processing this change on server, the UI rollbacks the value to original one.


bastilian (approved):
LGTM! Thank you @karelhala!

bastilian:
@karelhala One minor thing, we try to associate Jiras to PRs when we can. If you merge this, and have a Jira ID at hand, would you mind adding it to the commit? (something like: `fix(ESSNTL-...): ....`)

karelhala:
@bastilian sure thing, updated the initial commit. It's a regression bug so I linked the original one.

PR #1638, opened 2022-09-20
WIP: refactor: Refactor useOperatingSystemFilter hook

gkarat:
This revisits the hook useOperatingSystemsFilter that is used to obtain a filter configuration, chips and further info to be used in tables. The main focus is to have the hook be usable not only by the Inventory systems table, but further apps. Also this implements the group selection.


src/components/filters/useOperatingSystemFilter.js
```diff
@@ -1,37 +1,88 @@
...
+            // use the previous value if avalable
+            set(acc, [major, version], get(selected, [major, version]) || false);
+            // update group selection identifier
+            set(acc, [major, major], Object.values({ ...acc[major] }).filter(v => v !== major).every(Boolean));
+            return acc;
+        }, {}));
+    }, [operatingSystems]);
+
+    // PrimaryToolbar filter configuration
+    const config = useMemo(() => ({
```
gkarat, line 53: needed to wrap couple of things in useMemo: because if declared with simple assignment `=`, this could lead to infinite reloading of tables that use this hook (e.g., some of the hook's return values can be added to table's useEffect dependencies)

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -1,31 +1,44 @@
...
-        ...(acc[`${majorVersion}.0`] || {}),
-        [`${majorVersion}.${minorVersion}`]: true
+   required by ConditionalFilter component (group filter);
+   inverse to `getSelectedOsFilterVersions`  */
+export const toGroupSelection = (osVersions = []) => osVersions.reduce((acc, version) => {
+    const [major] = version.split('.');
+    acc[major] = {
+        ...(acc[major] || {}),
+        [version]: true,
+        [major]: Object.values({ ...acc[major] }).filter(v => v !== major).every(Boolean)
```
johnsonm325, line 13: There is some repeated code here that can be found in `src/components/filters/useOperatingSystemFilter`. I think we can create some helper functions that we can use globally and are named well so we know what is being done here. I'm having a hard time looking at this line and knowing what's happening. A well named function could make the difference.

johnsonm325 (changes requested):
Functionally, this code seems great. I've added a few improvements that I'd love to hear your take on. There is some repeated code that I think can be abstracted out into smaller functions for improved readability and maintainability.

Some of these code blocks are quite complex and make it hard to understand. Your comments are helpful, but I think we can take it one step further.

If you have a hard time knowing how to do this, or you understand a reason why what I'm asking for can't be easily implemented, I think we can talk so you can help me better understand. I may easily be missing something.

src/components/filters/useOperatingSystemFilter.js
```diff
@@ -1,37 +1,88 @@
...
+/**
+* OS version filter hook.
+* @param {Array} apiParams - an array containing parameters for GET /system_profile/operating_system call
+* @return {Array} An array containing config object, chips array and value setter function.
+*/
+const useOperatingSystemFilter = (apiParams = []) => {
+    const dispatch = useDispatch();
+    const operatingSystems = useSelector(({ entities }) => entities?.operatingSystems);
+    const operatingSystemsLoaded = useSelector(({ entities }) => entities?.operatingSystemsLoaded) || false;
+    const availableVersions = operatingSystems === undefined ? [] : operatingSystems;
```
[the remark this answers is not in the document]
karelhala: Yeah, good point! We are just pulling `operatingSystems` from redux and rename them if found. If the `useSelector` doesn't find anything it returns undefined.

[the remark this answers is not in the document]
gkarat: @karelhala, @johnsonm325, this was done because one of `useEffect`s uses `operatingSystems` as a dependency. If, in redux store we forget to set `operatingSystems` default value to `[]` (the key doesn't exist, which can sometimes happen and we ideally have to cover such cases), this will lead for the hook to infinitely reload, because shallow comparison of two empty arrays equals to `false`. 

For example, I could see this while running tests: assume that we declare the variable with
```
const operatingSystems = useSelector(({ entities }) => entities?.operatingSystems) || [];
```
then the result of running tests will be the following

изображение

The reason is that one of the tests doesn't have `operatingSystems` mocked in its mock store. And yep, we just can update all the tests, but in the future, this will be really hard to understand whether the value in a store is optional or not.

src/components/filters/useOperatingSystemFilter.js
```diff
@@ -1,37 +1,88 @@
...
+    useEffect(() => {
+        setGroups(groupOSFilterVersions(availableVersions));
+        setSelected((availableVersions).reduce((acc, { value: version }) => {
+            const [major] = version.split('.');
+            // use the previous value if avalable
+            set(acc, [major, version], get(selected, [major, version]) || false);
+            // update group selection identifier
+            set(acc, [major, major], Object.values({ ...acc[major] }).filter(v => v !== major).every(Boolean));
+            return acc;
+        }, {}));
```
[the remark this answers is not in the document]
gkarat: I managed to merge these two different pieces of code into one function `toGroupSelection`, PTAL

src/Utilities/OperatingSystemFilterHelpers.js
gkarat: yep, true, updated with the last push 👍🏼

src/components/filters/useOperatingSystemFilter.js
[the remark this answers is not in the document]
gkarat: OK I managed to go another way: just wrapped with `|| []` check in places where the value can be undefined, but an array is expected. PTAL the last push

Fewwy (approved):
Fetched it and ran it locally. OS Filter works, great job!

gkarat:
Put don't merge label: I would like to have https://github.com/RedHatInsights/insights-inventory-frontend/pull/1621 released to production first.

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -1,31 +1,50 @@
...
-    };
-    return acc;
-}, {});
-
-const compareVersions = (a, b, asc = true) =>
+import { OS_CHIP } from './constants';
+
+export const updateGroupSelectionIdentifier = (selection, major) =>
+    // if every minor version is selected, then mark the group as selected
+    set(selection, [major, major], Object.values({ ...selection[major] }).filter(v => v !== major).every(Boolean));
```
mkholjuraev, line 9: This line of code has multiple loops iterating over selection and major. Is there any way to do this in lesser array iterations? But I like that this looks readable.

src/Utilities/OperatingSystemFilterHelpers.js
```diff
@@ -89,3 +87,18 @@ export const buildOSChip = (operatingSystemValue = {}, operatingSystems) => {
...
+export const onOSFilterChange = (event, selection, clickedGroup, clickedItem) => {
+    const newSelection = Object.assign({}, selection);
+    const value = newSelection[clickedGroup.value][clickedItem.value];
+    const major = clickedGroup.value;
+
+    if (clickedItem.value === major) {
+        // group checkbox clicked => update all minor version selections
+        newSelection[major] = mapValues(newSelection[major], () => value);
+    } else {
+        newSelection[major][major] = Object.values(omit(newSelection[major], major)).every(Boolean);
```
mkholjuraev, line 100: Can we avoid having an object property with the an inner property with the same name?

gkarat: @mkholjuraev , why?

src/Utilities/OperatingSystemFilterHelpers.js
gkarat: @mkholjuraev, your suggestion does makes sense, however I would vote for keeping this initial version. Optimizing it by reducing cycles is a good thing, but when it comes to arrays with ~10 values at maximum, it doesn't make any difference in performance. Also I find this more readable instead of doing it in one cycle with complex OR/AND chaining and using of `reduce`... I hope I understood you  right

src/Utilities/OperatingSystemFilterHelpers.js
mkholjuraev: There might be some benefits of a current object that you are aware of, but I am not. However, for example, this seems to be not a very clean object. 

`const object = {  objectKey: { objectKey: anotherValue } }`
 Ignore if this does not make sense or if changing this takes a lot of effort

gkarat: The thing is, I needed to come up with some kind of data structure that could reflect all the selection cases also given the property format of PrimaryToolbar... I mean, yeah, this is complex and without the context this looks totally weird, but I tried my best to write this business logic explicitly and some comments should also be suggesting why this works like that...

johnsonm325 (approved):

PR #1639, opened 2022-09-22
fix(Tags): show all tags for all systems

mkholjuraev:
fixes: https://issues.redhat.com/browse/ESSNTL-3427

Removes all impacting filters from the tags list in the tag filter.

### DIFF
--- a/src/Utilities/TagsModal.js
+++ b/src/Utilities/TagsModal.js
@@ -8,7 +8,6 @@ import debounce from 'lodash/debounce';
 import flatten from 'lodash/flatten';
 
 const TagsModal = ({
-    customFilters,
     filterTagsBy,
     onToggleModal,
     onApply,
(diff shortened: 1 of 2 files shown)


gkarat (changes requested):
It looks like the tags filter is loading infinitely in *all* consumer applications that I could check, and also in Inventory in some cases: if you go to Inventory from any of the pages (e.g., 1. open Inventory, 2. check that the filter is working, 3. go to, e.g., Advisor and then navigate to Inventory again, 4. the filter is constantly loading). The issue is not present at console.stage.redhat.com/beta.

изображение

Also, it looks like in some circumstances, the filter windows is wider than other boxes (see the screenshot below). I remember this should have been fixed recently, and I also don't see this visual bug at console.stage.redhat.com/beta.

изображение

computercamplove:
@gkarat this visual bug still present on console.stage.redhat.com/beta - https://issues.redhat.com/browse/ESSNTL-2639

Fewwy (approved):
Need to fix the visual bugs in the other PR though. :) Other than that LGTM

mkholjuraev:
@Fewwy thanks for the review!

PR #1642, opened 2022-09-26
chore(InventoryTable): remove unused filters param

gkarat:
I have recently identified that InventoryTable has a "filters" parameter which is not utilized by the components nor any of the children components. I propose to remove this from the propTypes - a tiny contribution to reduce the components complexity/misleading interface.

### DIFF
--- a/src/components/InventoryTable/EntityTableToolbar.js
+++ b/src/components/InventoryTable/EntityTableToolbar.js
@@ -369,7 +369,6 @@ EntityTableToolbar.propTypes = {
     hasAccess: PropTypes.bool,
     filterConfig: PropTypes.object,
     total: PropTypes.number,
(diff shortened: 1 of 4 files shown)


mkholjuraev (approved):
LGTM! 

According to docs, this is useful for customizing the default filters in the InventoryTable, but I think we have enough tools for the same purpose (`filterConfig`, `customFilters`) and no consumer app is using the `filters` prop. Let's remove this.

gkarat:
Yes, it's just no used and even if you pass anything to it, it will do nothing. I was wondering that PrimaryToolbar can anyhow utilize this, but https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/doc/primary_toolbar.md says it doesn't expect it

PR #1644, opened 2022-09-27
Render "No versions available" if there are no OS versions available

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3341 and has to be merged only after https://github.com/RedHatInsights/insights-inventory-frontend/pull/1638.

This makes the systems table render OS filter with the single empty value "No versions available" if there is a network request error or the request is successful but doesn't have any values (empty array). 

## Screenshots

Screenshot 2022-09-27 at 15-43-15 Inventory Red Hat Insights


gkarat:
@katierik, could you please verify that the message is OK and makes sense?

Fewwy (approved):
Fetched it and ran it locally - didn't find any issues. Great job!

PR #1652, opened 2022-10-05
fix(ESSNTL-726): move change of display name to default details reducer

karelhala:
### Description

When changing display name in detail of any application the change is not reflected. This PR fixes such issue by moving the reducers around. Previously the change was applied just to the inventory application, that was incorrect. Now it's applied accross all apps by using the same reducers in the default reducers.

#### Sidenote

While checking this change in edge application I noticed the app was using `systemProfileStore` from general info package, this package is deprecated and should not be used anymore so I exported the profile store via federated modules.


bastilian (approved):
LGTM! Thank you @karelhala!

PR #1661, opened 2022-10-12
fix(InventoryTable): ESSNTL-3415 - Make defaultColumns a function

bastilian:


src/components/InventoryTable/EntityTable.js
```diff
@@ -66,21 +66,20 @@ const EntityTable = ({
         onSort?.({ index, key, direction });
     };
 
-    const columns = useRef([]);
-    useMemo(() => {
+    const columns =  useMemo(() => {
```
karelhala, line 69: extra space
```suggestion
    const columns = useMemo(() => {
```

karelhala (commented):
Looking good, just one minor fix

src/components/InventoryTable/EntityTable.js
bastilian: Removed.

jiridostal (approved):

PR #1670, opened 2022-10-18
ESSNTL-3456 - Hyperlink titles get into global filter tag and breaks everything

jiridostal:
There are 4 commits in the PR:

- For clickable elements that are used to open modals, don't append params and anchors to the URL
- Removal or routing workaround that was introduced before BZ1958379 was closed
- Modals in the general info page can now be routed directly 
- Test updates


bastilian (approved):
Nice fix! Thank you @jiridostal!

PR #1674, opened 2022-10-21
[inventory-refactor] Refactor inventory detail to simplify the usage

karelhala:
### Description

This is first attempt on simplifying inventory. Let's start with the inventory detail since that's the easiest place to start.

Main focus on this PR is not to pass inventory details trough redux, but much rather trough props. That way we can control which item is active etc.
AppInfo component did not utilized `appList` at all. It just added complexity on already complex system. This PR removes this property. I looked around other applications and non was using this property.

Once we merge this PR we should go trough all inventory apps and switch to start using this approach, once all apps are migrated we can remove some of the wrappers. They are kept as is for now so we can continue in incremental PRs.

### DIFF
--- a/src/components/InventoryDetail/ApplicationDetails.js
+++ b/src/components/InventoryDetail/ApplicationDetails.js
@@ -14,10 +14,11 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
     const history = useHistory();
     const dispatch = useDispatch();
     const searchParams = new URLSearchParams(search);
-    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || appList)
-    .filter(({ isVisible }) => isVisible !== false);
+    const items = useSelector(({ entityDetails }) => {
+        return (entityDetails?.activeApps || appList || []).filter(({ isVisible }) => isVisible !== false);
+    });
     const disabledApps = useSelector(({ systemProfileStore }) => systemProfileStore?.disabledApps);
-    const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp?.appName || items[0].name);
+    const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp?.appName || items?.[0]?.name);
     const [activeTabs, setActiveTabs] = useState(items);
 
     useEffect(() => {
--- a/src/components/InventoryDetail/ApplicationDetails.js
+++ b/src/components/InventoryDetail/ApplicationDetails.js
@@ -14,14 +14,11 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
     const history = useHistory();
     const dispatch = useDispatch();
     const searchParams = new URLSearchParams(search);
-    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || [])
+    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || appList)
     .filter(({ isVisible }) => isVisible !== false);
-    const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp);
     const disabledApps = useSelector(({ systemProfileStore }) => systemProfileStore?.disabledApps);
-    const defaultApp = activeApp?.appName || appList?.find(({ pageId, name }) => items?.[0]?.name === (
-        pageId || name))?.name || items?.[0]?.name;
-    let applications = appList || items;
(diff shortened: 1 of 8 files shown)


src/modules/AsyncInventory.js
```diff
@@ -23,7 +24,7 @@ const AsyncInventory = ({ component, onLoad, store, history, innerRef, ...props
     return (
         <RBACProvider appName="inventory">
             <Provider store={store}>
-                <Router history={history}>
+                <Router history={history} basename={getBaseName(window.location.pathname)}>
```
mkholjuraev, line 27: To my knowledge: why do we need a router here?

mkholjuraev (approved):
LGTM! everything seems to work as intended. Let's merge this as the initial step of refactoring.

mkholjuraev:
@karelhala One thing caught my attention when I run this PR with other apps (vuln, patch) I see the missing component message. Should we merge this anyway, and work on those apps to get rid of AppInfo? Or can this be fixed in this PR?

image

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -14,14 +14,11 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
     const history = useHistory();
     const dispatch = useDispatch();
     const searchParams = new URLSearchParams(search);
-    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || [])
+    const items = useSelector(({ entityDetails }) => entityDetails?.activeApps || appList)
```
mkholjuraev, line 17: The issues, I wrote as a comment can be solved by following, this also removes activeApps in redux completely:  
```suggestion
    const items =  (appList || [] )
```

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -14,14 +14,11 @@ const ApplicationDetails = ({ onTabSelect, appList, ...props }) => {
...
     .filter(({ isVisible }) => isVisible !== false);
-    const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp);
     const disabledApps = useSelector(({ systemProfileStore }) => systemProfileStore?.disabledApps);
-    const defaultApp = activeApp?.appName || appList?.find(({ pageId, name }) => items?.[0]?.name === (
-        pageId || name))?.name || items?.[0]?.name;
-    let applications = appList || items;
-    const [activeTabs, setActiveTabs] = useState(applications);
+    const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp?.appName || items[0].name);
```
mkholjuraev, line 20:
```suggestion
    const activeApp = useSelector(({ entityDetails }) => entityDetails?.activeApp?.appName || items.length && items[0].name );
```

mkholjuraev (changes requested):

mkholjuraev:
@karelhala I wanted to add those changes ^ required, but I could not. Would you mind updating this PR so that I can proceed?

src/components/InventoryDetail/ApplicationDetails.js
karelhala: Aaah, yeah! This needs safe accessor.

src/components/InventoryDetail/ApplicationDetails.js
karelhala: It'll be better to use safe accessor when selecting active app. But good point!

karelhala:
@mkholjuraev sure thing! I dug around and found one place in the wrapper where I used incorrect component maper. Should be fixed and working again.

src/modules/AsyncInventory.js
karelhala: The details sometimes rely on router, like clicking on link in gen info and showing modal. However this was because we were not sharing router. Let me give it a try without the router (it would at least ease up the upgrate to react v 6)

karelhala:
@mkholjuraev the PR should be good to go. I've tested in various places and looks like nothing breaks. I've also added a way of picking specific component based on new features so we can hurry up with the react router upgrade. We should start sending the `inventoryId` and `oomponentMapper` with `activeApp`. Once that is done we can start removing react-router-dom dependencies from detail and we should be good to go with new router changes.

mkholjuraev (approved):
LGTM! I will polish mine as well and will address making components easier to undertand&maintain.

PR #1675, opened 2022-10-21
Fix clicking on gen info detail links

karelhala:
### Description

When opening new tab from General information links the UI breaks because inventory ID is incorrectly calculated. This PR fixes it by consuming `inventoryId` from parent component (which should know about what is active by checking the router param). We should go around all apps and adapt to this new behavior, once that is done we can remove the wrapper component.

### DIFF
--- a/src/components/GeneralInfo/BiosCard/BiosCard.test.js
+++ b/src/components/GeneralInfo/BiosCard/BiosCard.test.js
@@ -6,6 +6,16 @@ import BiosCard from './BiosCard';
 import configureStore from 'redux-mock-store';
 import { biosTest } from '../../../__mocks__/selectors';
 
+jest.mock('react-router-dom', () => ({
+    ...jest.requireActual('react-router-dom'),
+    useLocation: () => ({
+        pathname: 'localhost:3000/example/path'
+    }),
+    useHistory: () => ({
+        push: () => undefined
+    })
+}));
+
 describe('BiosCard', () => {
     let initialState;
     let mockStore;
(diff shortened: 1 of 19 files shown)


gkarat (commented):
Could you please explain why we append `flag`, `installed_packages`, `repositories`, etc. to the end of the path name for the links under this tab? Opening a new tab with these links render absolutely the same content as it would be if I open the inventory details with `inventory/id`.

This also leads to a strange URL behavior if you narrow multiple times and the URL is just further appending these keywords. For example, open any system details, click with `Open a new tab` on flags, then in the new tab click on `IPv4 addresses` -> you will get `inventory/ID/flags/ipv4` 🤔

gkarat:
I would love to see Inventory, once I open a new tab with any of the params link, to render the same page but also open a smaller window with the parameter details. As it normally does if you stay on the same page and click with the left mouse button.

karelhala:
@gkarat should be fixed! I've added control over route, so now when user clicks on a link the route changes and based on the route changed if some modal is found it will open that modal. This way users can bookmark their opened modals without opening new tab.

gkarat (approved):
Yeah, now it works, I could reproduce this behavior locally, good job! I also couldn't find any issues in the PR itself, LGTM

gkarat:
However, there are tests failing. And it might be related to the fact that you introduced a wrapper around `InventoryDetail`, and something is broken with router and location parameter:
```
    TypeError: Cannot read properties of undefined (reading 'location')
```

karelhala:
@gkarat yeah. The failing tests are because of the router functions used in those components. Gotta wrap them in Memory router to fix this issue.

gkarat (approved):

PR #1676, opened 2022-10-21
[inventory-refactor] Remove inventory connector

karelhala:
### Description

Since we are no longer dealing with inventory connector from app itself, but rather regular components we don't need `inventoryConnector` anymore. This PR removes it and also adjusts the inventory modules to pass component instead of `componentName` which contains the component used later if user has access.


gkarat (commented):
https://github.com/karelhala/insights-inventory-frontend/blob/remove-inventory-connector/src/routes/InventoryTable.js#L17 still imports `inventoryConnector` which leads to an error.

karelhala:
@gkarat thank you! Was wondering why this was failing. Fixed it, while doing so I also noticed some missmatch of propTypes, so I fixed that as well.

Btw this PR fixes the infamous `Can't perform a React state update on an unmounted component` and should also help fixing long loading states inside of any app.

src/routes/InventoryTable.js
```diff
@@ -188,7 +183,11 @@ const Inventory = ({
                 <Grid gutter="md">
                     <GridItem span={12}>
                         {
-                            InvCmp && <InvCmp
+                            <AsyncInventory
```
bastilian, line 186: The AsyncInventory has providers for RBAC, redux and react-router, could that cause issues when using it here?

bastilian: Could we use just the `RenderWrapper` instead here?

karelhala: On contrary, it should help the inventory to show proper screens when user has no access.

mkholjuraev:
@karelhala both my PR and this PR suffer from this problem. Inventory app loads as expected from the direct URL, but if a user moves to some other app and comes back, the store gets lost and the app breaks with this:
image

karelhala:
@mkholjuraev good finding! I've checked the app and looks like classic issue of re defining redux store. Or rather not re-defining on each mount. I've removed the usage of redux registry as this utility can lead to these kind of problems and instead of that I used redux build ins functions to replace reducers on the fly.

src/AppEntry.js
```diff
@@ -1,29 +1,26 @@
...
 const InventoryApp = () => {
-    const [registry, setRegistry] = useState();
-    const store = registry?.getStore();
-
-    useEffect(() => {
-        setRegistry(IS_DEV ? init(logger) : init());
-
-        return () => {
-            setRegistry(undefined);
+    const registry = useMemo(() => {
```
bastilian, line 11: Why remove using the state? Memo as far as i understand it can be thrown away any time React likes to, or?

karelhala: The difference here is that `useMemo` doesn't generate new re-render. However using `useEffect` - `useState` will cause new render right after the initial mount.

mkholjuraev: I like lesser re-render. I do not see any issues with useEffect which has an empty dependency array. I did some googling.

Ref: https://beta.reactjs.org/reference/react/useMemo#my-calculation-runs-twice-on-every-re-render

mkholjuraev: I like this approach let's fix the tests and merge if everything works as expected. I want to add some of my changes as well :)

src/routes/InventoryTable.js
```diff
@@ -199,12 +196,12 @@ const Inventory = ({
                 <Grid gutter="md">
                     <GridItem span={12}>
                         {
-                            InvCmp && <InvCmp
-                                history={history}
-                                store={store}
+                            <RenderWrapper
```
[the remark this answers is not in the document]
karelhala: Good point, that way we can protect the whole route instead of just part of the components

mkholjuraev (approved):
LGTM! runs as expected.

src/AppEntry.js
bastilian: But we want it to re-render, do we not? It is also a bit iffy that the registry in the memo is not under our, but React's control.

mkholjuraev: Considering your concern about useMemo's unwanted recalculating behavior, I do not see anything that can lead to it in the react docs or anywhere [React caveats](https://beta.reactjs.org/reference/react/useMemo#caveats). Maybe, I am well wrong and happy to learn.

mkholjuraev:
@karelhala it would be great if we can finish up this PR :)

src/AppEntry.js
karelhala: @bastilian the other way around actually. We don't want it to re-render, in the initial render we calculate the value of provider and don't need additional renders.

karelhala:
@mkholjuraev I have no objections to merging this. @bastilian anything to bring up/change so we can merge this?

mkholjuraev:
@bastilian I would like to merge this to continue with where I left off if you do not have any objections.

src/AppEntry.js
bastilian: @karelhala "re-render" as in ensure that a value is set. In this case the advice the advice from the React docs to only use `useMemo` as a performance optimisation holds up here as well, as without using the useMemo the app would not work properly and has a chance of fail. This is why I feel a`useState` is more appropriate here and elsewhere where we need to have a registry for child components.

karelhala: @bastilian both `useEffect` with `useState` has same outcome as `useMemo`. We have to create the redux store only on initial render. This can be achieved either by using `useEffect` and setting the value in `useState, however this causes new render which can be optimized using `useMemo`. Basically `useEffect` is good for side effects, like querying server, dispatching new actions to redux. Whereas `useMemo` is used to memoize the value between renderes, meaning we run this code once on initial render, rember the value and not trigger new render. This approach has been used for instance in advisor https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/AppEntry.js#L13 because previous way (setting on each render) was causing re-creating redux store.

Some examples: https://medium.com/@zoenkatz/usememo-vs-usestate-and-useeffect-d906901d88c https://angeloocana.com/en/blog/react/why-usememo-instead-of-usestate-useeffect/

bastilian: @karelhala The thing here is that the instance of the registry/store has to stay exactly the same, which I don't trust `useMemo` can guarantee.

I'm fine with using the `useMemo` here, but I do not trust it and if there are any issues related to this, we should revert this to be in a state where the initialisation is only controlled by us.

mkholjuraev: How would useState make sure to have the same instance across page re-visits? I would think it would create a new instance from scratch.

Having `useState` and React context together to have the same instance always produced the same issue, as far as I understand.

mkholjuraev: @bastilian @karelhala merging this so that we can test it on stage env. In case there is any issue, it is not difficult to change `useMemo` to `useState`. If this is totally wrong, please ping me and I will revert.

PR #1697, opened 2022-11-08
[ESSNTL-3600] Set name filter as default filter

johnsonm325:
A previous change removed `value` from the operating system filter config for PrimaryToolbar. ConditionalFilter needs this to determine which filter is currently being used. A side-effect was that if one of the filters didn't have the `value` set, it would set it as the default filter.

The previous execution of the code was a bit verbose, and I don't think it was needed. It seems you can hard-code the value in and everything is handled normally.


gkarat (approved):
Looks good to me. Run locally, and I can see the default filter is Name again.

PR #1710, opened 2022-11-29
feat(ESSNTL-3576): introduces system update method filter

mkholjuraev:
Resolves: [ESSNTL-3576](https://issues.redhat.com/browse/ESSNTL-3576)

This introduces a new 'System update method' filter in the inventory table.

To test:

1. Visit the inventory page
2. Observe that the new filter is added to the filters dropdown
3. The filter works as expected. Sends a request as described in the ticket  (e.g `filter[system_profile][system_update_method][eq]=yum`).
4. Observe that the filter persists across refresh

### DIFF
--- a/src/components/filters/useUpdateMethod.test.js
+++ b/src/components/filters/useUpdateMethod.test.js
@@ -1,0 +1,55 @@
+/* eslint-disable react/prop-types */
+import React, { Fragment, useEffect } from 'react';
+import { useUpdateMethodFilter } from './useUpdateMethodFilter';
+import { updateMethodOptions, UPDATE_METHOD_KEY } from '../../Utilities';
+import { mount } from 'enzyme';
+
+const HookRender = ({ hookAccessor, hookNotify }) => {
+    const [filter, chip, value, setValue] = useUpdateMethodFilter();
+    useEffect(() => {
+        hookAccessor && hookAccessor([filter, chip, value, setValue]);
+    }, []);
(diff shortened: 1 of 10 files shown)


src/components/filters/useUpdateMethod.test.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable react/prop-types */
+import React, { Fragment, useEffect } from 'react';
+import { useUpdateMethodFilter } from './useUpdateMethodFilter';
+import { updateMethodOptions, UPDATE_METHOD_KEY } from '../../Utilities';
+import { mount } from 'enzyme';
```
bastilian, line 5: We should avoid `enzyme`. The `@testing-library/react-hooks` is already included in inventory and can be easily used here.

mkholjuraev: @bastilian thank you for having a look. Actually, `enzyme` is used in almost all filter tests, thus I prefer to get rid of it at once later. I would like to refactor the filtering in Inventory, while refactoring I will get rid of enzyme.

mkholjuraev: There is already a ticket to optimise filtering: [RHIF-112](https://issues.redhat.com/browse/RHIF-112)

bastilian (approved):
Works as expected! Thank you @mkholjuraev!

PR #1715, opened 2022-12-01
Stale fitler

mkholjuraev:
Resolves: [ESSNTL-3914](https://issues.redhat.com/browse/ESSNTL-3914)

This removes default filtering completely while adding the 'Unknown' option into the status filter to be able to filter by all 4 options.

To test:

1. Visit the inventory app
2. observe that there is not any active filter by default
3. open the status filter, and observe that there is a new option 'Unknown'
4. play with the filtering: should persist refresh, should be added into URL, should filter as expected

### DIFF
--- a/src/store/entities.js
+++ b/src/store/entities.js
@@ -20,7 +20,6 @@ import TitleColumn from '../components/InventoryTable/TitleColumn';
 import InsightsDisconnected from '../Utilities/InsightsDisconnected';
 import OperatingSystemFormatter from '../Utilities/OperatingSystemFormatter';
 import { Tooltip } from '@patternfly/react-core';
-import { defaultFilters } from '../Utilities/constants';
 import { verifyCulledInsightsClient } from '../Utilities/sharedFunctions';
 
 export const defaultState = {
@@ -107,7 +106,7 @@ function entitiesPending(state, { meta }) {
 function clearFilters(state) {
     return {
         ...state,
-        activeFilters: [defaultFilters]
+        activeFilters: []
     };
 }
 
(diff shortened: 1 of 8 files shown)


adonispuente:
Filtering is working correctly in most cases, but not all. When selecting [Fresh,Stale warning] as filters, you actually get more results than when you have no filter. 

It sends all the filters in the payload when these two are selected. Attached is a video 

https://user-images.githubusercontent.com/60629070/205320314-55d3fef4-7950-4502-b3c6-77643a5c9a33.mp4

mkholjuraev:
Great catch! It looks like the API also has all stale_warning systems filtered out by default. I will ask the backend to return all when there is no filter.

src/store/entities.js
```diff
@@ -107,7 +106,7 @@ function entitiesPending(state, { meta }) {
 function clearFilters(state) {
     return {
         ...state,
-        activeFilters: [defaultFilters]
+        activeFilters: []
```
bastilian, line 109: We should be very careful about changing things in the the redux store, since some applications are expecting certain things to be in there and those would need to be reworked before. Same for below.

bastilian: @mkholjuraev Thank you for double checking!

mkholjuraev:
@RedHatInsights/team-interact  Considering this issue is already reported, I am asking for a review again to deliver the ticket.

Fewwy (approved):
Tested it locally, LGTM
I think in the future we should add scalable cypress tests to test filter chips. We did that in OCP Advisor

mkholjuraev:
Merging, I am writing a follow-up mail to hcc-tech-lict

PR #1740, opened 2023-01-16
Optimisations

mkholjuraev:
This removes loading Inventory module from InventoryConnector. Instead it loads the module directly into the component. Also there is some cleanup as well. 

The metrics are varying across runs, but I am getting average numbers. Here you can see the effect of this optimisation work. Most improved metric is LCP which has increased from ~7.5 seconds to ~ 4 seconds. 

Befor the work: 
withoutOptimizationCommit

After the work:
withOptimizationCommit

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,8 +1,8 @@
 import { Route, Redirect, Switch } from 'react-router-dom';
 import React, { lazy, Suspense } from 'react';
-import { tagsMapper } from './constants';
-import { RHCD_FILTER_KEY, UPDATE_METHOD_KEY } from './Utilities/constants';
+import { getSearchParams } from './constants';
 import RenderWrapper from './Utilities/Wrapper';
+
 const InventoryTable = lazy(() => import('./routes/InventoryTable'));
 const InventoryDetail = lazy(() => import('./routes/InventoryDetail'));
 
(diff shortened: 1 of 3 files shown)


src/constants.js
```diff
@@ -112,3 +113,17 @@ export const INVENTORY_WRITE_PERMISSIONS = [
     'inventory:hosts:write',
     'inventory:*:write'
 ];
+
+export const getSearchParams = () => {
```
mkholjuraev, line 117: If I had time, I would do a larger refactoring of filtering in the Inventory App. but for now, this is bearable.

src/Routes.js
```diff
@@ -21,25 +22,19 @@ export const Routes = () => {
...
-                        rhcdFilter={searchParams.getAll(RHCD_FILTER_KEY)}
-                        updateMethodFilter={searchParams.getAll(UPDATE_METHOD_KEY)}
-                        page={searchParams.getAll('page')}
-                        perPage={searchParams.getAll('per_page')}
-                    />}
+                    render={() =>
+                        <RenderWrapper
+                            cmp={InventoryTable}
+                            isRbacEnabled
+                            {...getSearchParams()}
```
bastilian, line 36: It might make sense to wrap this in a `useMemo`.

mkholjuraev: Good point!

mkholjuraev:
Closing the PR for the sake of already merged PR.

PR #1741, opened 2023-01-16
perf: set perPage to 20 on Inventories table

mkholjuraev:
This sets the default per page of the inventory table from 50 to 20. 

The metrics are varying across runs, but I am getting average numbers. Here you can see the effect of this optimization work. Both LCP and TTI metrics have increased to more than ~1 second

Before:
image

After: 
image


mkholjuraev:
Closing this for the sake of keeping the same amount of data in the table.

PR #1746, opened 2023-01-26
fix: get rid of circular deps to avoid early access registery context

mkholjuraev:
This PR resolves the issue **Cannot access '__WEBPACK_DEFAULT_EXPORT__' before initialization** on stage-beta after a hard refresh on inventory pages on consumer apps. The culprit was circular deps between store/index.js and all of the activeApp components in the reducer context.

I will get rid of activeApps in the reducer as a next step improvement.

To test:
1. Please run the app with other consumer apps (Patch, vuln...)
2. Verify that all apps work together 
3. Hard refresh does not break any app
4. Inventory detail page on Inventory app works for all tabs.

Circular dep before:
image

Circular dep after:
image

### DIFF
--- a/src/components/InventoryTable/InventoryTable.test.js
+++ b/src/components/InventoryTable/InventoryTable.test.js
@@ -10,22 +10,8 @@ import { createPromise as promiseMiddleware } from 'redux-promise-middleware';
 import { BrowserRouter as Router } from 'react-router-dom';
 import toJson from 'enzyme-to-json';
 import { ConditionalFilter } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
-import * as actions from '../../Utilities/constants';
+import * as loadSystems from '../../Utilities/sharedFunctions';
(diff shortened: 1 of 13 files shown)


src/components/InventoryTable/InventoryTable.test.js
```diff
@@ -50,7 +36,8 @@ describe('InventoryTable', () => {
                 }
             }
         };
-        spy = jest.spyOn(actions, 'loadSystems').mockImplementation(() => ({ type: 'reload' }));
+        console.log('actionss', loadSystems);
```
karelhala, line 39: Extra `console.log`
```suggestion
```

karelhala (approved):
Codewise looking good!

karelhala:
@mkholjuraev tested locally with Advisor and vulnerability app, looks like everything works as expected. Looks like it even fixes the edge problem!

PR #1750, opened 2023-01-27
fix(ESSNTL-4149): fix applying default filters from URL, with some cl…

mkholjuraev:
Fixes: https://issues.redhat.com/browse/ESSNTL-4149

This PR passes filters as a custom filter to be able to apply default filters from URL. It also includes some cleanup. I have removed a bunch of code that was used to add URL filters. This is possible using customsFilter after this [commit](https://github.com/RedHatInsights/insights-inventory-frontend/commit/d9c18d4cd72c2b1dbdc9751c0278134599744484)

To test:
Please run the app and play with filters, give filters and try refreshing the page. Observe that filters are applied from URL

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,8 +1,8 @@
 import { Route, Redirect, Switch } from 'react-router-dom';
-import React, { lazy, Suspense } from 'react';
-import { tagsMapper } from './constants';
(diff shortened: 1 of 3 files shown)


Fewwy:
@mkholjuraev Tested it locally, filters work as expected, couldn't break them. LGTM

Fewwy (approved):

PR #1754, opened 2023-02-01
perf(RHIF-167): get rid of activeApps in redux

mkholjuraev:
This is the second step of improving the Inventory details page by getting rid of using the redux store to render active apps in the Main section tab. Now, an active application that needs to be displayed will be decided by the `appName` in the URL. If it is not set, the first application in `appsList` prop will get displayed. Also, this PR introduces wrapping general information sections with `Card` component as a result of [this](https://www.sketch.com/s/ee66c23d-3d18-4940-afdf-e8a3c983d22d/a/Qbn7lo7#Comment) UX ask. This PR is backward compatible. There are a few extra components that needs to be removed once all apps migrate to new approach safely. 

Benefits of this refactor:

1. Better tab layout. No need to render the tab menu separate from the tab content.
2. Consumer apps can adopt DetailHeader as a presentation component if it has its own data to populate.
3. AppInfo component is being removed with **activeApps** in redux which will decrease complexity.
4. No redux involved in changing, or rendering the tab. The URL parameter appName is used to change tabs, render the default tab
5. General information cards are separated as a card for better design
6. Inventory header component and Main section where the tab and apps render are decoupled. Now, both can load independently
7. InventoryDetail component truly renders the whole Inventory detail page. Before it would render the header, while main section needed to be rendered using AppInfo
 

Regarding consumer applications dealing with InventoryDetailHead and AppInfo components in the fed-mod, I have been working to remove AppInfo component and passing inventoryId into DetailsWrapper component to directly use DetailHeader as presentational component. This gives those apps more freedom, and reduces the complexity of how we deal with federated modules. In the following commit/PR I will target consuming InventoryDetailHeader and DetailWrapper in fed-modules.

General information tab before:

image

General information tab after:
image


mkholjuraev:
Putting this in the draft state to merge Karel's PR and this together as a unified refactoring approach.

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -60,7 +60,7 @@ class GeneralInformation extends Component {
     };
 
     componentDidMount() {
-        this.props.loadSystemDetail && this.props.loadSystemDetail(this.props.entity.id);
+        this.props.loadSystemDetail && this.props.loadSystemDetail(this.props.inventoryId);
```
karelhala, line 63: I'd change this so we don't make the `inventoryId` required yet (edge for instance doesn't pass it yet)
```suggestion
        this.props.loadSystemDetail?.(this.props.inventoryId || this.props.entity.id);
```

src/components/InventoryDetail/DetailHeader.js
```diff
@@ -0,0 +1,96 @@
...
+        {(loaded && verifyCulledInsightsClient(entity?.per_reporter_staleness)) && <InsightsPrompt />}
+        {children}
+    </Fragment>);
+
+    return (shouldWrapAsPage ?
+        (<PageHeader className={classnames('pf-m-light ins-inventory-detail', additionalClasses)} >
+            {BreadcrumbWrapper}
+            {HeaderInfo}
+        </PageHeader>) : HeaderInfo);
+};
```
karelhala, line 66: It's not a good idea to create new component on each render (especially with so many props). How about we'd split it into two components?
```suggestion

const HeaderInfo = ({ entity, loaded, UUIDWrapper, LastSeenWrapper, children, ...props }) =>  (<Fragment>
        <TopBar
            entity={entity}
            loaded={loaded}
            {...props}
        />
        <FactsInfo
            loaded={loaded}
            entity={entity}
            UUIDWrapper={UUIDWrapper}
            LastSeenWrapper={LastSeenWrapper}
        />
        {(loaded && verifyCulledInsightsClient(entity?.per_reporter_staleness)) && <InsightsPrompt />}
        {children}
    </Fragment>);

const DetailHeader = ({ shouldWrapAsPage, ...props }) => {
    return (shouldWrapAsPage ?
        (<PageHeader className={classnames('pf-m-light ins-inventory-detail', additionalClasses)} >
            {BreadcrumbWrapper}
            <HeaderInfo {...props} />
        </PageHeader>) : <HeaderInfo {...props} />);
};

HeaderInfo.propTypes = DetailHeader.propTypes;
```

I am not sure if I didn't missaligned anything, better to try this locally.

src/components/InventoryDetail/DetailWrapper.js
```diff
@@ -13,17 +13,23 @@ import {
...
     const store = useStore();
     const entity = useSelector(({ entityDetails }) => entityDetails?.entity);
     const isExpanded = useSelector(({ entityDetails }) => entityDetails?.isToggleOpened);
     const loaded = useSelector(({ entityDetails }) => entityDetails?.loaded);
 
+    useEffect(() => {
+        if (!entity || !(entity?.id === inventoryId) || !loaded) {
+            dispatch(loadEntity(inventoryId, { hasItems: true }, { showTags }));
+        }
+    }, []);
```
karelhala, line 31: What if app doesn't send `inventoryId`? Wouldn't this cause constant re-renders? Also, if  `entity` is changed via selector no new re-calculation is not triggered
```suggestion
    useEffect(() => {
        if (!entity || !(entity?.id === inventoryId) || !loaded) {
            dispatch(loadEntity(inventoryId, { hasItems: true }, { showTags }));
        }
    }, [entity, inventoryId, loaded]);
```

src/routes/InventoryDetail.js
```diff
@@ -53,29 +48,25 @@ const activeApps = [
...
     const entity = useSelector(({ entityDetails }) => entityDetails?.entity);
     const clearNotifications = () => dispatch(actions.clearNotifications());
+
     useEffect(() => {
         chrome?.hideGlobalFilter?.(true);
         chrome.appAction('system-detail');
         clearNotifications();
-        const appName = searchParams.get('appName');
-        if (appName) {
-            dispatch(detailSelect(appName));
```
karelhala, line 72: I remember removing this action caused incorrect tab activation. I guess the new refs deal with this, right?

src/routes/InventoryDetail.js
```diff
@@ -86,63 +77,49 @@ const Inventory = () => {
         insights?.chrome?.appObjectId?.(entity?.id);
     }, [entity?.id]);
 
+    const onTabSelect = useCallback((_, activeApp, appName) => {
+        searchParams.set('appName', appName);
+        const search = searchParams.toString();
+        history.push({
+            search
+        });
+    }, []);
```
karelhala, line 86: You should add at least to list of dependencies `searchParams`.

karelhala (commented):
Looking good! Just a few minor questions and adjustments. I'll test it locally later on.

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
mkholjuraev: Aha, I will do so

src/components/InventoryDetail/DetailHeader.js
mkholjuraev: Of course. I missed this caveat. I will adopt the suggestion

src/components/InventoryDetail/DetailWrapper.js
mkholjuraev: I am just thinking about why the entity would change on this page. But, maybe there is something that I am not aware of.

src/routes/InventoryDetail.js
mkholjuraev: Yup, this is resolved by refs

mkholjuraev:
@karelhala Hopefully, everything is as expected now. Here is the edge PR to align the app with this refactoring. Please run these apps to verify. The important thing is that with this PR, inventoryId prop will be used as a main factor to fetch host detail. Thus this prop is marked as required. This enables apps to independently render the header and the main section.

karelhala (approved):
Looking good! Tested it locally and everything seems to be working!

PR #1762, opened 2023-02-08
refactor(ESSNTL-3724): Connect /groups to feature flags

gkarat:
This makes Inventory hide /groups page component if the feature flag is set to false.

## How to test

1. Deploy the PR locally using stage environment.
2. Go to https://insights-stage.unleash.devshift.net/projects/default/features/hbi.ui.inventory-groups.
3. Set the flag to true in default and then check that the /groups page is rendered.
4. Set the flag to false in default and then check that the /groups page is not rendered and return 404 page.


Fewwy:
Checked it locally, works as expected! GOOD JOB!

PR #1763, opened 2023-02-08
chore: extend dependabot checking interval

mkholjuraev:
Let's extend `dependabot` checking interval to a weekly manner. This will most probably not miss any updates. It might just aggregate some updates.


gkarat (approved):
I support this change. The same weekly manner done for other frontends and it makes it less obtrusive.

mkholjuraev:
Let's merge this and see what we will get.

PR #1766, opened 2023-02-10
chore(RHIF-171): pass inventoryId to Patch module as a prop

mkholjuraev:
Passes inventory id as a prop to Patch module. This prop is required after [this PR](https://github.com/RedHatInsights/patchman-ui/pull/956) to render system detail components.

To test please run both PR together and make sure no application breaks. Functionality should not change.


bastilian (approved):
LGTM! Thank you @mkholjuraev!

PR #1767, opened 2023-02-14
feat(ESSNTL-4194): Show groups in a table

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4194.
Mock https://www.sketch.com/s/c19f555b-7887-4423-b55d-575c8dd1dfb7/a/dl8zwyb.

Create the groups list table with three columns: name, total systems and last modified. Implement sorting by each column, filtering by name and pagination.

⚠️  This doesn't implement actions (rename, delete, create group, bulk select) yet (will be in a separate PR).

## How to test

- Deploy locally with mock server (`npm run start:mock`).
- Navigate to /inventory/groups and check that the table is rendered with dynamically generated values.
- You can also disable the mock server and check whether network errors are properly handled with the error state.
- You can also tweak your mock server and generate empty response (to verify the empty state).
- The table is API paginated and filtered. Any filter change, pagination or sorting you make must trigger a new API request.

## Screenshots

image

No match state:
image

Failed request (any runtime error):
image

Pagination values:
image


src/components/GroupsTable/GroupsTable.cy.js
```diff
@@ -0,0 +1,216 @@
...
+import { groupsInterceptors as interceptors } from '../../../cypress/support/interceptors';
+import { getStore } from '../../store';
+import GroupsTable from './GroupsTable';
+
+const ORDER_TO_URL = {
+    ascending: 'ASC',
+    descending: 'DESC'
+};
+
+const DEFAULT_ROW_COUNT = 50;
```
computercamplove, line 32: @gkarat i remember @mkholjuraev raised a conversation about default pagination 20 - should we stick with it?

cypress/support/commands.js
```diff
@@ -22,4 +22,23 @@
 //
 //
 // -- This will overwrite an existing command --
-// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
\ No newline at end of file
+// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+
+Cypress.Commands.add(
+    'ouiaId',
```
Fewwy, line 28: I think this command is available in the frontend-component-utils library for cypress
https://github.com/RedHatInsights/frontend-components/blob/master/packages/utils/src/CypressUtils/CustomCommands.js

mmenestr:
Looks good, just a few comments:
1. Can we reposition the columns so that the "total systems" column is more centered in the table?
2. Can we update the empty state to look like this (just add icon and button) 
image
3. Search should use the "[Search input" component](https://www.patternfly.org/v4/components/search-input) where the icon is on the left
4. For the error empty state, wonder if that action button is really useful (?) is that what we usually use in other places? If so then leave it, but just wondering!

cypress/support/commands.js
gkarat: yeah nice

Fewwy:
When I tried to change the page I encountered this error

gkarat:
@Fewwy,
1. regarding sorting: this is expected, the table sends the `order_by` with the parameter which is not currently listed in the API schema and thus the mock server will throw you error ('name', 'updated_at', `host_ids` are not allowed currently for `order_by` parameter).
2. the same for pagination, the mock server generates crazy values for `count` and `total` fields, thus the requests will be broken too with very huge (forbidden) values.

All these problems will not be reproducible with the real API implementation soon.

Fewwy:
@gkarat Other than that the PR look fine :)

gkarat:
@mmenestr, thanks a lot for your review! I addressed your points in the last commits. For the button in the error state: this would require some work in our library (separate from the inventory frontend): I left the TODO note to fix it later.

src/components/GroupsTable/GroupsTable.js
```diff
@@ -0,0 +1,237 @@
...
+        setRows(newRows);
+    }, [groups]);
+
+    // TODO: convert initial URL params to filters
+
+    const onSort = (event, index, direction) => {
+        setFilters({ ...filters, sortIndex: index, sortDirection: direction });
+    };
+
+    const filterConfigItems = [
```
mkholjuraev, line 108: I am wondering if there is a way to create a more dynamic function to create filters and filter configs... This may be a starting point to improve the filtering on the Inventory table as well.

Now, the SearchInput component is re-created every time on re-render.

src/components/GroupsTable/GroupsTable.js
```diff
@@ -0,0 +1,237 @@
...
+import upperCase from 'lodash/upperCase';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { Link } from 'react-router-dom';
+import { TABLE_DEFAULT_PAGINATION } from '../../constants';
+import { fetchGroups } from '../../store/inventory-actions';
+import { generateLoadingRows } from '../InventoryTable/helpers';
+import NoEntitiesFound from '../InventoryTable/NoEntitiesFound';
+
+const GROUPS_TABLE_INITIAL_STATE = {
```
mkholjuraev, line 30: How about extracting assets into separate file?

src/components/GroupsTable/GroupsTable.js
```diff
@@ -0,0 +1,237 @@
...
+                activeFiltersConfig={activeFiltersConfig}
+            />
+            <Table
+                aria-label="Groups table"
+                ouiaId="groups-table"
+                /* ouiaSafe={!loadingState}> */
+                variant={TableVariant.compact}
+                cells={GROUPS_TABLE_COLUMNS}
+                rows={
+                    uninitialized || loading
```
mkholjuraev, line 186: I think this can be moved to a helper function to have cleaner code. What do you think?

src/components/GroupsTable/GroupsTable.js
gkarat: good point! I think I'll wrap it with useMemo since it looks like a good example where it can become useful

src/components/GroupsTable/GroupsTable.js
gkarat: i'd like to keep it in the same file since these constants are tied to the component's state only

Fewwy (approved):

src/components/GroupsTable/GroupsTable.js
mkholjuraev: Hmm, maybe there will be a new component that can re-use them? But, not a big deal, this is minor :)

mkholjuraev (commented):
Codewise looks good!

PR #1771, opened 2023-02-21
feat(ESSNTL-3726): Add the /groups/%id page (group detail)

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3726.
Mocks https://www.sketch.com/s/c19f555b-7887-4423-b55d-575c8dd1dfb7/a/52E5bap.

This adds a new route `/groups/%id` where `id` is a group id. The page contains the breadcrumbs, header with the group name or id, and two tabs that contain placeholder (table and further components will be implemented soon). The feature is covered with unit and component tests (the component ones primarily check the right behavior under certain API responses).

## How to test

While running the PR locally together with mock server, navigate to `/groups/%id` with any custom id. The page should render expected content.

## Screenshots

image

image


Fewwy:
I think you're missing the group actions dropdown on the top right
https://www.sketch.com/s/c19f555b-7887-4423-b55d-575c8dd1dfb7/a/ygRKL3q

Fewwy:
Other than missing dropdown - page looks good

src/Routes.js
```diff
@@ -51,6 +53,15 @@ export const Routes = () => {
                     }
                     rootClass="inventory"
                 />
+                <Route exact path={routes.groupDetail} component={groupsEnabled
+                    ? InventoryGroupDetail
+                    : () => (
+                        <EmptyState>
```
mkholjuraev, line 59: Is not this component exist on FEC with an already empty state? If this is the component for 404, it should be there, ready for consumption. Also, maybe we can extract this into some wrapper and avoid repetition?

src/components/InventoryGroupDetail/InventoryGroupDetail.js
```diff
@@ -0,0 +1,62 @@
...
+                    inset={{ default: 'insetMd' }} // add extra space before the first tab (according to mocks)
+                >
+                    <Tab eventKey={0} title={<TabTitleText>Systems</TabTitleText>} aria-label="Group systems tab">
+                        <PageSection>
+                            <GroupDetailSystems />
+                        </PageSection>
+                    </Tab>
+                    <Tab eventKey={1} title={<TabTitleText>Group info</TabTitleText>} aria-label="Group info tab">
+                        <PageSection>
+                            <GroupDetailInfo />
```
mkholjuraev, line 48: How feasible is it to load the second tab lazily? It would save API, rendering time. If the second tab is static or makes a fast request, you can ignore this :)

src/Routes.js
gkarat: Didn't found the empty state-like components that could replace it. if you could point me out to any would be cool. meanwhile, I extracted this empty state to a separate component in the last commit 👍🏼

mkholjuraev (approved):
Codewise looks good to me!

PR #1772, opened 2023-02-22
feat(invntory groups): added new group column behind feature flag

Fewwy:
This pr is adding a Group column to the Inventory table
I had to create a new array of default columns and wrap it in the feature flag
image


src/store/entities.js
```diff
@@ -35,7 +35,98 @@ export const defaultState = {
...
+                </CullingInformation>
+            ) : (
+                new Date(value).toLocaleString()
+            );
+        },
+        props: { width: 10 }
+    }
+];
+
+export const defaultColumns = () => [
```
gkarat, line 129: You don't have to keep two distinct arrays. Instead, try the conditional spread operator:

e.g., `[{ a: 42 }, ...(flag ? [{ test: 'abc' }] : []), { foo: 'bar' }]` for falsey `flag` will result in
```
[
  {
    "a": 42
  },
  {
    "foo": "bar"
  }
]
```
and for truthy:
```
[
  {
    "a": 42
  },
  {
    "test": "abc"
  },
  {
    "foo": "bar"
  }
]
```

gkarat (changes requested):

src/store/entities.js
Fewwy: Fixed that

src/store/entities.js
```diff
@@ -21,33 +21,44 @@ import InsightsDisconnected from '../Utilities/InsightsDisconnected';
...
+    groups: [],
     invConfig: {},
     sortBy: {
         key: 'updated',
         direction: 'desc'
     }
 };
 
-export const defaultColumns = () => ([
+export const defaultColumns = (groupsEnabled = useFeatureFlag('hbi.ui.inventory-groups')) => [
```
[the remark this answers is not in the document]
gkarat: And then in the component where you call `defaultColumns`, call the hook there and only then pass the parameter to this function.

gkarat (changes requested):

src/store/entities.js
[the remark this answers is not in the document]
Fewwy: Yep, that's a better solution

src/components/InventoryTable/hooks/useColumns.js
```diff
@@ -16,13 +18,13 @@ const useColumns = (columnsProp, disableDefaultColumns, showTags, columnsCounter
     );
     const disabledColumns = Array.isArray(disableDefaultColumns) ? disableDefaultColumns : [];
     const defaultColumnsFiltered = useMemo(() => (disableDefaultColumns === true) ?
-        [] : defaultColumns().filter(({ key }) =>
+        [] : defaultColumns(groupsEnabled).filter(({ key }) =>
             isColumnEnabled(key, disabledColumns, showTags)
         ), [disabledColumns, disableDefaultColumns, showTags]);
```
gkarat, line 23: We need to add `groupsEnabled` to these deps array so that the value is updated when the flag changes.

gkarat (changes requested):

src/components/InventoryTable/hooks/useColumns.js
Fewwy: Ok

gkarat (approved):
good job

PR #1774, opened 2023-02-23
chore: update inventory detail docs, handle system not found more nicely

mkholjuraev:
Updates inventory detail docs, and hides the main section if the system does not exist.

To test the main section related change:
1. Open one of the oldest systems (we need to choose a system that is not needed)
2. Click on the Delete button on the top right side.
3. Refresh the page. You should see the old nicely designed system not found component.


adonispuente (approved):
PR works as intended. Observed no issues when testing against other apps as well

PR #1775, opened 2023-02-28
feat(ESSNTL-4365): Select rows in the groups table

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4365 (1/2, the actions will be implemented in a separate PR).

This makes possible to select rows in the groups table separately or in bulk. Component tests are implemented to verify the correct request URLs. The selection drop down shows three options according to the UX tables audit document (Stefan Kukla).

## How to test

1. Run `npm run start:mock` and `npm run mock-server` in separate terminals.
2. Navigate to https://stage.foo.redhat.com:1337/insights/inventory/groups.
3. Mock server sends not consistent responses, this is why the selection will be working with bugs, but you can still play with it on the page.
4. Run `npm run test:openct` and find the new tests. You can play with the table there too.

## Screenshots

image


src/Utilities/hooks/usePromiseQueue.js
```diff
@@ -0,0 +1,33 @@
+import { useCallback, useState } from 'react';
+import pAll from 'p-all';
+
+const DEFAULT_CONCURRENT_PROMISES = 2;
+
+const usePromiseQueue = (limit = DEFAULT_CONCURRENT_PROMISES) => {
+    const [isResolving, setIsResolving] = useState(false);
```
mkholjuraev, line 7: How about merging these 2 states? This would make the code cleaner and also as far as I remember state updates in a promise do not get batched (reference: https://stackoverflow.com/questions/53048495/does-react-batch-state-update-functions-when-using-hooks).

src/components/GroupsTable/GroupsTable.js
```diff
@@ -208,6 +246,47 @@ const GroupsTable = () => {
                 }}
                 filterConfig={{ items: filterConfigItems }}
                 activeFiltersConfig={activeFiltersConfig}
+                bulkSelect={{
+                    items: [
```
mkholjuraev, line 250: A minor: I think if we move this bulk select config object outside of the JSX rendering section, we would have even cleaner code. We can also separate this into a callback as the function definition does not change often.

mkholjuraev (commented):
Overall, looks good as always. Just a minor suggestions that I though deserving

Fewwy:
select none works correctly
select/deselect page works correctly
select all didn't work for me(possibly because of the crazy number)
image

src/Utilities/hooks/usePromiseQueue.js
gkarat: nice suggestion, done in the last commit.

PR #1776, opened 2023-02-28
DRAFT feat(invntory groups): Inv table group filter

Fewwy:


gkarat (commented):
overall, LGTM, but needs a bit of more improvement. also, found out I cannot deselect the group from the toolbar, video:

[Screencast from 03-14-2023 11:01:14 AM.webm](https://user-images.githubusercontent.com/31385370/224968871-e2ed4952-d8fb-487f-8d70-f349ad5e79af.webm)

gkarat:
Just a note: it seems that after applying the group filter, the network request doesn't contain group_name parameter. It's gonna be added to spec here https://github.com/RedHatInsights/insights-host-inventory/pull/1285/files#diff-6b7977bd0eb33f42b9b33e2725d80da5d6618c59f96da50254d3fe660703c594R131-R133. Given that it's not implemented, let's omit it in this phase (this PR). I will create a separate Jira that will make sure we append this parameter in the requests.

gkarat (approved):
looks legit 👍🏼 approving

PR #1777, opened 2023-02-28
fix(ADIVOSR-3016): Add perPage arg to getEntities

adonispuente:
[ADVISOR-3016](https://issues.redhat.com/browse/ADVISOR-3016) describes a bug where the second half of items when loading in more than 50 (the default) items from inventory dont have the rhel version attached. This was because the inventory end point could never get all 100 as the arguments are filled in as undefined. This PR just allows it to get passed in.

Tested against Patch, Inventory, and Advisor.

**How to Reproduce:**
1. Go to console.redhat.com
2. Advisor -> Recommendations, select one advisor with more than 100 affected systems
3. Select 100 per page

**Expected results:**
All hosts should show the respective OS versions.


mkholjuraev (approved):
LGTM! 

Just one reminder, there might be some applications that might get affected by this. I had a similar case with the ROS app. So we need to watch out or make sure this will not have a negative affect on other apps.

adonispuente:
@mkholjuraev thanks for taking a look. I'll keep that in mind and be more cautious releasing to the other envs

PR #1781, opened 2023-03-03
feat(ESSNTL-4056): Add lastSeen filter

adonispuente:
[ESSNTL-4056](https://issues.redhat.com/browse/ESSNTL-4056)
This PR implements a new date range filter in inventory. The mvp for the filter is as follows :

- User can use the "Last seen" filter to select a date range.
- If a start date is selected, "updated_start" should be added to the GET /hosts query.
- If an end date is selected, "updated_end" should be added to the GET /hosts query.
- If no value is selected for the filter, neither "updated_start" nor "updated_end" should be added to the GET /hosts query, and the host results should remain unchanged.

### DIFF
--- a/src/components/InventoryTable/EntityTableToolbar.js
+++ b/src/components/InventoryTable/EntityTableToolbar.js
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-import React, { Fragment, useEffect, useCallback, useReducer } from 'react';
+import React, { Fragment, useEffect, useCallback, useReducer, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import xor from 'lodash/xor';
@@ -19,7 +19,8 @@ import {
     TAG_CHIP,
     arrayToSelection,
     RHCD_FILTER_KEY,
-    UPDATE_METHOD_KEY
+    UPDATE_METHOD_KEY,
+    LAST_SEEN_CHIP
 } from '../../Utilities/index';
 import { onDeleteFilter, onDeleteTag } from './helpers';
 import {
@@ -28,6 +29,7 @@ import {
     useRegisteredWithFilter,
     useTagsFilter,
     useRhcdFilter,
+    uselastSeenFilter,
     useUpdateMethodFilter,
     textFilterState,
     textFilterReducer,
@@ -40,10 +42,13 @@ import {
     rhcdFilterReducer,
     rhcdFilterState,
     updateMethodFilterReducer,
-    updateMethodFilterState
+    updateMethodFilterState,
+    lastSeenFilterReducer,
+    lastSeenFilterState
 } from '../filters';
 import useOperatingSystemFilter from '../filters/useOperatingSystemFilter';
 import useFeatureFlag from '../../Utilities/useFeatureFlag';
+import { DatePicker, Split, SplitItem } from '@patternfly/react-core';
 
 /**
  * Table toolbar used at top of inventory table.
@@ -81,6 +86,7 @@ const EntityTableToolbar = ({
         tagsFilterReducer,
         operatingSystemFilterReducer,
         rhcdFilterReducer,
+        lastSeenFilterReducer,
         updateMethodFilterReducer
     ]), {
         ...textFilterState,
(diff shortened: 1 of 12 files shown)


src/components/filters/useLastSeenFilter.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const lastSeenFilterState = { lastSeenFilter: [] };
+export const LAST_SEEN_FILTER = 'LAST_SEEN_FILTER';
+export const lastSeenFilterReducer = (_state, { type, payload }) => ({
+    ...type === LAST_SEEN_FILTER && {
+        lastSeenFilter: payload
+    }
+});
+
+export const uselastSeenFilter = ([state, dispatch] = [lastSeenFilterState]) => {
```
gkarat, line 12: nit-pick:
```suggestion
export const useLastSeenFilter = ([state, dispatch] = [lastSeenFilterState]) => {
```

src/components/filters/useLastSeenFilter.js
```diff
@@ -0,0 +1,36 @@
...
+    let [lastSeenStateValue, setLastSeenValue] = useState({});
+    const lastSeenValue = dispatch ? state.lastSeenFilter : [lastSeenStateValue];
+    const setValue = dispatch ? (newValue) => dispatch({ type: LAST_SEEN_FILTER, payload: newValue }) : setLastSeenValue;
+
+    const filter = {
+        label: 'Last seen',
+        value: 'last_seen',
+        type: 'radio',
```
gkarat, line 20: I believe this is not really important, but the mocks don't have radio buttons:

image

if it doesn't require big changes, I think we can align it more to the mock's look

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -341,10 +360,25 @@ const EntityTableToolbar = ({
             ...enabledFilters.registeredWith ? [registeredFilter] : [],
             ...enabledFilters.rhcdFilter ? [rhcdFilterConfig] : [],
             ...enabledFilters.updateMethodFilter ? [updateMethodConfig] : [],
+            ...enabledFilters.lastSeenFilter ? [lastSeenFilter] : [],
             ...showTags && enabledFilters.tags ? [tagsFilter] : []
         ] : [],
         ...filterConfig?.items || []
     ];
+    const [startDate, setStartDate] = useState();
+    const [endDate, setEndDate] = useState();
```
gkarat, line 369: nit-pick: could these two state values be replaced with the values from `lastSeenFilterValue`? Reducing the state variables for the toolbar component could help avoid extra updates/re-renders.

gkarat (approved):
great work! the PR is really neat. I could also test it successfully on my machine. there are some minor comments, but overall LGTM.

one idea: weird ancient dates like 199 will make the whole app crash 😁 probably we could limit the range of available "FROM" dates for a user? attaching video

[Screencast from 03-06-2023 01:45:38 PM.webm](https://user-images.githubusercontent.com/31385370/223114743-67492b20-b4d0-416e-9682-c01ea567fea9.webm)

src/components/filters/useLastSeenFilter.js
adonispuente: The ticket hasnt been updated, but there was a conversation with Margot from UX that gives the green light for radio-buttons to align more with the other apps!

src/components/InventoryTable/EntityTableToolbar.js
mkholjuraev: I am wondering if we can move all of these date-related functions to the filter itself. I believe you can access the filter values without saving them in a state.

mkholjuraev: If we keep these states in the toolbar component, we need to re-render the whole toolbar rather than the filter only

adonispuente: > 
Hmm it seems the api breaks when you send a date like that, good catch. Going to ask the backend what it should be limited too. As for the other suggestions, they are all implemented locally and I will push it up all together once I hear back as to include all changes in one commit. Thank you both for taking a look!

gkarat: We still can simply fix the issue on our end by limiting the acceptable time range.. and not passing abnormal values to the back end

adonispuente: When we pass in an abnormal value like the 199 you had, the datePicker converts it to a readable value , like 0199, which is technically correct.
The backend just wasnt aware of this edge case possibility and doesnt know what the minimum value it will accept is. Once the come back to me with it I can update the PR with that validating edge case and/or the backend will update to accept such edge case.

adonispuente:
ci test are failing for the same issue discussed in slack. All test pass locally, will merge today

PR #1782, opened 2023-03-03
test: Bring intercept responses to one format

gkarat:
This rewrites some interceptors to always have the third argument compliant with the interceptors' response object interface (e.g., always include statusCode, body): https://docs.cypress.io/api/commands/intercept#Response-object-properties. This should help avoid "flakiness" for some tests and make the test code utils consistent.

## How to test

The component tests must pass.


Fewwy (approved):
LGTM

PR #1785, opened 2023-03-07
test: Stabilize cypress component tests

gkarat:
This PR 
1. **rewrites some flaky tests** that assert network requests and their search parameters (e.g. filter, sorting, etc.). The problem is that cypress doesn't wait or flush all the requests between `it`s and thus some tests can assert on wrongly intercepted requests. It must be resolved now by `cy.wait`ing for all requests that we set up interceptors for.
   1. Example: the inventory groups table makes the initial request for all data via GET `/groups`. If in the tests we change the pagination and then `cy.wait` for the new paginated request, then it can happen that the first requests hasn't been finished yet and `cy.wait` will catch the very first request instead of the request made after changing pagination (the one we need to check).
2.  **makes cypress run component tests in chrome** browser instead of electron. Before we had a problem that CI would fail with `Could not process 'reset:browser:tabs:for:next:test'. No automation clients connected.` error. https://github.com/cypress-io/cypress/issues/22208 suggests to force running in chrome.

## How to test

CI must pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
     "verify": "npm-run-all build lint test",
     "nightly": "npm run travis:verify",
     "start:federated": "fec static --config config/dev.webpack.config.js",
-    "test:ct": "BABEL_ENV=componentTest cypress run --component",
+    "test:ct": "BABEL_ENV=componentTest cypress run --component --browser chrome",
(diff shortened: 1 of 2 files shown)


Fewwy:
@gkarat Tested it locally, tests are passing. LGTM

PR #1787, opened 2023-03-10
ESSNTL(4056): Adds checks against edge cases for last seen filter

adonispuente:
To test new updates check for the following:

When selecting last seen filter, if you refresh, filter persist

- Make sure you can copy/past the link with the filter present (this is true for all except custom, custom should populate the range picker)
- Make sure Start Date cant be larger than End Date
- Make sure End date cant be smaller than start Date
- End date and start date both cannot go past Todays date
-Should be able to delete both values manually
Filter should not break ever. I believe this PR should catch all edge cases

### DIFF
--- a/src/components/InventoryTable/EntityTableToolbar.js
+++ b/src/components/InventoryTable/EntityTableToolbar.js
@@ -328,7 +328,11 @@ const EntityTableToolbar = ({
         [OS_CHIP]: (deleted) => setOsFilterValue(xor(osFilterValue, deleted.chips.map(({ value }) => value))),
         [RHCD_FILTER_KEY]: (deleted) => setRhcdFilterValue(onDeleteFilter(deleted, rhcdFilterValue)),
         [LAST_SEEN_CHIP]: (deleted) =>
-        {setLastSeenFilterValue(onDeleteFilter(deleted, [lastSeenFilterValue.mark])), setStartDate(), setEndDate();},
+        {
+            setLastSeenFilterValue(onDeleteFilter(deleted, [lastSeenFilterValue.mark])),
+            setStartDate(),
+            setEndDate();
+        },
         [UPDATE_METHOD_KEY]: (deleted) => setUpdateMethodValue(onDeleteFilter(deleted, updateMethodValue)),
         [HOST_GROUP_CHIP]: (deleted) => setHostGroupValue(onDeleteFilter(deleted, hostGroupValue))
     };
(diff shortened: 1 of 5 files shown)


computercamplove:
@adonispuente 

i see still something went wrong because when one of date is invalid you are able to apply another right date and request will be sent because invalid date store old right value:
1. start date apply 2023-03-01 - request sent
2. apply end date 2023-03-09 - request sent
3.  start date apply invalid date "invalid" - no request
4. apply end date 2023-03-10 - request sent
5. end date 2023-02-28 - request sent -> Something went wrong

i think when we have one of the dates is invalid and we apply another correct date - we shouldn't send the request because one of the dates is invalid OR we shouldn't be able to select an end date earlier than the start date of the correct value (when it's invalid)

adonispuente:
I believe I've addressed that edge case with the new commit @computercamplove ! Thanks for taking a look

computercamplove:
1. you fixed case when start date invalid, but not for invalid end date

> (same thing can be reproduced with steps but with end date invalid)
2. when i apply start date, then end date - end date doesn't sent in api request. if i change start date api request sent with previous end date, but not with changed start date
3. if i remove text from input should new api request with updated start and end dated be sent? if i remove text from input it's still stores this information - for example if i apply 2 dates, then remove text input from start date and update end date - api request will store deleted start date and new end date

computercamplove:
@adonispuente verified - bugs are fixed, only left when we remove both text from inputs

computercamplove:
verified manually , seems good

adonispuente:
Rebased branch

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -326,7 +327,8 @@ const EntityTableToolbar = ({
         ),
         [OS_CHIP]: (deleted) => setOsFilterValue(xor(osFilterValue, deleted.chips.map(({ value }) => value))),
         [RHCD_FILTER_KEY]: (deleted) => setRhcdFilterValue(onDeleteFilter(deleted, rhcdFilterValue)),
-        [LAST_SEEN_CHIP]: (deleted) => setLastSeenFilterValue(onDeleteFilter(deleted, [lastSeenFilterValue.mark])),
+        [LAST_SEEN_CHIP]: (deleted) =>
+        {setLastSeenFilterValue(onDeleteFilter(deleted, [lastSeenFilterValue.mark])), setStartDate(), setEndDate();},
```
gkarat, line 331: nit-pick: this somewhat looks weird :-D Can't we rewrite it to the clearer:

```suggestion
        {
          setLastSeenFilterValue(onDeleteFilter(deleted, [lastSeenFilterValue.mark]));
          setStartDate();
          setEndDate();
        },
```

gkarat (approved):
It seems that when the user select the custom date range, "start" and "end" dates are not propagated to URL. Thus, if I want to copy the link and open it, there will be a chip "Custom" but without dates selected previously => all systems are shown.

but! codewise looks good to me, and all the features listed in the description work for me while running locally 👍🏼

adonispuente:
@gkarat Thanks for taking a look. I've made the small nit-pick change. As for the filter dates, that wasnt originally part of the MVP but I do see its use in the future. Theres a few other high prio things to do but afterwards that can be implemented.

PR #1790, opened 2023-03-13
feat(ESSNTL-3727): Display group systems

gkarat:
Implements: https://issues.redhat.com/browse/ESSNTL-3727.
Mocked at: https://www.sketch.com/s/c19f555b-7887-4423-b55d-575c8dd1dfb7/a/52E5bap.

This makes the group details page (/groups/%id) show the systems list if there is at least one host associated with that group. The table component imports `InventoryTable`.

- The table adds extra "Update methods" column. It is non-sortable yet due to the missing back end implementation by this key.
- The export feature is low-prioritized and was asked to be implemented later after main features ready.
- The actions (add/remove systems) will be added in a separate PR.
- Bulk select has only two options (select none and select page). I'll add the "select all" option later since I want to improve `SELECT_ENTITY` action and will track these changes in a separate PR.
- The component tests are almost identical to GroupsTable.cy.js.

## How to test

1. Run `npm run mock-server` and `npm run test:mock` in separate terminals.
2. Navigate to https://stage.foo.redhat.com:1337/insights/inventory/groups/foobar.
3. If the table is not showing and the empty state is rendered, this means mock API server generated `host_ids` as an empty array. Refresh again until the response contain non-empty `host_ids`.
4. Once the table is loaded, all the visible features (sorting, filtering, pagination, bulk selection) should work. Correctness is not guaranteed due to the randomness of the mock server (but nothing should crash).

## Screenshots

image


gkarat:
I will remove the draft label once I have tests implemented (gonna push the last commit soon).

Fewwy (approved):
I tested it locally twice. The only thing that hasn't worked for me is the System Update Method filter. Otherwise LGTM

PR #1793, opened 2023-03-14
chore: Mock GET groups related requests

gkarat:
This serves as a temporary solution for the missing GET endpoints implementation on the API level. This will allow QEs to open the views and find required components selectors.


Fewwy (approved):
LGTM

PR #1797, opened 2023-03-20
chore: Don't use extra span tag to render tab text

gkarat:
This makes the tabs headers on the Group details page not to render extra `span` tag inside `button`. This was caused by using unnecessary `TabTitleText` component.

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -15,6 +15,34 @@ jest.mock('react-router-dom', () => ({
     })
 }));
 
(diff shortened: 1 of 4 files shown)


gkarat:
cc: @computercamplove

computercamplove:
verified on local env, lgtm!

bastilian (approved):
LGTM! Thank you @gkarat!

computercamplove:
@gkarat to merge we need to wait for resolved akamai issue?

PR #1798, opened 2023-03-20
feat(ESSNTL-3728): Enable multiple hosts addition to group

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3728.
Mocked at https://www.sketch.com/s/c19f555b-7887-4423-b55d-575c8dd1dfb7/a/8yK3wl3.

This gives functionality to "Add systems" button in the empty state and also on the top of group systems table (/groups/%id view). The user is given the list of systems and chooses which to add to the current group. If one of the selected systems is already in a group, then the user is additionally prompted for confirmation.

## How to test

Be able to reproduce user flows from the mocks. Check the API requests are called and contain the required data about systems and groups.

## Screenshots

(the columns order has changed since https://github.com/RedHatInsights/insights-inventory-frontend/pull/1798/commits/d147b0641931510a3c52d29c196a9255a1db5ee7)

image

image

image

### DIFF
--- a/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
+++ b/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
@@ -1,0 +1,49 @@
+import { mount } from '@cypress/react';
+import FlagProvider from '@unleash/proxy-client-react';
+import React from 'react';
+import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router-dom';
+import {
+    featureFlagsInterceptors,
+    groupsInterceptors,
+    hostsInterceptors,
+    systemProfileInterceptors
+} from '../../../../cypress/support/interceptors';
+import { unleashDummyConfig } from '../../../../cypress/support/utils';
+import { getStore } from '../../../store';
+import AddSystemsToGroupModal from './AddSystemsToGroupModal';
+
+before(() => {
+    cy.window().then(
+        (window) =>
+            (window.insights = {
+                chrome: {
+                    isProd: false,
+                    auth: {
+                        getUser: () => {
+                            return Promise.resolve({});
+                        }
+                    }
+                }
+            })
+    );
+});
+
+describe('AddSystemsToGroupModal', () => {
+    it('renders', () => {
+        hostsInterceptors.successful();
+        featureFlagsInterceptors.successful();
+        systemProfileInterceptors['operating system, successful empty']();
+        groupsInterceptors['successful with some items']();
+
+        mount(
+            <FlagProvider config={unleashDummyConfig}>
+                <Provider store={getStore()}>
+                    <MemoryRouter>
+                        <AddSystemsToGroupModal isModalOpen={true} />
+                    </MemoryRouter>
+                </Provider>
+            </FlagProvider>
+        );
+    });
+});
--- a/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
+++ b/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
@@ -1,17 +1,30 @@
 import { mount } from '@cypress/react';
+import {
+    checkTableHeaders,
+    MODAL,
+    ouiaId,
+    TABLE
+} from '@redhat-cloud-services/frontend-components-utilities';
 import FlagProvider from '@unleash/proxy-client-react';
+import _ from 'lodash';
 import React from 'react';
 import { Provider } from 'react-redux';
 import { MemoryRouter } from 'react-router-dom';
 import {
-    featureFlagsInterceptors,
-    groupsInterceptors,
-    hostsInterceptors,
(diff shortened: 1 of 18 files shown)


src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -0,0 +1,208 @@
...
+import { addHostsToGroupById } from '../utils/api';
+import apiWithToast from '../utils/apiWithToast';
+import ConfirmSystemsAddModal from './ConfirmSystemsAddModal';
+
+export const prepareColumns = (initialColumns) => {
+    // hides the "groups" column
+    const columns = initialColumns;
+
+    // additionally insert the "update methods" column
+    columns.splice(columns.length - 2 /* must be the 3rd col from the end */, 0, {
```
mkholjuraev, line 25: Isn't it better to filter out the column instead of removing it with the splice method? Because, this can easily get affected if there is another new column added

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -0,0 +1,208 @@
...
+                    variant="large" // required to accomodate the systems table
+                >
+                    <InventoryTable
+                        columns={prepareColumns}
+                        variant={TableVariant.compact} // TODO: this doesn't affect the table variant
+                        tableProps={{
+                            isStickyHeader: false,
+                            canSelectAll: false
+                        }}
+                        bulkSelect={{
```
mkholjuraev, line 172: How about creating a shared bulkSelect config builder? I have seen really similar config passed to multiple components, having a it shared would benefit a lot.

mkholjuraev (commented):
Codewise looks neat. Just minor suggestions.

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
gkarat: @mkholjuraev, hm, I am not removing any columns here from the default set. The only thing which this func is doing is inserting the "Update methods" column as the 3d from the end. This shouldn't be broken if new columns are added (although if the default columns number will be decreased to 3 or less then yeah there will be a problem 😄 but it's the least probable)

gkarat: well, the comment above the line 22 is misleading, it was copy-pasted from another code. I'll remove it

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
gkarat: good idea! done in the last commit

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
```diff
@@ -0,0 +1,152 @@
...
+    hostsInterceptors
+} from '../../../../cypress/support/interceptors';
+import {
+    selectRowN,
+    unleashDummyConfig
+} from '../../../../cypress/support/utils';
+import { getStore } from '../../../store';
+import AddSystemsToGroupModal from './AddSystemsToGroupModal';
+
+const TABLE_HEADERS = ['Name', 'Group', 'Tags', 'Update methods', 'OS', 'Last seen'];
```
Fewwy, line 26: Update method, not methods
image

Fewwy:
Are you sure that the column order is correct?
This is what I see running your pr locally
image

This is sketch
image

Fewwy:
If I select systems on the single group page/group details page and click the button "Add systems" should the selection be saved when the modal window opens?
Currently, it is not saving the selection.
image

image

Fewwy:
Operating system filter looks squashed :) 
image

Fewwy:
I select the full page of items 
image
I open the second page and I already see some items selected
image

When I select the full page again it shows me that I selected 99 items, not 100( the range of incorrectly selected items could be different)
image

Fewwy:
Filter by update method doesn't work (but I think we already encountered this bug before)
image

image

computercamplove:
> Operating system filter looks squashed :)
it might be caused by https://issues.redhat.com/browse/ESSNTL-2639
>  update method doesn't work 
https://issues.redhat.com/browse/RHIF-169
https://issues.redhat.com/browse/ESSNTL-4561

gkarat:
@Fewwy, regarding https://github.com/RedHatInsights/insights-inventory-frontend/pull/1798#issuecomment-1488492910: it's strange but I cannot reproduce it. Tried your scenario, and I see the selections are correct when traversing thru pages/changing perPage parameter.

gkarat:
@Fewwy, regarding https://github.com/RedHatInsights/insights-inventory-frontend/pull/1798#issuecomment-1488485954: no, the first view shows systems that are already in this group.

Fewwy (approved):

PR #1802, opened 2023-03-22
fix: Fix the CI error related to lastSeen filter

gkarat:
This adds a fallback value for lastSeenFilter object which appears to be undefined or null in some CI tests.


adonispuente (approved):
Good catch, LGTM!

PR #1800, opened 2023-03-21
fix(RHCLOUD-24793): Show ROS tab is azure or aws cloud provider

karelhala:
### Description

ROS team brought to our attention that their tab is always hidden on inventory detail. This happend because of recent inventory refactoring and we forgot to move certain part of the code. This PR re-enables it and also removes unused part of redux code which was used in the past to calculate the visibility.

### Regression

https://github.com/RedHatInsights/insights-inventory-frontend/pull/1435

PR #1811, opened 2023-03-28
[master-stable] fix(RHCLOUD-24793): Show ROS tab is azure or aws cloud provider (#1800)

karelhala:
* fix(RHCLOUD-24793): Show ROS tab is azure or aws cloud provider

* Enable ros only for AWS


mkholjuraev:
@karelhala if you do not mind, I am merging this as the fix is needed ASAP.

PR #1820, opened 2023-03-31
fix(ESSNTL-3727): Fix minor issues for groups/%id view

gkarat:
This addresses bugs found in https://issues.redhat.com/browse/ESSNTL-3727.

## How to test:

1. Go to /groups/%any_id view.
2. Check the Tags filter is present and works as expected.
3. Try to select the current page and go to another. The checkbox should not be filled because the second page is not selected.
4. Try to click on any system from the list and check that you are navigated to the host details page.


Fewwy (approved):
LGTM

PR #1823, opened 2023-04-03
fix(ESSNTL-3727): Hide Group filter

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-3727.

Hides the Group filter on the group systems view (/groups/%id).


Fewwy (approved):
LGTM

PR #1824, opened 2023-04-03
fix(ESSNTL-3728): Fix links, show missing tags filter

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-3728.

This goes over minor bugs in the systems table that are present in the hosts addition modal.


Fewwy (approved):
LGTM

PR #1827, opened 2023-04-06
fix(ESSNTL-3730): fixed edge case with create group

Fewwy:

Fixing edge case when the user opened create group modal through add host group modal and then created a group. 
Expected behavior - after the group is created user should be returned to the inventory table
Bugged behavior before - user was returned to the add host group modal


mkholjuraev (commented):
Codewise looks clean, I was returned to the table after a group was created. One more thing, do we want to refresh the table so that the new group is shown?

mkholjuraev (approved):
LGTM! I am considering refreshing the table to be done in the future as described.

PR #1829, opened 2023-04-13
Fixing columns order in modal

Fewwy:
In the add system modal the group column was placed incorrectly 
image


adonispuente (approved):
LGTM!

PR #1837, opened 2023-04-19
fix(LastSeenFilter): Fixes Validators and adds test

adonispuente:


adonispuente:
Failing because container is already in use

PR #1838, opened 2023-04-20
fix(InventroyTable): Add activeFiltersConfig to prop cache to trigger refresh

bastilian:
Compliance adds filters via the `activeFiltersConfig` prop and not the `customFilters` prop and does not use the redux store to keep active filters and does therefore not trigger a refresh properly and only the first filter change in compliance would trigger a refresh, or if a filter is used coming from the inventory, like tags.

How to test:

 * Run inventory locally, possibly also with compliance
 * Open the Compliance -> Systems page
 * Filter by name, or any other "compliance filter"
 * Add to the filter and the table should refresh again

The last step would fail without this change.


mkholjuraev (approved):
LGTM!

PR #1831, opened 2023-04-17
chore(ESSNTL-4076): upgrade react-router to V6, refactor all class c…

mkholjuraev:
…omponents

Resolves: https://issues.redhat.com/browse/ESSNTL-4076

Upgrades react-router to v6 and refactors all class components to function components

PR #1843, opened 2023-04-26
Fed mod router

mkholjuraev:
This PR is intended for only demonstration and discussion. It is meant to be rebased on top of [this proper PR](https://github.com/RedHatInsights/insights-inventory-frontend/pull/1831). 

This is a draft solution for the react-router upgrade chicken and egg problem. With this in place, inventory does not use react-router-dom in federated modules enabling consumer apps to upgrade their package independently.

### DIFF
--- a/src/components/InventoryTable/EntityTable.js
+++ b/src/components/InventoryTable/EntityTable.js
@@ -2,7 +2,7 @@ import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { selectEntity, setSort } from '../../store/actions';
 import { useDispatch, useSelector } from 'react-redux';
-import { useHistory, useLocation } from 'react-router-dom';
+// import { useNavigate, useHistory } from 'react-router-dom';
 import {
     Table as PfTable,
     TableBody,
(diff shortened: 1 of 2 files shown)


mkholjuraev:
@RedHatInsights/team-interact FYI: I have found offical react-router v6-v5 compatible [package](https://github.com/remix-run/react-router/discussions/8753). This looks promising.

mkholjuraev:
@RedHatInsights/team-interact I have looked at if we can make use of MemoryRouter. As I suspected, this seems to be not what we want. Reasons are:

1. Memory router is again of type a router thus replacing [Router](https://github.com/RedHatInsights/insights-inventory-frontend/pull/1843/files#diff-0ea433d919229e00b8645c966063c8c1087f0ec2594d8a28efd49f795af00d3cL26) on AsyncInventory causes inventory fed module to crash with `ndors-node_modules_react-router-dom_dist_index_js.a2ed5d08fe3fc5d4.js:315 Uncaught Error: You cannot render a <Router> inside another <Router>. You should never have more than one in your app.
`
2. Even if theoretically MemoryRouter could have been used, [defaultOnClick](https://github.com/RedHatInsights/insights-inventory-frontend/pull/1843/files#diff-4d0c10d11d96258fd8cb42b1dcd83b5ba34189c7e5a1398459d48a029015a57bL68) would function would not work as MemoryRouter does not have history object hooked up.

mkholjuraev:
Not valid anymore

PR #1845, opened 2023-04-26
Revert Inventory groups back end mocks

gkarat:
This should

1. make Inventory  groups list (/groups) send real requests again to Inventory groups back end (was mocked with static values before),
2. use `per_page` instead of `perPage` in the requests to GET /groups,
3. use `name` instead of `hostname_or_id` filter parameter in the requests to GET /groups.

## How to test

Since responses from back end are empty currently, we can only rely on component cypress tests (= they should pass and you can play with them locally with `npm run test:openct`).


src/components/GroupSystems/GroupSystems.cy.js
```diff
@@ -229,14 +229,14 @@ describe('filtering', () => {
         applyNameFilter();
         cy.wait('@getHosts')
         .its('request.url')
-        .should('include', 'hostname_or_id=lorem');
+        .should('include', 'name=lorem');
```
Fewwy, line 232: In my case, it's failing here. If I change it back to 'hostname_or_id=lorem' it's passing.

src/components/GroupSystems/GroupSystems.cy.js
```diff
@@ -229,14 +229,14 @@ describe('filtering', () => {
...
     });
 
     it('can remove the chip or reset filters', () => {
         applyNameFilter();
         cy.wait('@getHosts')
         .its('request.url')
-        .should('contain', 'hostname_or_id=lorem');
```
Fewwy, line 239: In my case, it's failing here. If I change it back to 'hostname_or_id=lorem' it's passing.

src/components/GroupSystems/GroupSystems.cy.js
gkarat: Yeah, true, the tests are also failing with the same problem. Looking at it...

Fewwy (approved):

PR #1846, opened 2023-04-27
fix(ESSNTL-4714): fix display name unwanted reset on sort

mkholjuraev:
This fixes a random issue on the Inventory table. When you enter a string into the display name filter and remove the filter one char at a time in the input element and then sort by any column, the display name filter gets reset. 

**To test** the issue happens randomly, but this way seems more frequent, so please play with the filter and sorting more:

1. Go to the inventory app and table
2. Search for a string
3. delete the filter one by one in the input
4. sort the table by any column
5.  observe that the filter is still active


adonispuente:
I ran into this issue while testing 
https://user-images.githubusercontent.com/60629070/234872518-7e1e6ab9-0ec6-4b37-97d5-37125060087e.mp4

Fewwy:
Works! LGTM

Fewwy (approved):

PR #1853, opened 2023-05-04
fix(ESSNTL-4791): Hide all /groups requests under feature flag

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4791.

This makes sure Inventory doesn't make any network calls to the groups endpoints if the feature flag is disabled.

## How to test

1. Deploy locally with `ENVIRONMENT=prod npm run start:proxy`
2. (optional) Make sure https://insights.unleash.devshift.net/projects/default/features/hbi.ui.inventory-groups is disabled
3. Navigate to https://prod.foo.redhat.com:1337/insights/inventory, log in
4. Make sure the table is rendered, there is no filter "Group" available, the column "Group" is not rendered, and there are no requests to /api/inventory/v1/groups*-like endpoints in the network tab.

### DIFF
--- a/src/components/filters/useGroupFilter.js
+++ b/src/components/filters/useGroupFilter.js
@@ -4,6 +4,7 @@ import { useEffect, useMemo, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { fetchGroups } from '../../store/inventory-actions';
 import { HOST_GROUP_CHIP } from '../../Utilities/index';
+import useFeatureFlag from '../../Utilities/useFeatureFlag';
 
 //for attaching this filter to the redux
 export const groupFilterState = { groupHostFilter: null };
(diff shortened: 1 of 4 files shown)


src/components/filters/useGroupFilter.js
```diff
@@ -31,8 +32,12 @@ export const buildHostGroupChips = (selectedGroups = []) => {
...
 const useGroupFilter = (apiParams = []) => {
     const dispatch = useDispatch();
+    const groupsEnabled = useFeatureFlag('hbi.ui.inventory-groups');
+
     useEffect(() => {
-        dispatch(fetchGroups(apiParams));
+        if (groupsEnabled === true) {
+            dispatch(fetchGroups(apiParams));
+        }
     }, []);
```
bastilian, line 41: Do you think it would make sense to add the `groupsEnabled` as a dependency here?

bastilian (approved):
LGTM! Thank you @gkarat!

src/components/filters/useGroupFilter.js
gkarat: Good catch, will do

PR #1855, opened 2023-05-09
fix(ESSNTL-3730): Fix Add host to group API call

gkarat:

This makes the "add host to group" modal pass correct host ID and thus send correct API request. Previously, the group's ID in the network request was always equal to `undefined`.

## How to test

1. Run `npm run start:proxy:beta`
2. Authenticate with `insights-qa`, navigate to https://stage.foo.redhat.com:1337/preview/insights/inventory.
3. Select any system and add it to a group.
4. After submit, in the network tab, make sure the POST /groups/... request contains correct ids of both group and added system.

### DIFF
--- a/src/components/InventoryGroups/Modals/AddHostToGroupModal.js
+++ b/src/components/InventoryGroups/Modals/AddHostToGroupModal.js
@@ -37,7 +37,7 @@ const AddHostToGroupModal = ({
 
         apiWithToast(
             dispatch,
-            () => addHostToGroup(group.groupId, modalState.id),
+            () => addHostToGroup(group.id, modalState.id),
             statusMessages
         );
     };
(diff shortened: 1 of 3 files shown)


Fewwy (approved):
LGTM

mkholjuraev:
It seems only host_id is passed to API?

gkarat:
@mkholjuraev, hm, what's the full URL of the request? the URL must contain both the group id and the host id. The format is `POST /v1/groups/%groupId/hosts/%hostId`. But it's strange we also send the request body... I will change it now so that we don't append any extra info...

gkarat:
@mkholjuraev ,yeah, this looks as expected: URL contains both valid group and host IDs. the endpoint is not yet implemented, so the 405 code is also correct for now

PR #1858, opened 2023-05-10
fix(ESSNTL-4486): Update API functions names

gkarat:
After https://github.com/RedHatInsights/insights-inventory-frontend/pull/1847/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R19 upgrade, it turned out some `javascript-clients` functions changed their naming. This updates the places where incorrect function were called.

## How to test

Deploy the PR, navigate to Inventory and try to change display name, ansible name or delete host. The operations must be completed successfully and without errors.

PR #1860, opened 2023-05-10
feat(ESSNTL-1204): Select all systems, batch delete

AsToNlele:


src/components/InventoryTable/utils/api.js
```diff
@@ -0,0 +1,25 @@
...
+    let arr = [];
+    for (let i = 0; i < items.length; i += batchSize) {
+        let chunk;
+        chunk = items.slice(i, i + batchSize);
+        arr.push(chunk);
+    }
+
+    const results = resolve(
+        arr.map(itemArray => () => {
+            return hosts.apiHostDeleteById(itemArray); }
```
gkarat, line 21: We need to have https://github.com/RedHatInsights/insights-inventory-frontend/pull/1858 merged first, the naming for these API functions have changed

src/store/entities.js
```diff
@@ -162,7 +162,18 @@ function loadingRejected(state, { payload }) {
 
 function selectEntity(state, { payload }) {
     const rows = [...state.rows];
-    const toSelect = [].concat(payload);
+    let toSelect = [];
+    if (Array.isArray(payload?.id)) {
```
gkarat, line 166: Here you use optional chaining, whereas in https://github.com/RedHatInsights/insights-inventory-frontend/pull/1860/files#diff-ac2c6e5c21bc579a2232094c424395566464d8ab26301d894dff7a4e1d7a30d6R39 it's strict. I think we better stick to the strict one already: if we never expect `payload` to be undefined, then we will have better means to catch and recognize an error

src/Utilities/hooks/useBulkSelectConfig.js
```diff
@@ -0,0 +1,104 @@
...
+import * as actions from '../../store/actions';
+import React, { useCallback, useRef } from 'react';
+import useFetchBatched from './useFetchBatched';
+import { flatten, map } from 'lodash';
+import { loadSystems } from '../sharedFunctions';
+import { inventoryCache } from '../../components/InventoryTable/InventoryTable';
+import { Spinner } from '@patternfly/react-core';
+
+export const useBulkSelectConfig = (selected, filters, total, rows, loaded) => {
+    const [isBulkLoading, setBulkLoading] = React.useState(false);
```
gkarat, line 12: nit-pick:
```suggestion
    const [isBulkLoading, setBulkLoading] = useState(false);
```

src/Utilities/hooks/useBulkSelectConfig.js
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable camelcase */
+import { useDispatch, useStore } from 'react-redux';
+import * as actions from '../../store/actions';
+import React, { useCallback, useRef } from 'react';
+import useFetchBatched from './useFetchBatched';
+import { flatten, map } from 'lodash';
+import { loadSystems } from '../sharedFunctions';
+import { inventoryCache } from '../../components/InventoryTable/InventoryTable';
```
mkholjuraev, line 8: Can we not use inventoryCache? It can create bugs here as it is prone have an older filter.

src/Utilities/hooks/useBulkSelectConfig.js
```diff
@@ -0,0 +1,104 @@
...
+        const results = flatten(map(data, 'results'));
+        dispatch(actions.selectEntity(results, selected));
+        setBulkLoading(false);
+    };
+
+    return (
+        {
+            id: 'bulk-select-systems',
+            items: [{
+                title: 'Select none (0)',
```
gkarat, line 66: can we disable this item when there are no items selected? you can inspire by https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/GroupSystems/GroupSystems.js#L20

gkarat (changes requested):
It looks like the group filter is broken for some reason. To reproduce: use `insights-qa`, go to Inventory and select at least one group in the group filter. You will get the same list of hosts and the filter is not applied. Compared to the current version on console.stage.redhat.com/preview: the filter works OK.

gkarat:
I have also spotted that `group_name` is no longer added to the API request (GET /hosts) when you apply the group filter.

gkarat (approved):
LGTM, tested locally together with applying different filters, pagination, didn't spot any issues

PR #1862, opened 2023-05-11
fix(ESSNTL-4812): Fix Global filter on refresh

AsToNlele:
Ticket: [ESSNTL-4812](https://issues.redhat.com/browse/ESSNTL-4812)

Global filter now works after refresh

How to test:

1. Open Stage Beta and select a global filter
2. Notice the system count
3. Refresh and notice the system count is different
4. Now run my PR, select the same global filter and refresh
5. Notice the system count stays the same


Fewwy (approved):
LGTM

PR #1863, opened 2023-05-15
fix(ESSNTL-3730): Add delay before fetching groups again

gkarat:

When adding a host to a newly created group, there can be some back-end latency. Thus, new groups (in rare cases) can be not shown in the refreshed list (https://issues.redhat.com/secure/attachment/12968904/2023-05-15%2009-26-02.mp4). This implements a small delay (500 ms) for GET /groups request in the "add host to group" modal.

## How to test

1. Go to https://stage.foo.redhat.com:1337/preview/insights/inventory
2. Select any host and click "Add to group"
3. Click "Create new group", and try to create a new group with any name
4. Once redirected back to the host addition modal, check that the request to GET /groups is not sent immediately but with delay.

### DIFF
--- a/src/components/InventoryGroups/Modals/ModalSchemas/schemes.js
+++ b/src/components/InventoryGroups/Modals/ModalSchemas/schemes.js
@@ -4,6 +4,7 @@ import componentTypes from '@data-driven-forms/react-form-renderer/component-typ
 import { nameValidator } from '../../helpers/validate';
 import { Text } from '@patternfly/react-core';
 import { getGroups } from '../../utils/api';
+import awesomeDebouncePromise from 'awesome-debounce-promise';
 
 export const createGroupSchema = (namePresenceValidator) => ({
     fields: [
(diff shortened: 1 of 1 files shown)


src/components/InventoryGroups/Modals/ModalSchemas/schemes.js
```diff
@@ -4,6 +4,7 @@ import componentTypes from '@data-driven-forms/react-form-renderer/component-typ
 import { nameValidator } from '../../helpers/validate';
 import { Text } from '@patternfly/react-core';
 import { getGroups } from '../../utils/api';
+import awesomeDebouncePromise from 'awesome-debounce-promise';
```
bastilian, line 7: We could use the fec debounce function. Even though it is just a wrapper around the same thing, it would avoid having to add the package to inventory and possibly deal with version inconsistencies. 

https://github.com/RedHatInsights/frontend-components/blob/master/packages/utils/src/debounce/debounce.ts#L10

Fewwy (approved):

src/components/InventoryGroups/Modals/ModalSchemas/schemes.js
gkarat: yeah, I find it useful. I will try to replace it in a separate pr/commit

Fewwy: We already using this package in one other place. I think this is could be done in the future improvements :)

PR #1864, opened 2023-05-15
fix(systemProfileStore): Repositories are enabled by default

croissanne:
The `enabled` field in repository configurations defaults to 1, so consider repositories where this option is missing as enabled.


gkarat:
@croissanne, could you please provide more background to the change? Is there any Jira ticket we could reference?

croissanne:
@gkarat there is no jira ticket, but @jrusz noticed that instances which have  repositories defined which don't have `enabled=1` or `enabled=0` in their config (so just omit it) don't show their repos in the inventory correctly, while it actually defaults to 1. Happy to open a ticket if you want.

jrusz:
I actually have screenshots that show exactly this
image
image

gkarat (approved):
LGTM

PR #1867, opened 2023-05-17
fix: Provide search value when loading groups options

gkarat:
This is a quick patch that provides one extra `name` argument when loading groups options list in AddHostToGroupModal. 

## How to test

Try to find a host in the inventory table that is not in a group. Click "Add to group" and try to type some input in the groups search. The network requests to GET /groups must contain `name` parameter with the input you have typed.


AsToNlele (changes requested):
Hi,
the request has the name parameter correctly, but when I start typing a group name that doesn't exist the loading state never stops. This works on stage beta correctly

Fewwy (approved):
LGTM, will merge after the tests will pass

Fewwy:
I found the issue and had to remove an awesomeDebouncePromise wrapping the entire scheme.

PR #1870, opened 2023-05-19
Add host to group

Fewwy:
This is the same PR as https://github.com/RedHatInsights/insights-inventory-frontend/pull/1867 but with a FIX that prevents infinite loads. 

How to test
Try to find a host in the inventory table that is not in a group. Click "Add to group" and try to type some input in the groups search. The network requests to GET /groups must contain a name parameter with the input you have typed.
Make sure that if you type the group name in the search field - the search won't go "infinite spinner mode" and will either fail to find(if you are typing the name of the not existing group) or find a group you want.


AsToNlele (approved):
LGTM, thank you! :smile:

PR #1873, opened 2023-05-22
fix: Reload table when changing the host group

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3730, https://issues.redhat.com/browse/ESSNTL-3733.

This makes the inventory table reload data when the host is added to or removed from a group via action modals. This also fixed the wrong `ref` property propagation.

## How to test

1. Go to the inventory table
2. Try to add (the network call will fail - this is expected currently) to or remove hosts from groups (the "test" group has some hosts in `insights-qa` profile)
3. Check that after submit the rows are reloaded with the **same** filters.


Fewwy (approved):
LGTM

PR #1876, opened 2023-05-24
fix(Groups): Fix incorrect API keys resolution

gkarat:
(no jira)

This polishes some of the components and ensures tables read right data from responses according to https://console.stage.redhat.com/docs/api/inventory/v1.

## How to test

1. Navigate to /groups with `insights-qa`
2. Make sure all data is rendered correctly: columns have valid data. Verify sorting, filtering, pagination.
3. Go to any group with/without hosts, make sure there are no new visible bugs, data is rendered correctly, if group has hosts, then all these hosts are shown in the table.

### DIFF
--- a/cypress/fixtures/groups.json
+++ b/cypress/fixtures/groups.json
@@ -4,935 +4,352 @@
   "per_page": 50,
   "results": [
     {
-      "host_ids": null,
       "updated_at": "2017-11-13T00:00:00.0Z",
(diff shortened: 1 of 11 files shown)


Fewwy:
If I select groups using checkboxes and remove them the number in the primary toolbar doesn't change and have the same number of groups selected as before deletion.
image

Fewwy (approved):
Other than 2 bugs I found - everything works as expected. LGTM

PR #1877, opened 2023-05-26
fix: Use the same endpoint to add hosts to group

gkarat:
(no jira)

This makes the AddHostToGroupModal use the same endpoints as in the groups systems table. The previous has implementation problems and is still not available to use. The working once receives the list of hosts to add in the body instead of listing them in URL.

## How to test

1. Navigate to /inventory
2. Select any host which is not in a group
3. Add the host to any group and make sure the operation is successful


mkholjuraev (approved):
LGTM! I am able to assign multiple hosts to a group with this PR.

PR #1878, opened 2023-05-29
feat(ESSNTL-3732): Remove hosts from group at /groups/%id

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3732.

## How to test

1. Navigate to any group that has some hosts
2. Try to click on the Remove from group action for one of the hosts and verify that the removal works as expected. There is a delay caused by back end: the removed host can be still in the table after refresh - this issue is to be fixed soon too (in a separate PR).
3. Do the same but with bulk select

## Screenshots

image


gkarat:
@AsToNlele, yes, it looks uneven from my point of view too. thank you

PR #1879, opened 2023-05-29
fix(ESSNTL-4194): Minor improvements

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4194.

This PR follows up on https://github.com/RedHatInsights/insights-inventory-frontend/pull/1767 and fixes the following bugs:

1.  Sorting by Last modified lead to an error because the sorting is not implemented on the back end => the column is now not sortable,
2. Selecting and removing groups from a list was leaving the selected count the same => now the selected number is reset once the selected groups are removed,
3. Success message when removing more groups was containing undefined => fixed this too according to mocks,
4. When user is creating the very first group the empty state is not updated => now when the first group is created the empty message disappears and start to render the table.

## How to test

1. Go to /groups and make sure all of the bugs are resolved.

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -71,6 +71,7 @@ plugins.push(
             './BiosCard': resolve(__dirname, '../src/components/GeneralInfo/BiosCard/BiosCard.js'),
             './DataCollectorsCard': resolve(__dirname, '../src/components/GeneralInfo/DataCollectorsCard/DataCollectorsCard.js'),
             './LoadingCard': resolve(__dirname, '../src/components/GeneralInfo/LoadingCard/LoadingCard.js'),
+            './TextInputModal': resolve(__dirname, '../src/components/GeneralInfo/TextInputModal/TextInputModal.js'),
             // System detail data providers
             './selectors': resolve(__dirname, '../src/components/GeneralInfo/selectors/index.js'),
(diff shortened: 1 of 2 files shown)


src/components/InventoryGroups/InventoryGroups.js
```diff
@@ -28,6 +27,11 @@ const InventoryGroups = () => {
         }
 
         setIsLoading(false);
+    };
+
+    useEffect(async () => {
```
bastilian, line 32: Maybe we can improve this further. Like in the react docs: https://react.dev/learn/synchronizing-with-effects#fetching-data

gkarat: @bastilian, oh, nice, thanks for sharing. Anyway, for which scenario this would be useful here? The fetch function, ones finished, updates only the component state variables and it doesn't do anything global (in terms of redux store handling, for instance).

bastilian: Here for example, it could be used to prevent it calling the setState functions in `checkForGroups`, as this could result in it being called in for some reason this component was unmounted, which would throw an error.

gkarat:
There is now only one test failing which is not related to the PR changes (InventoryGroups). I will research on the cause, but other tests look OK.

fpjrh:
@RedHatInsights/team-interact can we get 1 approving review? Thank you!

Fewwy (approved):
LGTM

PR #1881, opened 2023-05-30
fix(essntl-4735): fix for the tags modal filter (filter case sensitivity)

Fewwy:
I added a filter based on regular expressions that is case insensitive and have improved performance over a .filter method 

Link to the card
https://issues.redhat.com/browse/ESSNTL-4735

Screenshots
image


src/Utilities/regExpInsensitiveFilter.js
```diff
@@ -0,0 +1,47 @@
...
+const toReg = str =>
+    new RegExp(str?.replace(/\//g, '//'), 'gi');
+
+// test a string with a regular expression
+const match = reg => x =>
+    reg?.test(x);
+
+// filter an array by a predicate
+// const filter = f => a => a.filter(a)
+const filter = f => a => {
```
bastilian, line 25: .... is this reimplementing `Array.filter`?

src/Utilities/regExpInsensitiveFilter.js
```diff
@@ -0,0 +1,47 @@
+// run a predicate function over each key of an object
+// const hasValue = f => o =>
+//  Object.keys(o).some(x => f(o[x]))
+const hasValue = f => o => {
+    let key;
+    for (key in o) {
+        if (f(o[key])) {return true;}
+
+        ;
```
bastilian, line 9: ?

bastilian:
@Fewwy This seems to be filtering only on the client, but it's not querying the API, or is it?

Fewwy:
@bastilian Yes, it fetches new list only when you open the Tags Modal

src/Utilities/regExpInsensitiveFilter.js
Fewwy: This assert's that the object has the value we need

bastilian:
@Fewwy But what if there are more tags to fetch that match the searched term?

Fewwy:
@bastilian I don't think this PR should add more fetching to the tags modal. This is a bug fix PR
If we want to add more fetching depending on the filter - that's an update of the existing feature and we have to ask a PM for that.

src/Utilities/regExpInsensitiveFilter.js
bastilian: Why is the `;` flying down here?

src/Utilities/regExpInsensitiveFilter.js
bastilian: Please, do not re-implement a native function. It's doubtful that it is more performant than a native implementation, even if it is more performant it's probably not worth the confusion it contributes.

src/Utilities/regExpInsensitiveFilter.js
Fewwy: Ah, I thought you are asking about the function itself. I removed these floating ;

src/Utilities/regExpInsensitiveFilter.js
Fewwy: @bastilian I rewrote the function

src/Utilities/TagsModal.js
```diff
@@ -36,10 +37,10 @@ const TagsModal = ({
     const activeSystemTag = useSelector(({ entities, entityDetails }) => entities?.activeSystemTag || entityDetails?.entity);
     const tags = useSelector(({ entities, entityDetails }) => {
         const activeTags = entities?.activeSystemTag?.tags || entityDetails?.entity?.tags;
+
         if (activeTags) {
-            return activeTags?.filter(
-                (tag) => Object.values(tag).some((val) => val?.includes(filterBy))
```
bastilian, line 41: Why not just 
```js
val?.toLowerCase().includes(filterBy.toLowerCase())
```

? Does case-insensitive filtering and is far simpler.

src/store/inventory-actions.js
```diff
@@ -170,11 +170,13 @@ export const toggleTagModal = (isOpen) => ({
     payload: { isOpen }
 });
 
-export const fetchAllTags = (search, pagination, getTags = defaultGetAllTags) => ({
-    type: ACTION_TYPES.ALL_TAGS,
-    payload: getTags(search, pagination),
-    meta: { lastDateRequestTags: Date.now() }
-});
+export const fetchAllTags = (search, pagination, getTags = defaultGetAllTags) => {
+    return {
```
bastilian, line 174: We should revert this to the implicit return as it was before to keep the same form as the other actions.

src/Utilities/TagsModal.js
Fewwy: Hm, you are right on that point, I tried to do that in the first place but couldn't make it to work for me

src/store/inventory-actions.js
Fewwy: Yep, that's a clean up I forgot to do, was checking the values of these parameters and forgot to change it back

src/Utilities/TagsModal.js
bastilian: This is pretty much how all (client side) filtering in Compliance works for string searches and it works fine there.

gkarat:
This merge with master ^ should fix the failing cy test not related to this PR.

bastilian (approved):
LGTM! Thank you @Fewwy!

PR #1895, opened 2023-06-08
feat(ESSNTL-3736): Add more scenarios to the group deletion modal

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3736.

It is requested to distinguish between empty and non-empty groups. For the latter, users cannot delete them until they remove all hosts associated with them. According to that, the group deletion modal should render the warning message and provide no options to remove such groups.

## How to test

Navigate to /inventory/groups and try to find both empty and non-empty groups. You should be able to remove the empty groups. You shouldn't be able to remove groups that have some hosts.

## Screenshots

image

image


Fewwy (approved):
LGTM works as expected

PR #1899, opened 2023-06-13
Lift async use effect callback.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-26362

The `useEffect` hook can't use async callback function. Any async side effects have to be lifted out of the hook.


gkarat (approved):
lgtm, thanks a lot!

PR #1900, opened 2023-06-16
fix(RederWrapper): ESSNTL-4954 - Remove check for all permissions

bastilian:
The permissions are too strict, this high up.
Read permissions should be enough to view an inventory component.

**How to test:**

1) Create an user with only the "Compliance viewer" role
2) Visit the Compliance -> Systems page
3) With this the page should be accessible


gkarat (approved):
Played with a user that has only Compliance viewer role, everything works OK, the inventory/compliance systems tables are accessible. LGTM, thank you

PR #1904, opened 2023-06-20
fix(ESSNTL-4746): LastSeen filter backend edge case

adonispuente:
The backend returns a 503 in specific scenarios, this PR adds additional edge cases to lastSeen filter while the backend figures out the issue. This also allows the check on both ends. 

To test, 
Try entering random strings in the custom date inputs, 
Try entering incomplete dates, like 201 or 192,
Enter years well past our time like 1900s

### DIFF
--- a/src/components/filters/helpers.js
+++ b/src/components/filters/helpers.js
@@ -1,6 +1,6 @@
 import moment from 'moment';
 
-export const oldestDate = new Date(1950, 1, 1);
+export const oldestDate = new Date(1970, 1, 1);
 //validators control what date ranges can be selected in the component.
 //both validators need to keep in mind todays date, and the other components inputed date.
 
--- a/src/components/filters/useLastSeenFilter.js
+++ b/src/components/filters/useLastSeenFilter.js
@@ -94,10 +94,10 @@ export const useLastSeenFilter = (
         const todaysDate = moment().endOf('day');
         const selectedFromDate = moment(date).startOf('day');
 
-        if (!containsSpecialChars(date) &&
+        if ((!containsSpecialChars(date) &&
          selectedFromDate < todaysDate &&
           date.length > DEFAULT_DATE_LENGTH &&
-           selectedFromDate > oldestDate) {
+           selectedFromDate > oldestDate) || date.length === 0) {
             if (date > newToDate) {
                 setStartDate();
                 return 'End date must be later than Start date.';
@@ -112,7 +112,7 @@ export const useLastSeenFilter = (
     //This date comes from patternfly component. This manages the 2nd date picker
     const onToChange = (date) => {
 
-        if (!containsSpecialChars(date) && date.length > DEFAULT_DATE_LENGTH) {
+        if ((!containsSpecialChars(date) && date.length > DEFAULT_DATE_LENGTH) || date.length === 0) {
             if (startDate > moment(date)) {
                 return 'Start date must be earlier than End date.';
             } else if (moment(date) > todaysDate) {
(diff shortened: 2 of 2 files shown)


src/components/filters/useLastSeenFilter.js
```diff
@@ -93,7 +93,7 @@ export const useLastSeenFilter = (
         const todaysDate = moment().endOf('day');
         const selectedFromDate = moment(date).startOf('day');
 
-        if (!containsSpecialChars(date) && selectedFromDate < todaysDate) {
+        if (!containsSpecialChars(date) && selectedFromDate < todaysDate && date.length > 9 && selectedFromDate > oldestDate) {
```
bastilian, line 96: Is there something that dictates what the `date.length` has to be, so we don't need to use a hardcoded `9` here?

adonispuente: its more like a 'standard; length because with the format used in the component, MM-DD-YYYY, you have to have at least 10 characters in the field, similar to a phone number

bastilian: I think in this case we could have it as a descriptive constant like `DEFAULT_DATE_LENGTH`.

adonispuente:
@bastilian pinging this back up

gkarat (approved):
LGTM, tried locally the scenarios you provided in the desc. Overall , I have a feeling we could have just wrapped the place in the code where it makes a request with try-catch block. so every time something is invalid, we will still request server and then if it's 503, write that the date is invalid

adonispuente:
@gkarat I like this idea. Right now its fired off in api.js , in a promise. It would cause a rewrite of how that request is sent but it does save us from having to implement more failsafes in the future. I will look into this

bastilian:
@adonispuente Thank you for changing that constant otherwise looks good to me. I'll leave the final approval up to @gkarat. :)

bastilian (approved):

adonispuente:
The minimum date the backend allowed changed so ive updated it. There was an issue when deleting dates with this new const so I added the edge case for that as well

src/components/filters/helpers.js
```diff
@@ -1,6 +1,6 @@
 import moment from 'moment';
 
-export const oldestDate = new Date(1950, 1, 1);
+export const oldestDate = new Date(1970, 1, 1);
```
bastilian, line 3: btw.: Screenshot 2023-06-27 at 00 44 47

bastilian: You can pass Date a value when constructing. From the MDN docs[0]:

> An integer value representing the [timestamp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date) (the number of milliseconds since midnight at the beginning of January 1, 1970, UTC — a.k.a. the [epoch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date)).


So, O milliseconds since January 1, 1970, gives January 1, 1970. :D

[0] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

adonispuente: That doesnt seem like new. Date(0), this is the oldestDate const I updated

adonispuente: Oh thats actually just a huge coincidence haha

bastilian: It's not a coincidence: https://en.wikipedia.org/wiki/Unix_time

But yeah, it does make sense to have it explicit. Maybe. :)

bastilian: That's probably also why the backend has that as the minimum date. :)

gkarat:
Gonna re-run the Travis Test step

gkarat:
LGTM

PR #1906, opened 2023-06-21
fix(RHIENG-747): Fix the group delete modal behavior

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-747.

Users should stay on the same page and the data they see must be not reloaded after closing group deletion modal.

## How to test

1. Find an Inventory group with at least one host and go to its details page.
2. Try to delete the group. You should see the modal saying you are unable to do that.
3. Click on the Close button and make sure the page is not reloaded/navigated to another view.

### DIFF
--- a/cypress/support/interceptors.js
+++ b/cypress/support/interceptors.js
@@ -55,14 +55,14 @@ export const groupsInterceptors = {
 };
 
 export const groupDetailInterceptors = {
-    successful: () =>
+    successful: (fixtures = groupDetailFixtures) =>
         cy
         .intercept(
             'GET',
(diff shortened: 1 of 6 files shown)


AsToNlele (approved):
LGTM thank you!

Unrelated to this PR, I created a new group and added 5 items then selected all(one by one) and clicked "Remove from group". The table refreshed but a few items were still there, after a full refresh the table was empty

gkarat:
@AsToNlele, yeah, there is a latency in the db update and the new GET request seems to be getting an incorrect replica. we are aware of that and gonna fix it soon

PR #1916, opened 2023-06-28
fix(RHINENG-736): Vulnerability and Patch tab in Inventory don't show data for RHSM stystems

jiridostal:
Fixes https://issues.redhat.com/browse/RHINENG-736.

This manipulates a couple of things:

- Makes usage of support function to check staleness more generic
- Introduces checking of RHSM systems to/not display Patch and Vulnerability modules.

As this touches multiple places, increased attention would be appreciated.

### DIFF
--- a/src/Utilities/sharedFunctions.js
+++ b/src/Utilities/sharedFunctions.js
@@ -27,12 +27,14 @@ export const verifyCollectorStaleness = (reporterStaleness) => {
   }
 };
 
-export const verifyCulledInsightsClient = (perReporterStaleness) => {
+export const verifyCulledReporter = (perReporterStaleness, reporter) => {
   //TODO: get rid of !perReporterStaleness condition when dependant apps have per_reporter_staleness info
   if (!perReporterStaleness) {
     return false;
-  } else if (perReporterStaleness.puptoo) {
-    return verifyCollectorStaleness(perReporterStaleness.puptoo) === 'Culled';
+  } else if (perReporterStaleness[reporter]) {
+    return (
+      verifyCollectorStaleness(perReporterStaleness[reporter]) === 'Culled'
+    );
   } else {
     return true;
   }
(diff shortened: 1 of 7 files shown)


src/Utilities/sharedFunctions.js
```diff
@@ -27,12 +27,15 @@ export const verifyCollectorStaleness = (reporterStaleness) => {
...
-export const verifyCulledInsightsClient = (perReporterStaleness) => {
+export const verifyCulledReporter = (perReporterStaleness, reporter) => {
   //TODO: get rid of !perReporterStaleness condition when dependant apps have per_reporter_staleness info
   if (!perReporterStaleness) {
     return false;
-  } else if (perReporterStaleness.puptoo) {
-    return verifyCollectorStaleness(perReporterStaleness.puptoo) === 'Culled';
+  } else if (perReporterStaleness[reporter]) {
+    return (
+      verifyCollectorStaleness(perReporterStaleness[reporter], reporter) ===
```
gkarat, line 36: Looks like the 2nd argument is extra

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -44,10 +51,24 @@ const ApplicationDetails = ({
...
+    const rhsmConduit =
+      !!perReporterStaleness?.[REPORTER_RHSM_CONDUIT] &&
+      !verifyCulledReporter(perReporterStaleness, REPORTER_RHSM_CONDUIT);
+    const rhsmBridge = !!perReporterStaleness?.[REPORTER_RHSM_PROFILE_BRIDGE];
+    return rhsmBridge && rhsmConduit;
+  }, [currentApp]);
+
+  const isEmptyState = (currentApp) =>
+    (currentApp === 'advisor' && isDisconnected) ||
+    (currentApp === 'vulnerabilities' && !isRHSMSystem && isDisconnected) ||
```
gkarat, line 69: I would probably put the apps names to constants so that there is less chance something will get unsynced in terms of naming

PR #1917, opened 2023-06-30
fix(popover): sorting a-z

vkrizan:
Item sorted ascendingly are fixed, sorted from A-Z. This might impact numbers, as those would be also ASC starting from 0.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

ESSNTL-3473
RHINENG-737


vkrizan:
CI Error:
```
navmazing.NavigationDestinationNotFound: Couldn't find the destination [Inventory] with the given class [ApplicationHostInventoryFrontend] the following were available [Dashboard]
```

PR #1918, opened 2023-07-02
feat(ESSNTL-4934): Hide host number if users are not allowed

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4934.

This makes the groups table render 'N/A' if users do not have hosts read permissions.

## How to test

1. Log into account that doesn't have the `inventory:hosts:read` permission
2. Open the groups page (/inventory/groups), and make sure all "Total systems" values are N/A.


mkholjuraev (approved):
LGTM! Testes with my account.

PR #1919, opened 2023-07-02
feat: Integrate /groups/%id with RBAC checks

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4937, https://issues.redhat.com/browse/ESSNTL-4938, https://issues.redhat.com/browse/ESSNTL-4940, https://issues.redhat.com/browse/ESSNTL-4941.

Users need to have at least the read permissions in order to see the group name, its systems and access the group info tab. Also, the write permissions are required in order to delete or rename a group.

## How to test

You can try to use the account that has only inventory:groups:read or inventory:groups:write.


src/components/InventoryGroupDetail/GroupDetailHeader.js
```diff
@@ -21,20 +21,33 @@ import PropTypes from 'prop-types';
...
+  );
+
   const [dropdownOpen, setDropdownOpen] = useState(false);
   const [renameModalOpen, setRenameModalOpen] = useState(false);
   const [deleteModalOpen, setDeleteModalOpen] = useState(false);
 
   const name = data?.results?.[0]?.name;
   const title =
-    uninitialized || loading ? (
+    canRead && (uninitialized || loading) ? (
```
mkholjuraev, line 50: This seems a bit off. If canRead is falsy, I suspect the title will be equal to name or groupId?  I believe, one if-else would make this comparison easier to understand the logic even if it is really basic and not trendy.

src/components/InventoryGroupDetail/InventoryGroupDetail.js
```diff
@@ -21,9 +25,30 @@ const InventoryGroupDetail = ({ groupId }) => {
   const chrome = useChrome();
   const groupName = data?.results?.[0]?.name;
 
+  const { hasAccess: canView } = usePermissionsWithContext(
+    REQUIRED_PERMISSIONS_TO_READ_GROUP(groupId)
+  );
+
+  const EmptyStateNoAccess = () => (
```
mkholjuraev, line 32: Let's refactor this empty state outside of this component. Creating a dedicated folder for such prone-to-be-shared components would be great.

mkholjuraev (commented):
Codewise looks good, a few comments though. Also, as I am not actively following groups UX guidelines, I focused on the code itself.

src/components/InventoryGroupDetail/GroupDetailHeader.js
gkarat: good catch, there was a bug, yes. revisited in the last commit

src/components/InventoryGroupDetail/InventoryGroupDetail.js
gkarat: done!:)

mkholjuraev (approved):
Codewise LGTM!

PR #1920, opened 2023-07-03
feat(ESSNTL-4939): Add/remove hosts from group permissions

gkarat:
Should be merged after https://github.com/RedHatInsights/insights-inventory-frontend/pull/1919.

Implements https://issues.redhat.com/browse/ESSNTL-4939.

This adds an RBAC check on the /groups/%id page: if users don't have sufficient permissions to modify the group, then they cannot add or remove systems, and the buttons are disabled.

## How to test

1. Tweak your account permissions/find another account and remove write `inventory:groups:write` permissions.
2. Check any group and make sure you cannot add new hosts/remove current.
3. A tooltip must be rendered too explaining why the buttons are disabled.


Fewwy (approved):
LGTM

PR #1914, opened 2023-06-27
feat(ESSNTL-4935): Should disable actions at /groups

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4935.

The /groups view must disable actions if users don't have `inventory:groups:write` permission.

image

PR #1922, opened 2023-07-03
feat(ESSNTL-4935): Check permissions in no groups state

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4935.

This is a follow-up on https://github.com/RedHatInsights/insights-inventory-frontend/pull/1914: this PR should disable Create group button when there are no groups and users do not have the write permission.

## How to test

Make sure you don't have any groups and the account is missing inventory:groups:write. Check that the Create group at /groups page is disabled.

## Screenshots

image


Fewwy (approved):
LGTM

PR #1926, opened 2023-07-04
feat(ESSNTL-5052): Disable create group button if not authorized

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-5052.

This disables "Create a new group" in the hosts addition modal if users don't have enough permissions (inventory:groups:write).

## How to test

1. Log in the account that doesn't have inventory:groups:write permission.
2. Navigate to /inventory
3. Find a host that is not in any group yet, and try to add to group
4. In the modal, make sure "Create a new group" button is disabled and the tooltip is shown.

## Screenshots

image


Fewwy (approved):
LGTM

PR #1932, opened 2023-07-12
fix(ESSNTL-4937): Display group name even when can't see hosts

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4937?focusedId=22605892&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22605892.

## How to test

Have an account without `inventory:hosts:read`, but having `inventory:groups:read` - enough to open a group details page. Check that the page title has the group name and not id.

## Screenshots

image


computercamplove:
LGTM!  now i am able to see group name with groups:read permission

PR #1933, opened 2023-07-12
fix(ESSNTL-4938): Hide systems count if no hosts:read

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4937?focusedId=22605892&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22605892.

If users don't have `inventory:hosts:read`, then the groups table (/groups) must not show number of hosts for each group. "N/A" must be rendered instead. This was broken with https://github.com/RedHatInsights/insights-inventory-frontend/pull/1928/files#diff-8a000812b9df4b6f250a6030383a065b82497add0b572416ef6e92fc4f469b95R132.

## How to test

Have an account with groups:read but with not hosts:read permissions. Go to /groups and check that all groups have "N/A" in their total systems number.


computercamplove:
lgtm! without hosts:read not able to see Total systems

PR #1935, opened 2023-07-13
fix(ESSNTL-5052): Opt in resource definition check everywhere

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5052.

We forgot to add `checkResourceDefinitions` flag to RBACProvider, and, thus, all RBAC checks were not taking RD into account. This, for example, impacted the linked Jira bug.

## How to test

Make sure you have `inventory:hosts:*` and `inventory:hosts:read` permissions. Also, you must not have `inventory:hosts:write` permissions. Go to /inventory and find any system which is not in a group. Open the addition to group modal and make sure "Create a new group" is disabled.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -46,7 +46,7 @@ Cypress.Commands.add('mountWithContext', (Component, options = {}, props) => {
     >
       <Provider store={getStore()}>
         <MemoryRouter {...routerProps}>
-          <RBACProvider appName="inventory">
+          <RBACProvider appName="inventory" checkResourceDefinitions>
             {path ? (
               <Switch>
                 <Route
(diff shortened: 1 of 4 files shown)


computercamplove:
now i am not able to see group even if i have group:read with RD, only if i have group:write with RD - i don't have access to Group page at all
@gkarat

gkarat:
Yeah, also found this issue. Must be fixed with https://github.com/RedHatInsights/frontend-components/pull/1781.

computercamplove:
Then this fix is working as expected - "Create button" is disabled on "Add systems to group" modal for user without inventory:hosts:write permissions with resource definition

thank you! @gkarat

PR #1936, opened 2023-07-13
fix: Address some of the RBAC and Inventory groups issues

gkarat:
This fixes:

- Always show the groups table. It is safe since the back end must pre-filter results based on permissions.
- Fixes typo in RD key for groups
- Updates RBAC utils package so that the group details page doesn't crash.


AsToNlele (approved):
This is what you talked about on the standup right? LGTM :smile: 
No idea why the PR build is failing tho

PR #1938, opened 2023-07-17
fix(ESSNTL-4937): Don't check RBAC for Total systems column

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4937.

This corrects the behavior of Total systems columns at /groups: we no longer want to calculate RBAC for each row (each group) and this is why the groups table now just renders the column value which comes from API.

## How to test

With at least inventory:groups:read permission with RD enabled for your account, make sure the Total systems always renders the value which comes from the response from backend ("hosts_count" key).


computercamplove:
works as expected!

PR #1940, opened 2023-07-17
fix(ESSNTL-5043): RBAC and InventoryTable actions

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-5043.

This changes the behavior of InventoryTable actions: now all the actions (delete host, add to group, etc.), both bulk and per-row, are **always** shown to users. However if a user doesn't have enough write permissions (inventory:groups:write or inventory:hosts:write), the actions are disabled with a information tooltip rendered on hover.

- This also disables RBAC check for the /inventory route: asking for ['inventory:hosts:read'] is strict because if users have this permission but limited with a resource definition, then they will still have no access to the page which is not correct.

## How to test

1. Have an account for which you can manipulate permissions,
2. Go to /inventory,
3. Remove your `inventory:hosts:write` permissions and make sure you cannot delete/rename hosts,
4. Remove your `inventory:groups:write` permissions and make sure you cannot add or remove hosts from groups,
5. Check that tooltips are rendered on hover,

## Screenshots 

image

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -40,11 +40,7 @@ export const Routes = () => {
           exact
           path={routes.table}
           render={() => (
-            <RenderWrapper
-              cmp={InventoryTable}
-              isRbacEnabled
-              {...searchParams}
-            />
+            <RenderWrapper cmp={InventoryTable} {...searchParams} />
           )}
           rootClass="inventory"
         />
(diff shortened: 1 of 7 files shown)


mmenestr:
All looks correct, except for bullet two should be make sure you can not delete hosts AND make sure you can not edit hosts (edit = rename, in this case).

gkarat:
Oh, forgot to mention, @mmenestr, this also must be covered with this PR. Will update the description, thanks.

computercamplove:
works as expected for default Inventory roles

PR #1943, opened 2023-07-18
fix(ESSNTL-4937): Disable Add systems in no systems state

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4937.

Disable the Add systems button if there are 0 systems in a group, but users still have not enough write permissions.


Fewwy (approved):
LGTM

PR #1945, opened 2023-07-24
Groups table minor enhancements

gkarat:
No jira. **Has to be merged, not squashed.**

## How to test:

1. Go to /groups
2. Check that when loading, there are checkboxes disabled, and the columns' widths are not changed,
3. Check that there is no "No results" message between the loading state and the state when the rows (groups) are loaded and rendered (you have to have at least one group created),

image


leSamo (approved):
Nice, the skeleton table matches very well, if you want to take this even further you can replace the pagination with skeleton as well to hide the 0 of 0 items in pagination during loading. There is also a slight delay after action and before the skeleton table is shown that does not happen in other apps :thinking: but this is quite minor

gkarat:
Yeah, thanks for the feedback. I think I can take these suggestions and implement in the future (soonest).

PR #1946, opened 2023-07-24
fix(ESSNTL-5064): Fix groups filter by more values

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5064.

This fixes the groups filter in Inventory table: before, we were calling a wrong action, instead, we have to use the one that updates values under `entities`.

## How to test

1. Go to /inventory
2. Try to apply the groups filter by more values, wait for refresh, and try again,
3. The table should show correct data and also update URL parameters accordingly.


Fewwy (approved):
LGTM

PR #1948, opened 2023-07-26
fix(ESSNTL-5110): Improve RBAC checks for delete and edit button on the System details view

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-5110.

This PR:
- Makes sure we always show **Delete**, **Edit display name** and **Edit Ansible hostname** buttons,
- Introduces more advanced check for the group level access: if user is eligible, the button is enabled, otherwise it is disabled with a helping tooltip,
- When user clicks on edit display name or ansible hostname buttons, the modal now has the Submit and Cancel buttons in the right order (more correct according to PF docs), and also the Submit button is disabled if the input value hasn't been changed.

## How to test

For the RBAC change: 

1. make sure you have an account you adjust RBAC permissions/roles
2. verify, that you are able to press Delete and edit buttons while having just `inventory:hosts:write` permission (with the Inventory Hosts Administrator role, for example)
3. then, try to remove this permissions and set `inventory:hosts:write` with a resource definition where the key is `group.id`, the operation `equal` (or `in`), the value is the ID of the group you can add hosts to (for the `in` operation you have to put the ID in an array)
4. once the host is in a chosen group, you have to be able to press the buttons
5. for the hosts outside of the group, you must be **not** able to do that (the buttons are disabled).

## Screenshots

image

image

image

### DIFF
--- a/cypress/support/interceptors.js
+++ b/cypress/support/interceptors.js
@@ -146,10 +146,10 @@ export const deleteGroupsInterceptors = {
 };
 
 export const hostsDetailInterceptors = {
-  successful: () => {
+  successful: (fixtures = hostDetail) => {
     cy.intercept('GET', '/api/inventory/v1/hosts/*', {
       statusCode: 200,
-      body: hostDetail,
+      body: fixtures,
     }).as('getHostDetail');
   },
 };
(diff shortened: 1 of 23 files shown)


Fewwy (approved):

Fewwy:
Reviewed it, everything works as expected

PR #1951, opened 2023-07-28
feat(RHIF-254) Add Zero State to Inventory

adonispuente:
Adds zero state to inventory
Run inventory on an account with 0 systems and run inventory with
ENVIORNMENT=prod npm run start:proxy:beta

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,9 +1,13 @@
 import { Redirect, Route, Switch } from 'react-router-dom';
-import React, { Suspense, lazy, useMemo } from 'react';
+import React, { Suspense, lazy, useEffect, useMemo, useState } from 'react';
 import { getSearchParams } from './constants';
 import RenderWrapper from './Utilities/Wrapper';
 import useFeatureFlag from './Utilities/useFeatureFlag';
 import LostPage from './components/LostPage';
+import axios from 'axios';
+import { Bullseye, Spinner } from '@patternfly/react-core';
+import AsynComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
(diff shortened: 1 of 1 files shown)


adonispuente:
@computercamplove Do you have an idea why the ci test could be failing? It looks like a timeout error maybe while its navigating?

computercamplove:
> @computercamplove Do you have an idea why the ci test could be failing? It looks like a timeout error maybe while its navigating?

@adonispuente yes,
This test just navigation - it navigates to Systems page without creating any host - so it's waited that there will be empty systems table. Since You added new view "Zero state" which doesn't exists in tests - it shows timeout because not able to find system table. I just need to update test or add new one.

i will update test and retest PR

computercamplove:
@adonispuente why Groups page should have Inventory zero state? To create new group you don't need to have systems and Groups page the only place where you can create a group - should it be without Zero state or has own "zero state" but suggesting user to create a group, not registering system?

bastilian (approved):
Works and looks as intended with no hosts! Thank you @adonispuente!

adonispuente:
@computercamplove When zero-state was created that was the originally ask, but now that groups are here, what youre saying makes sense. I'll confirm with UX and get back here

computercamplove:
image
@adonispuente 
when no systems uploaded to account Inventory has this issue
Frontend route: https://env-ephemeral-khsfwv-3b0nk4r5.apps.c-rh-c-eph.8p0c.p1.openshiftapps.com
Frontend login: jdoe | A1HbZXno7viF7x4Y

adonispuente:
This only seems to be a problem in ee. Can you confirm @computercamplove

PR #1952, opened 2023-07-31
fix(essntl-5712): fixed a bug with add system modal for group

Fewwy:
Fixed a bug that was preventing table from showing warning and disabling "add systems" button.
Will create a separate PR for resolving the empty state based on the new permissions

Mocks are in the card https://issues.redhat.com/browse/ESSNTL-5172


AsToNlele (approved):
LGTM :smile:

PR #1959, opened 2023-08-01
fix(ESSNTL-5176): inventory table access with no read permission

Fewwy:
https://issues.redhat.com/browse/ESSNTL-5176
BUG: 
the inventory table were crashing without :hosts:read permission.
It was caused by the fact that if rbacEnabled was false we were passing "true" to the inventory table. This "true" were causing api fetches, but because there were no permission for hosts -> table throws an error and crashes. 
I moved RBAC hasAccess check higher up the component tree to the Wrapper.js to prevent that and removed the condition.
Tested it with 3 different permissions:
No hosts:read permission
image

User have hosts:read permission 
image

### DIFF
--- a/src/Utilities/Wrapper.js
+++ b/src/Utilities/Wrapper.js
@@ -21,7 +21,7 @@ const RenderWrapper = ({
         ref: inventoryRef,
       })}
       isRbacEnabled={isRbacEnabled}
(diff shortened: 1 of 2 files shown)


AsToNlele (approved):
LGTM :smile:

PR #1963, opened 2023-08-04
fix(ESSNTL-5176): hosts read granular permissions for inventory table

Fewwy:
Fixes https://issues.redhat.com/browse/ESSNTL-5176.


src/Utilities/Wrapper.js
```diff
@@ -9,10 +9,44 @@ const RenderWrapper = ({
...
+
+      user[permissionName] = true;
+    }
+
+    return user;
+  }
+
+  //this useEffect trigger once on the page load, get user permissions from the promise
+  //then trigger mapping function and set's user permission data to useState
+  //that way we can get the data from the rbac early and work with the group ID's before we request API or call Hooks
```
gkarat, line 41: I don't see a real value in such extensive comments: basically, the code describes itself pretty well already.

src/Utilities/Wrapper.js
```diff
@@ -9,10 +9,44 @@ const RenderWrapper = ({
...
+  useEffect(() => {
+    async function getUserPermissionData() {
+      const data = await getUserRbacPermissions();
+      const mappedData = mapRbacPermissionsToUser(data);
+      setAllUserPermissions(mappedData);
+    }
+    getUserPermissionData();
+  }, []);
```
gkarat, line 49: While reviewing the PR, I have remembered that we can use `usePermissionsWithContext` and `usePermissions` without RD check. That means, this would be pretty enough if we only rewrite this function call and explicitly tell that we are not interested in validating RD.

```suggestion
  const { hasAccess } = usePermissions(
    'inventory',
    ['inventory:hosts:read'],
    undefined,
    true,
    false
  );
```

Fewwy: Yep, that works!

computercamplove:
Tested in UI, working as expected!

gkarat (approved):
LGTM! Thanks @Fewwy

PR #1964, opened 2023-08-04
fix(RHINENG-767): Fix group table loading

johnsonm325:
Fixes https://issues.redhat.com/browse/RHINENG-767, https://issues.redhat.com/browse/ESSNTL-5029.

Changing between group systems table and add systems to group modal table properly loads. Also, when you delete a system from a group, it properly reloads the table with the deleted system.


johnsonm325:
I'm not sure of how to fix these failing ct tests. these tests fail in master as well.

computercamplove:
fyi @johnsonm325 
Adam Drew is currently working on resolving issue with failed CI `iqe` tests in ephemeral env.:
image

gkarat:
Merge from master should fix the most recent failed Travis build ^

src/components/InventoryGroups/Modals/RemoveHostsFromGroupModal.js
```diff
@@ -55,8 +55,8 @@ const RemoveHostsFromGroupModal = ({
 
     apiWithToast(
       dispatch,
-      () =>
-        removeHostsFromGroup(
+      async () =>
+        await removeHostsFromGroup(
```
gkarat, line 59: I still can see the deleted host after I submit the deletion modal:
1. Select any host,
2. Delete it and wait for the table to be reloaded
3. The deleted host is still present in the table. However, if I do refresh again, it will disappear. It is most probably caused by the backend delay.

gkarat: There are more ways of how to tackle this: we can delay the inventory table reload so that in most of the cases, users will see the accurate data in the table. Or, we can fake the data: don't request hosts again, but just remove the deleted host from the list "manually in UI".

gkarat (commented):
Hey, thanks a lot @johnsonm325! The table works great when adding new hosts: it no longer displays hosts that are not in the group. However, I still see the issue with removing hosts. I described the issue more in review comments

src/components/InventoryGroups/Modals/RemoveHostsFromGroupModal.js
gkarat: For the first option, this works well for me (but yeah, it creates some delay in response to users):

`delay(() => inventory.current.onRefreshData({}, false, true), 1000);` where delay is imported from lodash. But `setTimeout` can be used here too as well.

johnsonm325: I think this is caused by React 18 and their batch api calls. Usually using a useEffect that looks for something that the first api call changes, and then makes the second api call is the way to go. I'm just having trouble reloading the table after the `removeHostsFromGroupModalOpen` is set to false.

johnsonm325: I think this is caused by React 18 and their batch api calls. Usually using a useEffect that looks for something that the first api call changes, and then makes the second api call is the way to go. I'm just having trouble reloading the table after the removeHostsFromGroupModalOpen is set to false.

gkarat: I have pushed a new commit where I am trying to make it possible to change modal reload timeout.

gkarat (approved):
Looks like now both removal and addition reload data properly. Approving it, let's move it to stage. Thank you! @johnsonm325

PR #1966, opened 2023-08-09
fix(ESSNTL-5209): Manage access button and RBAC

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5209.

Enable the button and the link only when user has User Access Administrator role.

## How to test

1. Find/create a group and navigate to its details page,
2. Add User access administrator role to your user and refresh the page. Make sure, at the Group info tab, you can see the Manage access button enabled.,
3. Remove the role and refresh the page. The button must be disabled.

## Screenshots

image


computercamplove:
@gkarat 
working as expected, tested in UI:
- in org admin account
- in non-admin account with granular group access

Fewwy (approved):
LGTM

PR #1967, opened 2023-08-09
fix(ESSNTL-5172): Alter permissions for add systems button

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5172.

Make sure users can open the Add systems modal for any group they have a write permission (and not check for any other permission). Also changes the warning text inside the modal.

## How to test

1. Add `inventory:groups:write` and `inventory:groups:read` permission with group-level access to any of the groups available to your user, and also `inventory:hosts` - same write and read - limited to this group,
2. Navigate to the group you set up permissions for,
3. Make sure you can add new systems (either from the empty state when there are 0 systems in a group, or from the table action otherwise).

## Screenshots

image

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -85,3 +85,15 @@ Cypress.Commands.add(
     );
   }
 );
+
+Cypress.Commands.add(
+  'shouldHaveAriaDisabled',
+  { prevSubject: true },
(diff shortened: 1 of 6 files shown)


AsToNlele (approved):
Works exactly as described, thank you!

computercamplove:
Verified in UI, everything is working as expected

PR #1972, opened 2023-08-11
fix(inventory groups): bug with usepermissions hook

Fewwy:
This reverts the change to granular hosts read permissions.
It caused issues in the service apps

To confirm the issue try running service apps locally using proxy instruction
https://github.com/RedHatInsights/frontend-components/blob/a7976ed5d7d88f1cfd47a78159222da0994be758/packages/config/README.md?plain=1#L214


gkarat:
Let's follow up on this here https://github.com/RedHatInsights/insights-inventory-frontend/pull/1977

PR #1973, opened 2023-08-13
fix(ESSNTL-5239): Improve RBAC check for per-row group actions

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-5239.

Must be merged after https://github.com/RedHatInsights/frontend-components/pull/1885 is released.

This makes sure that both per-row actions (Remove from group and Add to group) on the /inventory page comply with the RBAC requirements: specifically that the table no longer checks on the global level permissions.

- [x] Write cy tests

## How to test

1. Set up `inventory:groups:write` permission only to specific group
2. Navigate to /inventory
3. Find any ungroupped host and make sure you have the "Add to group" (not bulk) enabled and you are able to add the host to the permitted group
4. Refresh the page and make sure that the same host now has "Remove from group" (not bulk) enabled and you are able to remove it from the group.


gkarat:
Ready for review ✔️

PR #1977, opened 2023-08-14
fix: Use the context version of usePermissions

gkarat:
Follows up on https://github.com/RedHatInsights/insights-inventory-frontend/pull/1963.

usePermissions is unstable and cause bugs in apps that import the InventoryTable module. This makes it stick to the RBAC context and guarantees that RD are not checked.

## How to test

Try to run this PR together with any other app - Inventory consumer. Make sure Inventory and the apps are not crashing (that used to be the case when there was `usePermissions` version).


Fewwy (approved):
LGTM

PR #1978, opened 2023-08-14
feat(ESSNTL-5214): Granular RBAC for actions & groups table

Fewwy:
Implements https://issues.redhat.com/browse/ESSNTL-5214.

Granular rbac permissions for each rows actions and the bulk delete action.

how to test:
use account with granular groups write permissions + some groups without granular write permissions(only read permissions). Check the actions for the groups table


gkarat (approved):
Thanks a lot @Fewwy ! I just helped to migrate the remaining actions and also wrote some cy test scenarios.

PR #1981, opened 2023-08-14
fix(RHINENG-993): Use p tag instead small for list items

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-993.

## How to test

On the Inventory details page, all simple lists (like Kernel modules) should have normal black text color.
## Screenshots

image


src/components/GeneralInfo/InfoTable/InfoTable.js
```diff
@@ -153,7 +153,7 @@ class InfoTable extends Component {
         ) : (
           <TextContent>
             {prepareRows(activeRows, pagination).map((row, key) => (
-              <Text component={TextVariants.small} key={key}>
+              <Text component={TextVariants.p} key={key}>
```
vkrizan, line 156: Doesn't this change font size as well?

PR #1982, opened 2023-08-14
fix(ESSNTL-4264): Remove link from inventory groups empty state

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4264.

Remove obsolete links from inventory groups empty states.

## How to test

1. When you don't have any groups created, make sure the empty state doesn't render any secondary links,
2. Check the same for the group without any systems.

## Screenshots

image


adonispuente (approved):
LGTM

PR #1985, opened 2023-08-16
fix(ESSNTL-4667): Disable global filter at groups views

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-4667.

Keep the global filter always disabled at /groups and /groups/%id.

## Screenshots

image


AsToNlele (approved):
LGTM, works great thank you! :smile:

PR #1986, opened 2023-08-16
feat(RHINENG-462): Add CentOS version filtering to inv table

johnsonm325:
This PR adds operating system filtering for CentOS Linux systems. It should act as it always has, but will now filter on more. Definitely try to break it because the data structure being used for filtering has changed significantly and there are many opportunities for this system to break.


johnsonm325:
I need to work on these tests, but this PR can be tested in the meantime.

src/api/api.js
```diff
@@ -117,10 +117,28 @@ export const calculateSystemProfile = ({
...
+        rhelVersions.push(filterValue.value);
+      }
+
+      if (filterValue.osName === 'CentOS Linux') {
+        centosVersions.push(filterValue.value);
+      }
+    });
+
     systemProfile.operating_system = {
+      'CentOS Linux': {
```
bastilian, line 134: If there are no `centosVersions` or `rhelVersions` should their filter even be added to the filter?

src/constants.js
```diff
@@ -141,7 +141,18 @@ export const getSearchParams = () => {
...
-  const operatingSystem = searchParams.getAll('operating_system');
+  const operatingSystem = searchParams
+    .getAll('operating_system')
+    .map((osFilter) => {
+      if (typeof osFilter !== 'object') {
+        const index = osFilter.search(/\d/);
+        const osName = osFilter.substring(0, index);
+        const osVersion = osFilter.substring(index);
+        const [major] = osVersion.split('.');
+        const groupName = `${osName} ${major}`;
```
bastilian, line 152: Could this maybe be accomplished with a RegEx and match the parts needed instead of splicing and dicing the the strings?

vcwild: hope it helps
```suggestion
  const operatingSystem = searchParams
    .getAll('operating_system')
    .map((osFilter) => {
      if (typeof osFilter !== 'object') {
        const match = osFilter.match(/([a-zA-Z\s]+) (\d+\.\d+)/);
        if (match) {
          const [osName, osVersion] = match;
          const [major] = osVersion.split('.');
          const groupName = `${osName} ${major}`;
          return { osName, value: osVersion, groupName };
        }
      }
    });

Comment (johnsonm325) [reply src/constants.js]: @vcwild Oh man!!! I never saw this suggestion. I ended up solving it another way, which also worked. :)

File: src/Utilities/OperatingSystemFilterHelpers.js [hunk 3]
```diff
@@ -4,24 +4,44 @@ import mapValues from 'lodash/mapValues';
...
+      selectedVersion[i].value === osVersion.value &&
+      selectedVersion[i].osName === osVersion.osName
+    ) {
+      return true;
+    } else {
+      continue;
+    }
+  }
+
+  return false;
```
bastilian, line 28: This could be a `.some`[0] like:

```js

const isVersionSelected = (selectedVersions, osVersion) =>
  selectedVersions.some(({value, osName}) =>  value === osVersion.value && osName === osVersion.osName)

```

[0] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/Utilities/OperatingSystemFilterHelpers.test.js
```diff
@@ -118,82 +64,82 @@ describe('groupOSFilterVersions', () => {
...
-    value: '8.5',
+    value: '8.10',
   };
 
   it('returns new selection with group selection enabled', () => {
-    const selection = cloneDeep(testValue);
-    set(selection, [8, 8], false);
+    const selection = cloneDeep(testValue3);
+    selection['RHEL 8'] = {
+      // eslint-disable-next-line prettier/prettier
```
bastilian, line 95: This is not really necessary, or?

src/Utilities/OperatingSystemFilterHelpers.test.js
```diff
@@ -118,82 +64,82 @@ describe('groupOSFilterVersions', () => {
...
     const clickedItem = {
-      value: '8',
+      value: 'RHEL 8',
     };
-    const selection = cloneDeep(testValue);
-    selection[8] = {
-      8: false,
+    const selection = cloneDeep(testValue4);
+    selection['RHEL 8'] = {
+      'RHEL 8': false,
```
bastilian, line 137: Shouldn't all key here get an OS name prefixed?

src/Utilities/OperatingSystemFilterHelpers.test.js
```diff
@@ -118,82 +64,82 @@ describe('groupOSFilterVersions', () => {
...
     const clickedItem = {
-      value: '8',
+      value: 'RHEL 8',
     };
-    const selection = cloneDeep(testValue);
-    selection[8] = {
-      8: true,
+    const selection = cloneDeep(testValue3);
+    selection['RHEL 8'] = {
+      'RHEL 8': true,
```
bastilian, line 122: Same as below.

src/Utilities/OperatingSystemFilterHelpers.test.js
johnsonm325: are you referring to the eslint-disable? It is necessary. It doesn't like `'8.5'`. It wants me to put `8.5`, but when I do that, it messes up the `toEqual`.

src/Utilities/OperatingSystemFilterHelpers.test.js
johnsonm325: Not sure what you mean here. `'RHEL 8': false` is the key value pair needed to check/uncheck the entire group.

src/Utilities/OperatingSystemFilterHelpers.test.js
johnsonm325: ignore my last comment

src/Utilities/OperatingSystemFilterHelpers.test.js
bastilian: I mean for consistency, shouldn't all the keys on the following lines also get a OS name prefix? Otherwise it might be confusing later for someone, why some do and some don't have an OS name attached.

computercamplove:
tested against stage - filter and sorting by Os is working, thank you!

bastilian:
@johnsonm325 Shouldn't the CentOS Linux version be in an separate "CentOS Linux 7" group for example? 

Screenshot 2023-10-05 at 10 18 08

src/Utilities/OperatingSystemFilterHelpers.test.js
bastilian: After understanding it now, it is still odd to me. The separate state for the major version itself seems redundant, since we can assume that if all minor versions are selected the group will be selected, otherwise it'll always be false, or?

computercamplove:
@bastilian did you update PR? i see this 
image

bastilian:
@computercamplove Yeah, i did not. :D

bastilian (approved):
LGTM! Thank you @johnsonm325!

PR #1987, opened 2023-08-17
fix(Systems): THEEDGE-3474 - Fix tab selection

acosferreira:
Adjust tab behavior to URL params respect the filters applied.
Solution based on the current and previous values.
THEEDGE-3474


computercamplove:
@acosferreira tested in local env - works as expected! thank you!

src/routes/InventoryTable.js
```diff
@@ -177,11 +178,23 @@ const Inventory = ({
...
+    setCurrentTab(tabIndex);
+
+    if (currentTab !== tabIndex) {
+      setPrm(searchParams.search);
+      let tabPath = tabsPath[tabIndex];
+      if (tabPath !== undefined) {
+        if (tabPath === '') {
+          tabPath = '/';
+        }
+        history.push(`${tabPath}${prm}`);
```
mkholjuraev, line 194: Please use [InsightsNavigate](https://github.com/RedHatInsights/insights-inventory-frontend/commit/2d9450e7427b8d37b96ddc3c77d693ed935ec4c4#diff-0142c446b1a05f59efb5cca9c6a920822b2dbecac00e598399ba2f2c3015606cR4) for the client-side routing as the router upgrade has been merged to the stage.

mkholjuraev (commented):
It would be great if this search param is also handled on [calculateFilters](https://github.com/RedHatInsights/insights-inventory-frontend/blob/aff393b91d5856938174794ce63a0015e685e636/src/routes/InventoryTable.js#L96) function that is used for other params. If there is not way or it is complex, you can ignore this.

acosferreira:
@mkholjuraev i've added the support to v6 on edge-frontend https://github.com/RedHatInsights/edge-frontend/pull/726

src/routes/InventoryTable.js
```diff
@@ -177,10 +177,20 @@ const Inventory = ({
...
+    setCurrentTab(tabIndex);
+    if (currentTab !== tabIndex) {
+      setPrm(searchParams.search);
+      let tabPath = tabsPath[tabIndex];
+      if (tabPath !== undefined) {
+        if (tabPath === '') {
+          tabPath = '/';
+        }
+        navigate(`${tabPath}${prm}`);
+      }
```
bastilian, line 193: We might be able to simplify this. We don't need to keep the current tab in a state as well, but only use the searchParams. There is an example in the react router migrate to v6 in remediations[0] as an example and an explanation in Slack[1].


[0] https://github.com/RedHatInsights/insights-remediations-frontend/blob/7400ad209414835b67f637855d5d7ffa8b0a770e/src/routes/RemediationDetails.js#L261-L277
[1] https://redhat-internal.slack.com/archives/C02T90PLJ5V/p1692634189934329

acosferreira: @bastilian i've made the change to use the urlSearchParam approach, thanks for the guidance.
@computercamplove the behavior and the ur changed a bit, can you take a look with this new solution?
Thanks

acosferreira: THEEDGE-3474-2

bastilian (approved):
Awesome! Nice work! Thank you @acosferreira!

computercamplove:
@acosferreira 
when i apply filter first in edge tab and then i go to conventional tab - i see in URL filter from edge **has_filters=true**
`https://stage.foo.redhat.com:1337/preview/insights/inventory?page=1&per_page=50&activeTab=conventional&has_filters=true`

and same for conventional tab - when i apply filter in conventional tab **hostname_or_id=test** and go to edge tab - i see in URL filter from conventional tab
`https://stage.foo.redhat.com:1337/preview/insights/inventory?hostname_or_id=test&page=1&per_page=50&activeTab=manage-edge-inventory`

acosferreira:
@bastilian based on the behavior described by @computercamplove I've made the changes to implement the state of tabs and filter and dont duplicate the values whenever you change tabs or click more than once in the same.
Can you please review it?
Thanks

mkholjuraev:
While testing locally, I am not able to properly load the edge tab. The error is 'Scalprum encountered an error! Oe is undefined TypeError: Oe is undefined' which is probably coming from edge fed-module.

PR #1989, opened 2023-08-21
feat(RHIF-104): Add hotjar classes to inventory details page

adonispuente:
This just adds hotjar obfuscating classes to hide host name in inventory details page in the header, and the 'System properties' section in the general tab. There should be not changes monitored in the actual app itself. For testing you can look in the console and search for the data-hj-suppress class and verify that the class is over the system properties section and over the title in the header

Attached are the two things id like to hide with these classes
Screen Shot 2023-08-21 at 10 41 49 AM
Screen Shot 2023-08-21 at 10 44 39 AM

### DIFF
--- a/src/components/GeneralInfo/SystemCard/SystemCard.js
+++ b/src/components/GeneralInfo/SystemCard/SystemCard.js
@@ -90,7 +90,7 @@ class SystemCardCore extends Component {
     } = this.props;
     const { isDisplayNameModalOpen, isAnsibleHostModalOpen } = this.state;
     return (
-      <Fragment>
+      <div className="sentry-mask data-hj-suppress">
         <LoadingCard
           title="System properties"
           isLoading={!detailLoaded}
@@ -242,7 +242,7 @@ class SystemCardCore extends Component {
           )}
           className="sentry-mask data-hj-suppress"
         />
-      </Fragment>
+      </div>
     );
   }
 }
--- a/src/components/GeneralInfo/SystemCard/SystemCard.js
+++ b/src/components/GeneralInfo/SystemCard/SystemCard.js
@@ -90,7 +90,7 @@ class SystemCardCore extends Component {
     } = this.props;
     const { isDisplayNameModalOpen, isAnsibleHostModalOpen } = this.state;
     return (
-      <div className="sentry-mask data-hj-suppress">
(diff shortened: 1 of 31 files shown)


src/components/InventoryDetail/TopBar.js
```diff
@@ -69,7 +69,11 @@ const TopBar = ({
           <Flex>
             <FlexItem>
               <TitleWrapper>
-                <Title headingLevel="h1" size="2xl">
+                <Title
+                  headingLevel="h1"
+                  size="2xl"
+                  className="sentry-mask data-hj-suppress"
```
LiorKGOW, line 75: I can see those classNames in the DOM in the console 👍🏼

LiorKGOW (commented):
Hey @adonispuente 
I can see those classNames in the DOM in the console now
Did have a question regarding your changes though

adonispuente:
Id squad the commits once this is merged @LiorKGOW. Some on the commits are just rebasing against master

src/components/GeneralInfo/SystemCard/SystemCard.js
```diff
@@ -152,6 +154,7 @@ class SystemCardCore extends Component {
                       </Fragment>
                     ),
                     size: 'md',
+                    customClass: 'sentry-mask data-hj-suppress',
```
LiorKGOW, line 157: Just because you are using it multiple times, I would consider adding those className combination to the constants.js file 
The name of the constant could be "HIDE_CLASSES" or something similar or better suited for the reason those values are being hidden for
Just an idea of course if you want 😇  I don't think it is a must, but it could improve readability of the code in my opinion 😀

LiorKGOW (approved):
I can see those classNames on the elements in the DOM
Nice work Adonis :) 

I Did have a small recommendation regarding the classNames string, though it is not a must
I leave the choice to you 

Just a quick reminder to squash the commits when this is merged 😇

src/components/GeneralInfo/SystemCard/SystemCard.js
adonispuente: I dont mind this idea, but I think it would be better implemented if we had that in something like the FEC repo and then imported the class since we use this in a good amount of repos.

PR #1993, opened 2023-08-22
fix(ESSNTL-5252): Hide Create a new group button if not permitted

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5252.

This PR:
- makes the "Add selected hosts to group" modal hide the create group button when the user doesn't have general groups write permission.
- corrects the style for the modal so that the list of available groups is visible and can be easily scrolled.

## How to test

Remove general `inventory:groups:write` from your account. Navigate to /inventory and try to add any host to a group. In the modal, you shouldn't see the "Create a new group" button.

## Screenshots

image

image


AsToNlele (approved):
LGTM, great job! :smile:

PR #1996, opened 2023-08-22
fix(RHINENG-1420): Fix options duplicates in group search

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-1420.

The add hosts to group modal was behaving weirdly when searching for groups: creating duplicates of the same results with each input change from user. This bug is caused by improper handling of objects as values for data-driven-form options: https://github.com/data-driven-forms/react-forms/issues/1401. This creates a workaround to fix this issue.

## How to test

Go to /inventory and try to add any host to a new group. In the modal, make sure that the search for group works correctly and provides correct results.


AsToNlele (approved):
LGTM, good job :smile:

PR #1997, opened 2023-08-23
add notification when edit/delete device

mgold1234:
Fixed # THEEDGE-3465

Description
this commit fix the missing notification in case of editing name of inventory and delete device,
when working with federation at the inventory side.
Screenshot 2023-08-23 at 17 38 35


mgold1234:
I opened new one with relevant changes

PR #1999, opened 2023-08-25
fix(ESSNTL-5252): In Add system modal, load writable groups

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5252.

While adding systems to existing group, provide to user only the groups that they have inventory:hosts:write permission. This help avoid errors when users accidentally add a host to the group they don't have access.

Also changing debounce timeout from 500 to 250 ms to make the UX a bit faster.

## How to test

Limit your groups write permission to only specific groups. Navigate to /inventory and make sure, while adding a host to group, that you see only the groups you have write access to. Try to apply general write permissions and check that you see all of the groups available for your account.

### DIFF
--- a/config/overrideChrome.js
+++ b/config/overrideChrome.js
@@ -6,4 +6,5 @@ export default () => ({
   on: () => {},
   getApp: () => 'inventory',
   getBundle: () => 'rhel',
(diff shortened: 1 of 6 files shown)


Fewwy (approved):
LGTM

PR #2001, opened 2023-08-25
fix: Do not add history entry on host details page

gkarat:
(no jira)

This improves how the system details page change the browser history on navigating between tabs. Users used to stay on the same host details page if they clicked on different tabs like Advisor or Patch and then tried to click on browser's Back/Forward button. This is not really comfortable, and pressing Back should navigate to the previous page instead. To do this, the current browser's history entry should be replaced instead of adding a new one.

## How to test

Go to /inventory and navigate to any host details page. Click on tabs, then click Back in your browser and make sure you are navigated back to /inventory table (users used to stay on host details page which is not expected behaviour).

### DIFF
--- a/src/routes/InventoryDetail.js
+++ b/src/routes/InventoryDetail.js
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector, useStore } from 'react-redux';
-import { Link, useLocation, useParams } from 'react-router-dom';
+import { Link, useLocation, useNavigate, useParams } from 'react-router-dom';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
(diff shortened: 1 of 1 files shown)


Fewwy (approved):

src/routes/InventoryDetail.js
```diff
@@ -113,9 +112,12 @@ const Inventory = () => {
   useEffect(() => {
     if (searchParams.get('appName') === null) {
       searchParams.set('appName', appList[0].name);
-      navigate({
-        search: searchParams.toString(),
-      });
+      navigate(
```
bastilian, line 115: You can use `useSearchParams` for this.

gkarat: Good suggestion:) added to the PR

gkarat:
Again the __coverage__ false negative in cy tests... The rest is fine so assuming tests pass

PR #2003, opened 2023-08-28
fix(ESSNTL-5253): Change required permissions for tabs

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5253. Should be merged after https://github.com/RedHatInsights/frontend-components/pull/1900 is released.

With this PR, Inventory fetches permissions for all applications, not only from the inventory scope. This way, we can read all the required permissions before rendering contents of each tab on the System details page (Advisor, Vulnerability, etc.). This also updates the list of permissions according to the stage, not prod, RBAC config, and makes exception for org. admins to always show the tab contents.

## How to test

Have an account for which you can alter roles/permissions. Go to /inventory and any system's details page. Leave only the Inventory Read role, and verify that all tabs except General and Advisor are unavailable (should write no access). Try to add some viewer roles for each app and check that the tabs are now viewable.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,7 +9,7 @@ const App = () => {
   return (
     <div className="inventory">
       <NotificationsPortal />
-      <RBACProvider appName="inventory" checkResourceDefinitions>
+      <RBACProvider checkResourceDefinitions>
         <Routes />
       </RBACProvider>
     </div>
--- a/src/App.js
+++ b/src/App.js
@@ -9,7 +9,10 @@ const App = () => {
   return (
     <div className="inventory">
       <NotificationsPortal />
-      <RBACProvider checkResourceDefinitions>
+      <RBACProvider
+        appName={null /* fetch permissions from all scopes */}
+        checkResourceDefinitions
+      >
         <Routes />
       </RBACProvider>
     </div>
--- a/src/constants.js
+++ b/src/constants.js
@@ -249,24 +249,22 @@ export const USER_ACCESS_ADMIN_PERMISSIONS = ['rbac:*:*'];
 export const TAB_REQUIRED_PERMISSIONS = {
   /**
    * Should be up to date with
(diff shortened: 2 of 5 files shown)


Fewwy (approved):
LGTM

src/constants.js
```diff
@@ -249,24 +249,22 @@ export const USER_ACCESS_ADMIN_PERMISSIONS = ['rbac:*:*'];
 export const TAB_REQUIRED_PERMISSIONS = {
   /**
    * Should be up to date with
-   * https://github.com/RedHatInsights/rbac-config/tree/88ab3a3adb9526d3dcdb0e1e26c30cc98f51f76e/configs/prod/roles
+   * https://github.com/RedHatInsights/rbac-config/tree/master/configs/stage/roles
    * viewer roles.
    */
-  advisor: ['advisor:*:*', 'inventory:*:read'],
+  advisor: [],
```
adonispuente, line 255: We're removing all permissions for advisor? Shouldnt they have read perms at least?

gkarat: according to the viewer roles, there are no permissions needed to see advisor. so it was mapped to empty array here

src/App.js
```diff
@@ -9,7 +9,10 @@ const App = () => {
   return (
     <div className="inventory">
       <NotificationsPortal />
-      <RBACProvider appName="inventory" checkResourceDefinitions>
+      <RBACProvider
+        appName={null /* fetch permissions from all scopes */}
```
bastilian, line 13: What if we extend the `RBACProvider` to support something like `scopes`, with `*` being all?

gkarat: I was an author of the recent RBACProvider changes actually, and yeah, this is the good suggestion. There were some suggestions from Karel to keep it simple for now though. So there are three values accepted ATM: null for fetching all scopes, undefined for none, and string that represents the name of app and also scope.

PR #2007, opened 2023-08-29
feat(ESSNTL-4485): Add Select All in Groups detail, Add systems modal

AsToNlele:
[ESSNTL-4485](https://issues.redhat.com/browse/ESSNTL-4485)

This implements Select All option in Groups detail page and "Add Systems" modal inside the Groups detail

Merged the implementation of Group bulk select and Inventory bulk select, so Inventory bulk select has to be verified as well.

How to test:
1. Make sure correct amount gets added when you Add a system / deleted when you remove systems.
2. Make sure selection still behaves correctly in Inventory systems with and without applied filters.

### DIFF
--- a/src/components/GroupSystems/GroupSystems.js
+++ b/src/components/GroupSystems/GroupSystems.js
@@ -1,6 +1,4 @@
 import { TableVariant, fitContent } from '@patternfly/react-table';
-import difference from 'lodash/difference';
-import map from 'lodash/map';
 import PropTypes from 'prop-types';
 import React, { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
@@ -19,35 +17,8 @@ import {
   ActionDropdownItem,
 } from '../InventoryTable/ActionWithRBAC';
 import { clearEntitiesAction } from '../../store/actions';
-
-export const bulkSelectConfig = (
-  dispatch,
-  selectedNumber,
-  noneSelected,
-  pageSelected,
-  rowsNumber
-) => ({
-  count: selectedNumber,
-  id: 'bulk-select-systems',
-  items: [
-    {
-      title: 'Select none (0)',
-      onClick: () => dispatch(selectEntity(-1, false)),
-      props: { isDisabled: noneSelected },
-    },
-    {
-      title: `${
-        pageSelected ? 'Deselect' : 'Select'
-      } page (${rowsNumber} items)`,
-      onClick: () => dispatch(selectEntity(0, !pageSelected)),
-    },
-    // TODO: Implement "select all"
-  ],
-  onSelect: (value) => {
-    dispatch(selectEntity(0, value));
-  },
-  checked: selectedNumber > 0 && pageSelected, // TODO: support partial selection (dash sign) in FEC BulkSelect
-});
+import { useBulkSelectConfig } from '../../Utilities/hooks/useBulkSelectConfig';
+import { difference, map } from 'lodash';
 
 export const prepareColumns = (initialColumns, hideGroupColumn) => {
   // hides the "groups" column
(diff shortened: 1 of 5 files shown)


src/routes/InventoryTable.js
```diff
@@ -51,6 +51,7 @@ import {
 } from '../components/InventoryTable/ActionWithRBAC';
 import uniq from 'lodash/uniq';
 import useInsightsNavigate from '@redhat-cloud-services/frontend-components-utilities/useInsightsNavigate/useInsightsNavigate';
+import { difference, map } from 'lodash';
```
gkarat, line 54: IIRC, with this import we will have the issue of loading the whole lodash library. And since it's not yet solved, we better rewrite it to individual imports per each function required.

```suggestion
import difference from 'lodash/difference';
```

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -71,6 +81,16 @@ const AddSystemsToGroupModal = ({
...
   );
 
+  const calculateSelected = () => (selected ? selected.size : 0);
+
+  const handleModalClose = () => {
+    setIsModalOpen(false);
+    if (calculateSelected() > 0) {
+      dispatch(selectEntity(-1, false));
+    }
+    dispatch(setFilter([]));
```
gkarat, line 91: nit-pick: I think you can use CLEAR_FILTERS action or something similar, it will erase the applied filters and also selections for InventoryTable.

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -1,19 +1,20 @@
...
+  fetchGroupDetail,
+  selectEntity,
+} from '../../../store/inventory-actions';
+import { prepareColumns } from '../../GroupSystems/GroupSystems';
 import InventoryTable from '../../InventoryTable/InventoryTable';
 import { addHostsToGroupById } from '../utils/api';
 import apiWithToast from '../utils/apiWithToast';
 import ConfirmSystemsAddModal from './ConfirmSystemsAddModal';
+import { useBulkSelectConfig } from '../../../Utilities/hooks/useBulkSelectConfig';
+import { difference, map } from 'lodash';
```
gkarat, line 16: The same applies here.

src/components/GroupSystems/GroupSystems.js
```diff
@@ -19,35 +17,8 @@ import {
...
-    },
-    // TODO: Implement "select all"
-  ],
-  onSelect: (value) => {
-    dispatch(selectEntity(0, value));
-  },
-  checked: selectedNumber > 0 && pageSelected, // TODO: support partial selection (dash sign) in FEC BulkSelect
-});
+import { useBulkSelectConfig } from '../../Utilities/hooks/useBulkSelectConfig';
+import { difference, map } from 'lodash';
```
gkarat, line 21: 👀

gkarat (approved):
LGTM, thanks a lot. Let's just fix the lodash stuff, other than that we are ready to merge

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
AsToNlele: I've found `CLEAR_ENTITIES` but it doesn't set selected to empty, but replaced the `setFilter([])` with `clearFilter()`

src/routes/InventoryTable.js
AsToNlele: Good point :smile:

AsToNlele:
I've thought about putting your lodash magic inside the `useBulkSelectConfig` so we don't repeat it twice

gkarat (approved):

PR #2010, opened 2023-08-30
feat(ESSNTL-4494): Implement searchable group filter

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-4494.

This enhances the group filter for InventoryTable: the group options are now can be searched.

This changes the filter type to 'custom', but I tried to keep all of the input/return values of `useGroupFilter` untouched so that it's easier to integrate.

## How to test

Try to search for your groups using this filter and apply any of them and check whether the table is filtered correctly. You can also use keyboard (arrows, tab, enter) to navigate within the filter. The placeholder text also should show correct values in different scenarios when there are items found or not. The filter chips must be synchronised with the filter selected values too.

The filter is updated also at the group details view, add to group modal (everywhere where InventoryTable is used with default filters).

## Screenshots

Screenshot 2023-08-31 at 10 44 33

Screenshot 2023-08-31 at 10 44 39

### DIFF
--- a/src/components/InventoryTable/InventoryTable.cy.js
+++ b/src/components/InventoryTable/InventoryTable.cy.js
@@ -213,8 +213,8 @@ describe('with default parameters', () => {
       it('options are populated correctly', () => {
         cy.get('button[data-ouia-component-id="ConditionalFilter"]').click();
         cy.get(DROPDOWN_ITEM).contains('Group').click();
-        cy.ouiaId('Filter by group').click();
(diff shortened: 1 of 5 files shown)


gkarat:
Travis again wrongly failed with the coverage-like issue in tests. But all passed.

Fewwy (approved):
Amazing work, I remember how much time I spent trying to make this work. :) I am very impressed

PR #2011, opened 2023-08-31
feat(RHIF-282): Render get help expandable

gkarat:
Implements https://issues.redhat.com/browse/RHIF-282.

This adds some helping notes on the top of the groups table. The second link must be hidden if user is not an org admin nor user access adminstrator.

## Screenshots

Screenshot 2023-09-01 at 12 45 07


Fewwy (approved):
LGTM

gkarat:
Will wait for UX ack before merging

PR #2012, opened 2023-08-31
feat(ESSNTL-4485): Fix BulkSelect text

AsToNlele:
[ESSNTL-4485](https://issues.redhat.com/browse/ESSNTL-4485)

Changed bulk select text in parentheses. Now it should be eg. (5 items) and (1 item) if there is only one.
It's described in the doc inside the ticket.

How to test:
1. Make sure the text is correct
2. Verify for a single item it should say "Select page/all (1 item)", easiest is to use a inventory group with a single system inside

| Before               | After               |
| ---------------------- | ---------------------- |
| image | image|


gkarat (approved):
works great, thank you! @AsToNlele 

will wait for tests before merging

PR #2013, opened 2023-08-31
Refactor the /routes/InventoryTable into hybridInventory tabs

mkholjuraev:
This PR is intended to consolidate the current state of /routes/InventoryTable component by extracting the tabs into the hybridInventoryTabs module and preparing the insights-inventory to expose this new module by federation.

How to test:
1. Run the PR locally
2. No change should be present in functionality at this phase


src/components/InventoryTabs/HybridInventoryTabs.js
```diff
@@ -0,0 +1,67 @@
...
+import useFeatureFlag from '../../Utilities/useFeatureFlag';
+import { resolveRelPath } from '../../Utilities/path';
+import useInsightsNavigate from '@redhat-cloud-services/frontend-components-utilities/useInsightsNavigate/useInsightsNavigate';
+import { manageEdgeInventoryUrlName } from '../../Utilities/edge';
+
+const HybridInventoryTabs = ({
+  ConventionalSystemsTab,
+  ImmutableDevicesTab,
+}) => {
+  const navigate = useInsightsNavigate();
```
gkarat, line 14: I think we better leave some TODO notes on removing the useInsightsNavigate usage: once we complete all of the router work, we have to migrate to useNavigate from react-router-dom.

src/components/InventoryTabs/HybridInventoryTabs.js
```diff
@@ -0,0 +1,67 @@
...
+    if (tabPath !== undefined) {
+      navigate(`${tabPath}`);
+    }
+    setActiveTabKey(tabIndex);
+  };
+
+  const EdgeParityEnabled = useFeatureFlag('edgeParity.inventory-list');
+
+  return EdgeParityEnabled ? (
+    <Tabs
```
gkarat, line 38: nit-pick: adding `aria-label` would be a little but useful a11y improvement.

src/components/InventoryTabs/HybridInventoryTabs.js
```diff
@@ -0,0 +1,67 @@
...
+  const initialActiveTabKey =
+    tabsPath.indexOf(pathname) >= 0 ? tabsPath.indexOf(pathname) : 0;
+  const [activeTabKey, setActiveTabKey] = useState(initialActiveTabKey);
+  useEffect(() => {
+    setActiveTabKey(initialActiveTabKey);
+  }, [pathname]);
+  const handleTabClick = (_event, tabIndex) => {
+    const tabPath = tabsPath[tabIndex];
+    if (tabPath !== undefined) {
+      navigate(`${tabPath}`);
```
gkarat, line 30:
```suggestion
      navigate(tabPath);
```

src/components/InventoryTabs/ImmutableDevices/EdgeDevicesTab.js
```diff
@@ -0,0 +1,33 @@
...
+import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
+import { resolveRelPath } from '../../../Utilities/path';
+import {
+  getNotificationProp,
+  manageEdgeInventoryUrlName,
+} from '../../../Utilities/edge';
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useDispatch } from 'react-redux';
+
+const ImmutableDevicesTab = () => {
```
gkarat, line 12: nit-pick: wouldn't it be bad to align the module and filename naming? So that we use `EdgeDevicesTab` or `ImmutableDevicesTab` everywhere?

src/components/InventoryTabs/ImmutableDevices/EdgeDevicesTab.js
```diff
@@ -0,0 +1,33 @@
...
+      urlName={manageEdgeInventoryUrlName}
+      notificationProp={notificationProp}
+    />
+  );
+};
+
+ImmutableDevicesTab.defaultProps = {
+  initialLoading: true,
+  notificationProp: PropTypes.object,
+};
```
gkarat, line 32: Seems to be obsolete given that the component doesn't disclose any props above.

src/components/InventoryTabs/ConventionalSystems/useTableActions.js
```diff
@@ -0,0 +1,120 @@
...
+  NO_MODIFY_GROUPS_TOOLTIP_MESSAGE,
+  NO_MODIFY_GROUP_TOOLTIP_MESSAGE,
+  NO_MODIFY_HOST_TOOLTIP_MESSAGE,
+  REQUIRED_PERMISSIONS_TO_MODIFY_GROUP,
+  REQUIRED_PERMISSION_TO_MODIFY_HOST_IN_GROUP,
+} from '../../../constants';
+import { ActionDropdownItem } from '../../InventoryTable/ActionWithRBAC';
+
+// some actions are hidden under feature flag
+export default (
```
gkarat, line 13: We should assign a display name for this component so that it helps with debugging in the future.

```
const useTableActions = ...

export default useTableActions;
```

src/components/InventoryTabs/ConventionalSystems/useTableActions.js
```diff
@@ -0,0 +1,120 @@
...
+  groupsEnabled,
```
gkarat, line 14: We can simplify the hook interface by removing this argument and calling the feature hook inside the useTableActions instead.

src/components/InventoryTabs/ImmutableDevices/EdgeDevicesTab.js
mkholjuraev: Actually, most of the apps are going to have consolidated tabs in the future. Thus, this is better approach

src/components/InventoryTabs/HybridInventoryTabs.js
mkholjuraev: Hmm, should we keep useInsightsNavigate hook afterward also for a unified routing experience? I kinda like it now :)

bastilian: I would vote for keeping the `useInsightsNavigate` hook. We should improve it though.

There might be a way to accomplish the same with the raw react router hook or `Link` component, and if we do find out how to do it that way, we can deprecate our custom `useInsightsNavigate` and `InsightsLink` components.

gkarat: Yeah, this makes sense

src/components/InventoryTabs/ImmutableDevices/EdgeDevicesTab.js
```diff
@@ -0,0 +1,31 @@
...
+      pathPrefix={resolveRelPath('')}
+      urlName={manageEdgeInventoryUrlName}
+      notificationProp={notificationProp}
+    />
+  );
+};
+
+ImmutableDevicesTab.defaultProps = {
+  initialLoading: true,
+};
```
gkarat, line 30: This is obsolete

gkarat (commented):
Hosts actions seem to be broken (also shown in failed cypress tests)

https://github.com/RedHatInsights/insights-inventory-frontend/assets/31385370/efdb9101-bfea-4bdc-9975-8afdea1edd57

gkarat (approved):
The table actions issue is fixed, and the table works fine. I couldn't spot any issues. LGTM! Thanks @mkholjuraev 

One thing noticed: when you apply some filters in the Conventional tab, navigate to the Immutable and getting back to the Conventional, then URL filter params are not preserved. We should support the tabs change and filter URL params in the future I believe.

mkholjuraev:
@gkarat thank you for the review! Aha, I am aware of it, using URL query parameters and URL hash for the tab change was not possible because of Inventory and edge fed-mod manipulating the same URL. I will address this when I am working with the edge fed-module

PR #2014, opened 2023-08-31
fix(ESSNTL-5315): Fix last seen filter

adonispuente:
A patternfly updated affected the params that the datepicker uses, causing the component not to work anymore with just the 'date' param. This PR fixes that.

To test simply head to /systems page, enable the custom filter, and make sure the filter works as expected

### DIFF
--- a/src/components/filters/useLastSeenFilter.js
+++ b/src/components/filters/useLastSeenFilter.js
@@ -92,7 +92,7 @@ export const useLastSeenFilter = (
   };
 
   //This date comes from patternfly component. This manages the 1st date picker
-  const onFromChange = (date) => {
+  const onFromChange = (_event, date) => {
     const newToDate = moment(endDate).endOf('day');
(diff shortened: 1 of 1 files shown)


src/components/filters/useLastSeenFilter.js
```diff
@@ -92,41 +92,41 @@ export const useLastSeenFilter = (
   };
 
   //This date comes from patternfly component. This manages the 1st date picker
-  const onFromChange = (date) => {
+  const onFromChange = (_event, str, date) => {
```
bastilian, line 95: The renaming it to `str` isn't really needed, or?

adonispuente: I was mostly just following the convention in the doc, I could change it

gkarat (approved):
LGTM, the custom filter works fine 👍🏼 thanks @adonispuente

PR #2016, opened 2023-09-07
feat(THEEDGE-3524) Include list of vulnerabilities and action based on image Set for device

acosferreira:
[THEEDGE-3524](https://issues.redhat.com/browse/THEEDGE-3524) insert vulnerability details for edge devices and enabling update action
THEEDGE-3524

### DIFF
--- a/src/api/edge/imagesInfo.js
+++ b/src/api/edge/imagesInfo.js
@@ -77,29 +77,32 @@ export const getImageSets = ({ query }) => {
   if (query === '') {
     query = { limit: 20, offset: 0, sort_by: '-updated_at' };
   }
-  const q = getTableParams(query);
-  return instance.get(`${EDGE_API}/image-sets/view?${q}`);
+  const params = getTableParams(query);
+  return instance.get(`${EDGE_API}/image-sets/view?${params}`);
 };
 
-export const getTableParams = (q) => {
-  if (q === undefined) {
+export const getTableParams = (param) => {
+  if (param === undefined) {
     return '';
   }
-  const query = Object.keys(q).reduce((acc, curr) => {
+  const query = Object.keys(param).reduce((acc, curr) => {
     let value = undefined;
     if (
-      typeof q[curr] === 'object' &&
-      typeof q[curr].length === 'number' &&
-      q[curr].length > 0
+      typeof param[curr] === 'object' &&
+      typeof param[curr].length === 'number' &&
+      param[curr].length > 0
     ) {
-      value = q[curr].reduce(
+      value = param[curr].reduce(
         (multiVals, val) =>
           multiVals === '' ? `${curr}=${val}` : `${multiVals}&${curr}=${val}`,
         ''
       );
     }
-    if (['string', 'number'].includes(typeof q[curr]) && q[curr] !== '') {
-      value = `${curr}=${q[curr]}`;
+    if (
+      ['string', 'number'].includes(typeof param[curr]) &&
+      param[curr] !== ''
+    ) {
+      value = `${curr}=${param[curr]}`;
     }
     return value === undefined
       ? acc
--- a/src/api/edge/imagesInfo.js
+++ b/src/api/edge/imagesInfo.js
@@ -3,27 +3,27 @@ import { instance } from '@redhat-cloud-services/frontend-components-utilities/i
 
 export const getImageSet = ({
   id,
-  q = {
+  params = {
     limit: 10,
     offset: 0,
     sort_by: '-created_at',
   },
 }) => {
-  const query = getTableParams(q);
+  const query = getTableParams(params);
   return instance.get(`${EDGE_API}/image-sets/${id}?${query}`);
 };
 
 export const getImageSetViewVersions = ({
   imageSetID,
-  query = {
+  params = {
     limit: 20,
     offset: 0,
     sort_by: '-created_at',
   },
 }) => {
-  const q = getTableParams(query);
+  const query = getTableParams(params);
   return instance.get(
-    `${EDGE_API}/image-sets/view/${imageSetID}/versions?${q}`
+    `${EDGE_API}/image-sets/view/${imageSetID}/versions?${query}`
   );
 };
 
@@ -40,24 +40,24 @@ export const getImagePackageMetadata = (id) => {
 };
 
 export const fetchEdgeImages = (
-  q = {
+  params = {
     limit: 20,
     offset: 0,
     sort_by: '-created_at',
   }
 ) => {
-  const query = getTableParams(q);
+  const query = getTableParams(params);
   return instance.get(`${EDGE_API}/images?${query}`);
 };
 
 export const fetchEdgeImageSets = (
-  q = {
+  params = {
     limit: 20,
     offset: 0,
     sort_by: '-created_at',
   }
 ) => {
-  const query = getTableParams(q);
+  const query = getTableParams(params);
   return instance.get(`${EDGE_API}/image-sets?${query}`);
 };
 
--- a/src/api/edge/imagesInfo.js
+++ b/src/api/edge/imagesInfo.js
@@ -1,91 +1,16 @@
 export const EDGE_API = '/api/edge/v1';
-import { instance } from '@redhat-cloud-services/frontend-components-utilities/interceptors/interceptors';
 import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
-export const axiosInstance = () => useAxiosWithPlatformInterceptors();
-export const getImageSet = ({
-  id,
-  params = {
-    limit: 10,
-    offset: 0,
-    sort_by: '-created_at',
-  },
-}) => {
-  const query = getTableParams(params);
-  return axiosInstance.get(`${EDGE_API}/image-sets/${id}?${query}`);
-};
 
-export const getImageSetViewVersions = ({
-  imageSetID,
-  params = {
-    limit: 20,
-    offset: 0,
-    sort_by: '-created_at',
-  },
-}) => {
-  const query = getTableParams(params);
-  return axiosInstance.get(
-    `${EDGE_API}/image-sets/view/${imageSetID}/versions?${query}`
-  );
-};
-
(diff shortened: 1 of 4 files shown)


mgold1234:
can you please squash sll commits?

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
...
+    if (deviceData?.Device?.UpdateAvailable) {
+      return 'updateDevice';
+    }
+    if (newImageStatus === 'BUILDING' || newImageStatus === 'CREATED') {
+      return 'imageBuilding';
+    }
+    return 'updateImage';
+  };
+
+  // useEffect(() => {
```
adonispuente, line 60: We should remove this comment

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
...
+  const [newImageStatus, setNewImageStatus] = useState(null);
+  useEffect(() => {
+    if (CVEs?.data?.length > 0) {
+      const data = getActiveAlert(CVEs, deviceData, newImageStatus, imageId);
+      setActiveAlert(data);
+    }
+  });
+  useEffect(() => {
+    (async () => {
+      const a = await getDevice(inventoryId);
```
adonispuente, line 27: Can we rename this 'a' to something more meaningful? maybe something like deviceInfo? doesnt necissarly have to be that its just that 'a' doesnt mean anything

src/api/edge/imagesInfo.js
```diff
@@ -0,0 +1,116 @@
...
+
+export const getImageById = ({ id }) => {
+  return instance.get(`${EDGE_API}/images/${id}/details`);
+};
+
+export const getImageSets = ({ query }) => {
+  if (query === '') {
+    query = { limit: 20, offset: 0, sort_by: '-updated_at' };
+  }
+  const q = getTableParams(query);
```
adonispuente, line 80: Maybe instead of 'q' this is named 'params'. same thing with the prop passed in on line 53

src/api/edge/imagesInfo.js
```diff
@@ -0,0 +1,116 @@
...
+  return instance.get(`${EDGE_API}/image-sets/view?${q}`);
+};
+
+export const getTableParams = (q) => {
+  if (q === undefined) {
```
adonispuente, line 85: same here

src/api/edge/imagesInfo.js
```diff
@@ -0,0 +1,116 @@
+export const EDGE_API = '/api/edge/v1';
+import { instance } from '@redhat-cloud-services/frontend-components-utilities/interceptors/interceptors';
```
bastilian, line 2: This shouldn't be used directly like this. 

The axios instance returned by this will have interceptors attached that will use `chrome` functions via the window object. 

Instead the `useAxiosWithPlatformInterceptors` should be used, which will use `chrome` functions via the `useChrome` hook.

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
...
+        variant="info"
+        isInline
+        title="To remediate CVEs, update the image."
+        actionLinks={
+          <Button
+            className="pf-u-mt-sm"
+            isSmall
+            onClick={(e) => {
+              e.preventDefault();
+              window.location.href = `insights/image-builder/manage-edge-images/${imageSetId}?update_image=true`;
```
adonispuente, line 82: instead of window.location, we should use something like useNavigate or useHistory depending on the react version of the app.

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
...
+        appName="vulnerability"
+        module="./SystemDetail"
+        getRegistry={getRegistry}
+        customIntlProvider
+        entity={{ id: inventoryId }}
+        canSelect={false}
+        canEditPairStatus={false}
+        canManageColumns={false}
+        linkToCustomerPortal
+        defaultColumns={[
```
adonispuente, line 144: This is more of a nitpick, but we typically dont set columns like this in the prop. We define it in a variable and then slap that variable in the prop. Something like 
`defaultColumns={builtColumns}`

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
...
+          'synopsis',
+          'public_date',
+          'impact',
+          'cvss_score',
+          'advisory',
+        ]}
+        filters={[
```
adonispuente, line 151: same here

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
 import React from 'react';
```
bastilian, line 1: If this is going to become a bigger component we could put it into it's own "VulnerabilityEdgePrompt" folder under components.

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,39 +1,174 @@
...
+  // useEffect(() => {
+  //   !CVEs?.isLoading &&
+  //     !CVEs?.meta?.filter &&
+  //     setUpdateCveModal((prevState) => ({
+  //       ...prevState,
+  //       cveCount: CVEs?.data?.length,
+  //     }));
+  // }, [CVEs, deviceData, newImageStatus, imageId]);
+
+  const alerts = {
```
bastilian, line 69: These could be in their own component files instead of defining them within another component.

adonispuente (changes requested):
Along with the comments I left, Could you update the description of the PR to include testing steps, and what we should expect? 
Could you also include the ticket associated with the PR in the description and commit hash.
Ideally the commit hash would read something like `feat(ticketNum): short description'

src/components/VulnerabilityEdgePrompt.js
bastilian: +1 `useNavigate`

src/components/VulnerabilityEdgePrompt.js
mkholjuraev: I believe that editing this component is not expected. This is consumed by other components as it is.

acosferreira: THis component is used only if its a edge host.
As soon we start to port our code, this file should not exists.

mgold1234:
@acosferreira can you add the ticket number to description?

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -1,38 +1,101 @@
...
-
+import { useContext, useEffect, useState } from 'react';
+import { useNavigate, useParams } from 'react-router-dom';
+import RegistryContext from '../store/registeryContext';
+import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
+import { getDevice, getImageSet } from '../api/edge/imagesInfo';
+import VulnerabilityEdgeAlert from './VulnerabilityEdgeAlert';
 const VulnerabilityEdgePrompt = () => {
-  const chrome = useChrome();
+  // consxt navigate = useNavigate();
```
adonispuente, line 9: Lets remove this too

adonispuente (commented):
This looks much better! I think all thats left that should change is that instance import in the imagesInfo file, then some minor comments I left. Otherwise great job!

src/api/edge/imagesInfo.js
```diff
@@ -0,0 +1,120 @@
...
+  id,
+  params = {
+    limit: 10,
+    offset: 0,
+    sort_by: '-created_at',
+  },
+}) => {
+  const query = getTableParams(params);
+  return axiosInstance.get(`${EDGE_API}/image-sets/${id}?${query}`);
+};
```
bastilian, line 15: This should rather be something like:

```js
 const useGetImageSet = () => {
   const axios = useAxiosWithPlatformInterceptors();
   
   return (id) => 
    axios.get(`${EDGE_API}/image-sets/view/${id}`);
 }
```

to be a proper hook and not break React's rules of hooks.

In components where `getImageSet` needs to be called the function returned by the `useGetImageSet` hook should be called, like:


```js
const Components = () => {
    const getImageSet = useGetImageSet();

    ....
    useEffect(() => {
      (async () => {
        const resp = await getImageSet(params);
        ...
      })();
    }, []);

   return <>...</>
}

Comment (bastilian) [reply src/api/edge/imagesInfo.js:15]: This ^ applies to all functions in this file.

File: src/components/VulnerabilityEdgeAlert.js [hunk 13]
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import { Alert, Button } from '@patternfly/react-core';
+import { InProgressIcon } from '@patternfly/react-icons';
+const VulnerabilityEdgeAlert = ({
+  activeAlert,
+  navigateProp,
```
bastilian, line 6: This here should not be needed to be passed down, the components, InventoryDetails & VulnerabilityAlert, are in the same router context and calling it here directly will yield the same `navigate` behaviour. 

Use `useNavigate` here directly.

src/routes/InventoryDetail.cy.js
```diff
@@ -51,11 +51,9 @@ describe('renders correctly', () => {
     cy.get('[data-cy="compliance-tab"]')
       .parent('.pf-c-tabs__item.pf-m-disabled')
       .should('have.length', 1);
-
     cy.get('[data-cy="vulnerabilities-tab"]')
       .parent('.pf-c-tabs__item')
-      .click();
-    cy.get('[data-cy="vulnerability-edge-prompt"]').should('have.length', 1);
+      .should('have.length', 1);
```
bastilian, line 56: What is the reason for removing the `click`ing here?

acosferreira: before we had a link to go to edge management vulnerability tab, now we are opening the list of cves

adonispuente (approved):
I believe everything has been addressed, LGTM! Thank you!

bastilian (approved):
Nice work! Thank you @acosferreira!

acosferreira:
@mkholjuraev fyi

PR #2018, opened 2023-09-13
fix(InventoryTable): Prevent fetching OSs without permission

bastilian:
In cases where the user does not have permissions to read the inventory we should not fetch OS versions.


src/components/filters/useOperatingSystemFilter.js
```diff
@@ -34,7 +34,11 @@ const useOperatingSystemFilter = (apiParams = []) => {
   const [groups, setGroups] = useState([]);
 
   useEffect(() => {
-    dispatch(fetchOperatingSystems(apiParams));
+    if (typeof hasAccess === 'undefined') {
+      dispatch(fetchOperatingSystems(apiParams));
+    } else if (typeof hasAccess !== 'undefined' && hasAccess === true) {
```
bastilian, line 39: This may look weird. I know. However, this is in order to keep compatibility for now with places that use the hook but do not provide a `hasAccess` argument.

src/components/filters/useOperatingSystemFilter.js
```diff
@@ -34,7 +34,11 @@ const useOperatingSystemFilter = (apiParams = []) => {
   const [groups, setGroups] = useState([]);
 
   useEffect(() => {
-    dispatch(fetchOperatingSystems(apiParams));
+    if (typeof hasAccess === 'undefined') {
```
gkarat, line 37:
```suggestion
    if (hasAccess === undefined) {
```

src/components/filters/useOperatingSystemFilter.js
gkarat, line 39:
```suggestion
    } else if (hasAccess !== undefined && hasAccess === true) {
```

gkarat (approved):
LGTM! Just a minor suggestions (I think the condition can be written simpler). But works locally

PR #2019, opened 2023-09-13
Tab change

adonispuente:
This PR just adds a conditional on the inventory details page so that the correct tabs render when a system is a CentOs system as opposed to undefined or RHEL system. 

To test, use the insights-qa acc.
Navigate to /systems and filter for a system with display name 'adonis' -> this is a CentOs system
select the system and on invDetails you should see the tabs update as the entity is loaded in. 

Grab another random RHEL system and verify that the tabs are kept the same


src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -38,6 +39,15 @@ const ApplicationDetails = ({
   const disabledApps = useSelector(
     ({ systemProfileStore }) => systemProfileStore?.disabledApps
   );
+  const centOsAppList = [
```
bastilian, line 42: We could make something that works for both RHEL and CentOS, like:

```
const osAppList = {
"centos-linux: [{ title: 'General Information', ...}],
"rhel": [{ title: 'General Information' ...}, ...],
};
```

This could be somewhere outside the component more surfaced in a constants file. 

To match them here in the component operating system name would just need to be slugified, like 

```
const osSlug = entity?.system_profile?.operating_system?.name.replace(' ', '-').toLowerCase();
```

and then `osAppList[osSlug]`.

adonispuente: Originally I had some conditional logic in the const, but the issue is that this depends on the entity info, so we have to wait for that to load in. With this idea I would run into the same issue, the app would try to load in this osAppList before that osAppList meant anything.

however we cooould do this, but originally set things to an empty array until this got loaded in. Do you think that routes better? 

We cant make to components async because of react rules, so theres no await here.

bastilian: I think it is totally fine if the app list array without the OS name would be empty.

The inventory details higher up shouldn't actually attempt to render the section before it has the entity information, or? 

Even if it does now, even the header will be empty or render some skeleton loader before it has any all the info.

adonispuente: It does, and the logic for this would need to be done in InventoryDetail as opposed to ApplicationDetails for it to make sense to change since thats where a lot of the logic for appList is made (a few maps and alterations). 

However, because we cant make it async, we cant rely on the entity to finish loading before the alteration functions are preformed. The osSlug will be undefined at first, and that will cause appList[osSlug] not to work. 

I can make all the alterations on the appList conditionally set afterwords, but theres a useMemo that will cause a scalprum error if call it conditionally. and if we use that useMemo with appList[osSlug], it'll break

bastilian: > because we cant make it async

What do you mean by that?

adonispuente: Post gm call, I believe this new commit slaps on the alternative solution properly

src/routes/InventoryDetail.js
```diff
@@ -98,30 +108,37 @@ const Inventory = () => {
...
+        return app;
+      });
+
+    setAvailableApps(newApps);
+  }, [entity, cloudProvider]);
 
   useEffect(() => {
     if (searchParams.get('appName') === null) {
-      setSearchParams('appName', appList[0].name);
+      setSearchParams('appName', appList['RHEL'][0].name);
```
bastilian, line 141: This should probably also be something like `appList[osSlug][0].name`.

adonispuente: Since this runs once on page load, we need this to be a set value from the jump. We dont want to rerender the page off of the params here. Since General information will most likely almost come first, I kept it as is. 
If we really want to keep away from that, what about setting it to 'general_information'?

bastilian: The `searchParams.get('appName') === null` should prevent this state update to happen, even if `entity` is a dependency of this effect as well to get the right OS name/slug.


TBH. I think this useEffect here is a bit odd. I think it should either update the state, or navigate, but not both. When the navigate is called, it probably throws away this state update here, but it will then have a correct tab selected, because it then has the `appName` set in the URL.


I believe instead of doing any state update or navigate here, we should just default to the first app in the app array of each OS.

adonispuente: I dont fully understand what you mean in the last sentence. What exactly are you saying we are defaulting? Are you saying the appName in the url should default to that?

bastilian: @adonispuente What I mean is, if there is no `appName` in the URL initially, when a user comes first to the page, we should not set any `appName` parameter in the URL, but simply default to the first "app" in the array for each OS. We should only set a appName in the URL if a user does click a tab.

src/routes/InventoryDetail.js
```diff
@@ -98,30 +108,37 @@ const Inventory = () => {
...
+        app['data-cy'] = `${app.name}-tab`;
+        app.name === 'ros' && {
+          ...app,
+          isVisible: cloudProvider === 'aws',
+        };
+        return app;
+      });
+
+    setAvailableApps(newApps);
+  }, [entity, cloudProvider]);
```
mkholjuraev, line 137: An idea: Can we move this set of code to decide the new apps list out of the hook? Because the entity should not change when the user is viewing the page, also cloudProvider does not change consequently. They change only when the user goes to another page or views another host detail intentionally, leaving no necessity for the hook. Moving the code out of the hook gives more speed as we will not have to set a state change

adonispuente: On page load the entity takes a bit to come in, initially without the osSlug we could have done that, but now that we wait on entity to decide what apps are displayed, this is necessary

src/routes/InventoryDetail.js
```diff
@@ -94,42 +104,38 @@ const BreadcrumbWrapper = ({ entity, inventoryId, entityLoaded }) => (
 const Inventory = () => {
   const chrome = useChrome();
   const { inventoryId } = useParams();
+  // eslint-disable-next-line no-unused-vars
```
bastilian, line 107: sus. :)

adonispuente: ahh youre right, for some reason I had tried wrapping it in {}. And yeah it was because of that

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -105,7 +106,7 @@ const ApplicationDetails = ({
         </Tabs>
       </section>
       <section>
-        {activeTabs?.length &&
+        {activeTabs?.length > 0 &&
```
bastilian, line 109: We could wrap the whole `Fragment` in this and either render nothing or a `Spinner`, while it has no `entity` and no `activeTabs`.

bastilian (approved):
Nice! Thank you @adonispuente!

src/components/InventoryDetail/ApplicationDetails.js
adonispuente: Its small, but I like this change. I've gotten a few messages to slap this on qa already so im going to merge and then make a separate PR for this.

bastilian: @adonispuente Oh, when you do that. a minor thing we could also improve... the `React.Fragment` is redundant. All child components are already wrapped in one `section`.

PR #2023, opened 2023-09-21
chore(loadingState): Loading state in AppDetails

adonispuente:
Simple PR just introduces the spinner for when entitys is not loaded and there are no tabs yet. 
This also adds a default value for the osSlug in the case where the Os system is undefined/ not returned in the api

### DIFF
--- a/src/components/InventoryDetail/ApplicationDetails.js
+++ b/src/components/InventoryDetail/ApplicationDetails.js
@@ -77,65 +77,70 @@ const ApplicationDetails = ({
 
   return (
     <React.Fragment>
-      <section className="pf-u-pr-lg pf-u-pl-lg pf-u-background-color-100-on-md">
-        <Tabs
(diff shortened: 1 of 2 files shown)


src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -105,9 +105,9 @@ const ApplicationDetails = ({
           ))}
         </Tabs>
       </section>
-      <section>
-        {activeTabs?.length > 0 &&
-          activeTabs?.map((item) => {
+      {activeTabs?.length > 0 ? (
```
bastilian, line 108: The condition could already start on line 80, and wrap really all of the component. Then the odd "<>" when it loads form the tabs should be gone too.

bastilian (approved):
Looks good to merge! Thank you @adonispuente!

PR #2024, opened 2023-09-21
feat(THEEDGE-3524): remove advisory column

acosferreira:
From cve detail, removed the advisoy colum list and filter.
Fixed the imageSetId Parameter
image


src/api/edge/imagesInfo.js
```diff
@@ -5,7 +5,8 @@ export const useGetImageSet = () => {
   const axios = useAxiosWithPlatformInterceptors();
   const query = { limit: 10, offset: 0, sort_by: '-created_at' };
   const params = getTableParams(query);
-  return (id) => axios.get(`${EDGE_API}/image-sets/${id}?${params}`);
+  return (p) =>
```
bastilian, line 8: This is still getting passed an `id` or? 

In general we should avoid using one character variable names and try to be more descriptive.

src/components/VulnerabilityEdgePrompt.js
```diff
@@ -46,7 +39,7 @@ const VulnerabilityEdgePrompt = () => {
       setImageSetID(device?.ImageInfo?.Image?.ImageSetID);
       const params = {
         id: 'view',
```
mkholjuraev, line 41: Nitpick: this looks obsolete.

mkholjuraev (approved):
LGTM! Did not face issues while testing. As @acosferreira is on PTO, I will remove that line and then merge

src/components/VulnerabilityEdgePrompt.js
mkholjuraev: Weird, if I remove this line of code update page does not get opened. However, I am sure that this `id` is not used anywhere.

PR #2029, opened 2023-09-25
feat(RHIN-447): Remaining Items for M1 Hostaleness

adonispuente:
This PR includes
- Toast alerts when a user saves Save/Failure
- Reset values coming from the backend
- Permissions/Rbac views. 
https://www.sketch.com/s/dd09e06a-9bb3-4cda-9bf3-2fe92d949145 
 Link in Empty state is currently pointing to nothing -> waiting on UX

### DIFF
--- a/src/components/InventoryHostStaleness/BaseDropDown.js
+++ b/src/components/InventoryHostStaleness/BaseDropDown.js
@@ -40,6 +40,7 @@ const BaseDropdown = ({
     setSelected(currentItem);
     formValidation(newFormValues, setIsFormValid);
   }, [edit, currentItem]);
+
   return (
     <React.Fragment>
       <Flex direction={{ default: 'column' }} gap={{ default: 'gapNone' }}>
--- a/src/components/InventoryHostStaleness/BaseDropDown.js
+++ b/src/components/InventoryHostStaleness/BaseDropDown.js
@@ -98,7 +98,7 @@ BaseDropdown.propTypes = {
   onSelect: PropTypes.bool,
   placeholder: PropTypes.string,
   title: PropTypes.string,
-  currentItem: PropTypes.string,
+  currentItem: PropTypes.any,
   filter: PropTypes.object,
   newFormValues: PropTypes.any,
   setNewFormValues: PropTypes.any,
--- a/src/components/InventoryHostStaleness/HostStalenessCard.js
+++ b/src/components/InventoryHostStaleness/HostStalenessCard.js
@@ -38,6 +38,7 @@ const HostStalenessCard = () => {
     immutable_stale_warning_delta: '',
     immutable_culling_delta: '',
   });
+  const [defaultApiValues, setDefaultApiValues] = useState({});
 
   const axios = useAxiosWithPlatformInterceptors();
   const [newFormValues, setNewFormValues] = useState(filter);
@@ -90,9 +91,24 @@ const HostStalenessCard = () => {
     setFilter(newFilter);
     setNewFormValues(newFilter);
   };
+  //keeps track of what default the backend wants
+  const fetchDefaultValues = async () => {
+    let results = await axios
+      .get(`${INVENTORY_API_BASE}/account/staleness/defaults`)
+      .then((res) => res);
+    console.log(results, 'res here');
+
+    let newFilter = {};
+    hostStalenessApiKeys.forEach(
+      (filterKey) =>
+        (newFilter[filterKey] = secondsToDaysConversion(results[filterKey]))
+    );
+    setDefaultApiValues(newFilter);
+  };
 
   const batchedApi = async () => {
     fetchStalenessData();
+    fetchDefaultValues();
     setIsLoading(false);
   };
 
@@ -156,6 +172,7 @@ const HostStalenessCard = () => {
                   setNewFormValues={setNewFormValues}
                   isFormValid={isFormValid}
                   setIsFormValid={setIsFormValid}
+                  defaultValues={defaultApiValues}
                 />
               </Tab>
               <Tab
@@ -183,6 +200,7 @@ const HostStalenessCard = () => {
                   setNewFormValues={setNewFormValues}
                   isFormValid={isFormValid}
                   setIsFormValid={setIsFormValid}
+                  defaultValues={defaultApiValues}
                 />
               </Tab>
             </Tabs>
--- a/src/components/InventoryHostStaleness/HostStalenessCard.js
+++ b/src/components/InventoryHostStaleness/HostStalenessCard.js
@@ -13,6 +13,7 @@ import {
   TabTitleText,
   Tabs,
   Title,
+  Tooltip,
 } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import TabCard from './TabCard';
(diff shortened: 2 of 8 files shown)


src/components/InventoryHostStaleness/BaseDropDown.js
```diff
@@ -98,7 +99,7 @@ BaseDropdown.propTypes = {
   onSelect: PropTypes.bool,
   placeholder: PropTypes.string,
   title: PropTypes.string,
-  currentItem: PropTypes.string,
+  currentItem: PropTypes.any,
```
bastilian, line 102: 🕵️‍♂️ Why this?

adonispuente: newFormValues contains the 'id' string passed down from the api. So when we pass down newFormValues on page load, its an object of key/values of 6 numbers but one key === a string

bastilian: Then let's maybe use `PropTypes.oneOf([PropTypes.object, PropTypes.string])`. It is weird though that it changes type during runtime. If it is possible to correct that and make sure that currentItem is always either a string or an object, but doesn't change.

adonispuente: It seems the decision to have this all be ints has been made on the backend, so ive changed this to number. I wont remove the string conversions until that backend peace is finished. That way if the review for this is completed beforehand, QE can test the full version, and if its not then I can just add on to this PR

src/components/InventoryHostStaleness/HostStalenessNoAccess.js
```diff
@@ -8,14 +8,14 @@ import {
...
+const HostStalenessNoAccess = () => {
   return (
     <EmptyState>
       <EmptyStateIcon icon={LockIcon} />
       <Title headingLevel="h5" size="lg">
-        Acess permissions needed
+        Access permissions needed
       </Title>
-      <EmptyStateBody>
+      <EmptyStateBody className="pf-u-mb-xl">
```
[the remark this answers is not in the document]
gkarat: Also, the width is more narrow I guess according to the mocks. Like setting it to the M/L size would fit better

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -116,14 +174,27 @@ const HostStalenessCard = () => {
             </p>
             <Flex className="pf-u-mt-md">
               <Title headingLevel="h6">System configuration </Title>
```
gkarat, line 176:
```suggestion
              <Title headingLevel="h6">System configuration</Title>
```

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -116,14 +174,27 @@ const HostStalenessCard = () => {
...
+                  }}
+                >
+                  Edit
+                </Button>
+              ) : (
+                <Tooltip content="You do not have the Inventory staleness and culling viewer role required to perform this action. Contact your org admin for access.">
+                  <Button
+                    variant="link"
+                    style={{ color: '#6A6E73', 'padding-left': '0px' }}
+                    disabled={!canModifyHostStaleness}
```
gkarat, line 192:
```suggestion
                    style={{ 'padding-left': '0px' }}
                    isDisabled
```

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -116,14 +174,27 @@ const HostStalenessCard = () => {
...
-                }}
-              >
-                Edit
-              </a>
+              {canModifyHostStaleness ? (
+                <Button
+                  variant="link"
+                  role="button"
+                  onClick={() => {
+                    setEdit(!edit);
```
gkarat, line 182: nit-pick: maybe some more explicit naming for the state variable would be better like `isEditing`

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -244,4 +317,8 @@ const HostStalenessCard = () => {
   );
 };
 
+HostStalenessCard.propTypes = {
+  canModifyHostStaleness: PropTypes.bool,
```
gkarat, line 321:
```suggestion
  canModifyHostStaleness: PropTypes.bool.isRequired,
```

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -68,13 +64,61 @@ const HostStalenessCard = () => {
         filterKey !== 'id' &&
         (apiData[filterKey] = daysToSecondsConversion(newFormValues[filterKey]))
     );
+
     //system_default means the account has no record, therefor, post for new instance of record.
     if (filter.id === 'system_default') {
-      axios.post(`${INVENTORY_API_BASE}/account/staleness`, apiData);
+      axios
```
gkarat, line 70: These API funcs are better placed under src/api.js. Also, take a look at the `instance` variable available there. It's an axios instance that you can already use to build and make the requests with.

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -68,13 +64,61 @@ const HostStalenessCard = () => {
...
+        .post(`${INVENTORY_API_BASE}/account/staleness`, apiData)
```
gkarat, line 71: Hm, why it's a patch, not post method?

src/components/InventoryHostStaleness/HostStalenessCard.js
```diff
@@ -90,15 +134,29 @@ const HostStalenessCard = () => {
     setFilter(newFilter);
     setNewFormValues(newFilter);
   };
+  //keeps track of what default the backend wants
+  const fetchDefaultValues = async () => {
+    let results = await axios
+      .get(`${INVENTORY_API_BASE}/account/staleness/defaults`)
```
gkarat, line 140: Same here: let's extract it to src/api.js

src/components/InventoryHostStaleness/HostStalenessCard.js
adonispuente: Its how the backend was designed. If the user never created an instance of staleness, the default one is applied. Once one is applied, if you try to make a post request, the api will return a failure messages because an instance already exist.

src/components/InventoryHostStaleness/HostStalenessNoAccess.js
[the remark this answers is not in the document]
adonispuente: The width was changed/talked about on a call with UX. The mocks dont reflect those changes. 
As for the variant property, that does change the width, but the class on the body is specifically to add more spacing between the link and the main text. 
The spacing is also something UX changed on the meeting

src/components/InventoryHostStaleness/HostStalenessCard.js
adonispuente: The instance version isnt resolving the promise in a way that the .catch will read. It will always hit the .then side of things. I've sent up a compromise where I move it to the api file but still use the useAxiosWithPlatformInterceptors hook

gkarat (approved):
LGTM 🚀 code-wise only one thing about the return instance thingy I mentioned in slack. other than that, all looks pretty good. Thanks @adonispuente

PR #2030, opened 2023-09-26
chore(RHIF-277): expose HybridInventoryTabs as fed-mod, preserve URLs

mkholjuraev:
This exposes the new HybirdInventoryTabs though fed-mods so that other apps can consume. Also preserves the URL parameters when a user comes back from another tab.

To test:
1. open Systems page
2. on conventional systems tab, change a page so that there is URL parameter active
3. navigate to immutable devices tab
4. navigate back to conventional tab
5. observe that the URL parameters  are still present in the URL in the conventional systems tab

### DIFF
--- a/src/components/InventoryTabs/HybridInventoryTabs.js
+++ b/src/components/InventoryTabs/HybridInventoryTabs.js
@@ -74,8 +74,6 @@ const HybridInventoryTabs = ({
     }
   };
 
-  console.log('Hellooo');
-
   return EdgeParityEnabled && hasEdgeImages ? (
     <Tabs
       className="pf-m-light pf-c-table"
--- a/src/components/InventoryTabs/HybridInventoryTabs.js
+++ b/src/components/InventoryTabs/HybridInventoryTabs.js
@@ -80,6 +80,8 @@ const HybridInventoryTabs = ({
           : hybridInventoryTabKeys.conventional.key
       }
       aria-label="Hybrid inventory tabs"
+      mountOnEnter
+      unmountOnExit
     >
       <Tab
         eventKey={hybridInventoryTabKeys.conventional.key}
--- a/src/components/InventoryTabs/HybridInventoryTabs.js
+++ b/src/components/InventoryTabs/HybridInventoryTabs.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import axios from 'axios';
 import PropTypes from 'prop-types';
 import { useLocation } from 'react-router-dom';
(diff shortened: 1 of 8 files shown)


src/components/InventoryTabs/HybridInventoryTabs.js
```diff
@@ -1,22 +1,35 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useLocation } from 'react-router-dom';
 import { Tab, TabTitleText, Tabs } from '@patternfly/react-core';
 import useFeatureFlag from '../../Utilities/useFeatureFlag';
-import useInsightsNavigate from '@redhat-cloud-services/frontend-components-utilities/useInsightsNavigate/useInsightsNavigate';
 import { hybridInventoryTabKeys } from '../../Utilities/constants';
+import { useNavigate } from 'react-router-dom';
```
bastilian, line 7: Why does this need to be returned to a raw react router `useNavigate`? 

If it is due to `useInsightsNavigate` not handling states properly than, please wait with replacing it. I am working on extending it to handle these properly.

mkholjuraev: Yup, useInsightsNavigate does not handle states. I would prefer to go with this so that I can continue with my work. Once useInsightsNavigate can handle, then easy to change..

mkholjuraev:
@RedHatInsights/team-interact Ready for a review

mkholjuraev:
@computercamplove Hey, this PR is failing in **ci.int.devshift.net PR build**. Rebuilding does not help. Something needs to be changed in IQE?

computercamplove:
> @computercamplove Hey, this PR is failing in **ci.int.devshift.net PR build**. Rebuilding does not help. Something needs to be changed in IQE?

last update was 27.09.2023 in IQE tests and nothing changed in smoke tests, another thing is ci.int had outage. going to check what's changed.

computercamplove:
@mkholjuraev i think here is no issues with IQE tests, I see in 1611 build this:
`
/container_workspace/universal_build.sh: line 17: npm: command not found`

bastilian:
@mkholjuraev With "active tab" in the last step do you mean the path or some parameter that should be present in the URL?

mkholjuraev:
> 5\. ctive

@bastilian thank you for having a look. I meant the URL params are active on the convention systems tab. I will fix this in the Test steps in the description.

bastilian:
@mkholjuraev I still don't really understand what is meant. There are no URL parameters. even when the page is changed on any tab, switching to the other tab resets all URL parameters and appends the global filter hash.

mkholjuraev:
> @mkholjuraev I still don't really understand what is meant. There are no URL parameters. even when the page is changed on any tab, switching to the other tab resets all URL parameters and appends the global filter hash.

Seems the last commit to isolate tabs affected the useNavigate behavior. navigation does not properly preserve the state. I am not yet sure how the Patternfly Tab [prop](https://www.patternfly.org/components/tabs#unmounting-invisible-children) can affect the router context. Let me investigate and fix

bastilian:
> observe that the URL parameters are still present in the URL in the conventional systems tab

There are still no URL parameters present or restored when switching tabs. However, it does appear the state persists and page or filter previously set is shown, but not populated to the URL.

https://github.com/RedHatInsights/insights-inventory-frontend/assets/7757/19f2cd21-edac-4554-b6dc-0ff847ee81ce

mkholjuraev:
Aha, this looks interesting. When I run the PR locally, everything seems ok. To make sure we are looking at the same thing, can you make sure you have the latest changes? In the following video, you can see that I have the latest changes, and URL params are seeded back after a tab change.

Screen Recording 2023-10-03 at 14.20.48

src/components/InventoryTabs/HybridInventoryTabs.js
```diff
@@ -28,57 +64,23 @@ const HybridInventoryTabs = ({
...
-                handleTabClick(undefined, hybridInventoryTabKeys.immutable.key);
-                setIsOstreeTabFocusPriority(true);
-              }
-            });
-        });
-    } catch (e) {
-      console.log(e);
-    }
-  }
+  console.log('Hellooo');
```
bastilian, line 77: 👋

AsToNlele (approved):
Great job! For a short while it goes back to page 1 and then back to the correct page, but it works great nevertheless. :smile:

mkholjuraev:
> Great job! For a short while it goes back to page 1 and then back to the correct page, but it works great nevertheless. 😄

@AsToNlele Thank you for the review! Yup, it is expected as there is 2 level of URL push that can not be avoided easly.

PR #2033, opened 2023-09-27
test(RHINENG-3032): Migrate from Enzyme to RTL

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-3032.

This mirgates all unit tests to the RTL library. 

- Many tests were moved from asserting against snapshots to specific queries depending on what each test is trying to verify. The reason is that RTL snapshot generation is different to enzyme's and does not support shallow rendering which made many tests unfunctional.
- All enzyme and its eco-system packages are removed.
- setupTests.js was cleaned and now contains only the most necessary mocks/globals.
- All tests were checked with the testing-library/react plugin for eslint.


src/components/GeneralInfo/BiosCard/BiosCard.test.js
```diff
@@ -1,11 +1,9 @@
 /* eslint-disable camelcase */
 import React from 'react';
-import { render } from 'enzyme';
-import toJson from 'enzyme-to-json';
 import BiosCard from './BiosCard';
 import configureStore from 'redux-mock-store';
 import { biosTest } from '../../../__mocks__/selectors';
-import { mountWithRouter } from '../../../Utilities/TestingUtilities';
+import { renderWithRouter } from '../../../Utilities/TestingUtilities';
```
bastilian, line 6: Maybe it would be better to move the `render` out of the `mount/renderWithRouter` (and maybe even call it `TestWrapper`) and import the `render` directly here. 

I think that it has the advantage of it being visible right away what test framework is used. Another thing is, that `renderWithRouter` adds another layer, and while I doubt it, it has the potential of introducing unwanted behaviour. 

Having a `TestWrapper` component might be a pattern we could apply to more applications. It would basically be the same as any `App.js` with all the necessary provider components.

src/components/GeneralInfo/InfrastructureCard/InfrastructureCard.test.js
```diff
@@ -75,69 +76,74 @@ describe('InfrastructureCard', () => {
...
       location.pathname = 'localhost:3000/example/ipv4';
       useParams.mockImplementation(() => ({ modalId: 'ipv4' }));
-      const wrapper = mountWithRouter(
+      renderWithRouter(
         <InfrastructureCard handleClick={onClick} store={store} />
       );
-      wrapper.find('dd a').first().simulate('click');
-      expect(onClick).toHaveBeenCalled();
+
+      await userEvent.click(
```
bastilian, line 93: Wouldn't it be better to wrap these `userEvent`s in `act`[0]?


[0] https://testing-library.com/docs/react-testing-library/api#act

src/components/GeneralInfo/BiosCard/BiosCard.test.js
gkarat: Yeah, sounds very similar to what happened to `mountWithContext` from cypress tests. I'll rewrite this part now

src/components/GeneralInfo/InfrastructureCard/InfrastructureCard.test.js
gkarat: `act` is already called internally

gkarat:
I might introduce minor changes or clean-up again, but I believe the PR is ready for review.

gkarat:
^ Did rebase against upstream/master, reworded commits missing prefix, fixed test errors that occurred after rebase

src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.test.js
```diff
@@ -125,13 +119,19 @@ describe('ConventionalSystemsTab', () => {
     systemProfileStore: { systemProfile: { loaded: false } },
   };
 
-  const mount = (children, store) =>
-    enzymeMount(
-      <ReactRouterDOM.MemoryRouter>
+  const render = (children, store) =>
+    _render(
```
bastilian, line 123: Usually when prepending `_` it is meant to signal that this const/variable is not actually used.

Maybe the function could be named `renderWithProviders` to would allow avoiding the renaming of the `render` import?

src/Utilities/DeleteModal.test.js
```diff
@@ -1,48 +1,33 @@
...
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 describe('EntityTable', () => {
   describe('DOM', () => {
-    it('should render correctly', () => {
-      const wrapper = shallow(<DeleteModal />);
-      expect(toJson(wrapper)).toMatchSnapshot();
-    });
+    it('should render correctly with no system', () => {
```
bastilian, line 9: The snapshot for this test doesn't really tell anything anymore.[0]
In this case, it should be safe to just remove it.
If in case of no systems the DeleteModal would/should render a warning we could query that, but that is not the case here. 


[0] https://github.com/RedHatInsights/insights-inventory-frontend/pull/2033/files#diff-64d05750a711d29ab9558378f56ac984e52c92e258eafb6402ccdf0fa4366934R204-R207

src/Utilities/TagsModal.test.js
```diff
@@ -26,15 +26,39 @@ describe('TagsModal', () => {
...
+        },
+      });
+      const view = render(
         <Provider store={store}>
           <TagsModal />
         </Provider>
       );
+
+      // TODO: improve skeleton accessibility
+      // baseElement is used since the PF modal is rendered outside parent div
```
bastilian, line 45: I'm a bit confused by this. Is this meant to mean to add a test for the loading state in later on?

src/components/GeneralInfo/EditButton/EditButtonWithAccess.test.js
```diff
@@ -29,24 +27,20 @@ describe('EditButton with access', () => {
   });
 
   it('enables with permission', () => {
-    const wrapper = mount(<EditButton onClick={onClick} link={link} />);
+    render(<EditButton onClick={onClick} link={link} />);
 
-    expect(wrapper.find(Tooltip)).toHaveLength(0);
-    expect(wrapper.find(PencilAltIcon)).toHaveLength(1);
-    expect(wrapper.find('a').props().href).toEqual(
+    expect(screen.getByRole('link', { name: /edit/i })).toBeVisible();
```
bastilian, line 32: Should this not query the Tooltip? ... or is that not possible anymore with the HTML rendered?

src/components/GeneralInfo/OperatingSystemCard/OperatingSystemCard.test.js
```diff
@@ -41,30 +49,24 @@ describe('OperatingSystemCard', () => {
...
-        entity: {},
-      },
-    });
-    const wrapper = mountWithRouter(<OperatingSystemCard store={store} />);
-    expect(toJson(wrapper)).toMatchSnapshot();
+    expect(
+      screen
+        .getAllByRole('definition')
+        .reduce((prev, cur) => prev.concat(cur.textContent), '')
+    ).toBe('test-releasetest-kerneltest-archNot available0 modules');
```
bastilian, line 69: Let's at least separate these with a `,`. :D

src/components/InventoryDetail/DetailWrapper.test.js
```diff
@@ -80,61 +85,68 @@ describe('DetailWrapper', () => {
...
+      render(
         <Provider store={store}>
           <DetailWrapper />
         </Provider>
       );
-      expect(
-        toJson(wrapper.find('BasicInfo'), { mode: 'shallow' })
-      ).toMatchSnapshot();
+
+      expect(BasicInfo).toHaveBeenCalledWith(
```
bastilian, line 127: This test is a bit odd. Maybe we can avoid the mocking of `BasicInfo` and let it just render as normal and query for the effects of the props? For example, query if tags are or are not shown, when passing in `showTags`.

src/components/InventoryTable/InventoryList.test.js
```diff
@@ -97,19 +114,17 @@ describe('InventoryList', () => {
...
-      expect(onRefreshData).toHaveBeenCalled();
+      rerender(
+        <Cmp
+          sortBy={{
+            ...sortBy,
+            direction: 'desc',
+          }}
+        />
+      );
+      expect(onRefreshData).toHaveBeenCalledTimes(1);
```
bastilian, line 127: If the mock has been cleared before the test, wouldn't onRefreshData be expected to have been called twice, once without direction, when it initially loads, and the second time with direction?

src/Utilities/TagsModal.test.js
gkarat: Hmm, I think I wanted to improve the skeleton rows attributes and add something like aria-busy=true or other aria label to mark the rows as loading. With the idea that this would also help while testing with RTL. Since the component is imported from FEC, I decided to put this off and thus left the snapshot check.

gkarat: I know see that the <table> has aria-label="Loading" attribute. I will replace the snapshot check with just asserting this attribute, but also leave the note so that we can also query rows (let's say, "expect 15 skeleton rows" rendered) in the future.

src/components/GeneralInfo/EditButton/EditButtonWithAccess.test.js
gkarat: Here the button is active and doesn't have any supportive tooltips. Though, I check the tooltips content at EditButtonWithNoAccess.test.js (the other neighbour test file).

src/components/filters/useGroupFilter.test.js
```diff
@@ -97,39 +99,45 @@ describe('some groups available', () => {
...
-      expect(config.filterValues.children).toMatchInlineSnapshot(`
-        <SearchableGroupFilter
-          initialGroups={Array []}
+    const { result, waitForNextUpdate } = renderHook(() =>
+      useGroupFilter(true)
+    );
+
+    await waitForNextUpdate();
+    const [config] = result.current;
+    expect(config.filterValues).toMatchInlineSnapshot(`
```
bastilian, line 111: Could we use `toMatchObject` here as well? All other hook tests appear to use it instead of an inline snapshot.

src/components/InventoryTable/InventoryTable.test.js
```diff
@@ -1,296 +1,616 @@
...
+
+const errorState = {
+  entities: {
+    ...initialState.entities,
+    error: new Error('Loading error'),
+  },
+};
+
+describe('InventoryTable', () => {
+  useFeatureFlag.mockReturnValue(true); // assuming all features are enabled
```
bastilian, line 202: Defaulting to disabling all features might be better. When introducing a new feature all tests might be affected. But with them disabled by default, we can enable them only in tests related to that feature and still verify that other parts still work as intended.

src/components/GeneralInfo/dataMapper/dataMapper.test.js
```diff
@@ -38,7 +36,51 @@ it('diskMapper', () => {
         mounttype: 'type',
       },
     ])
-  ).toMatchSnapshot();
+  ).toMatchInlineSnapshot(`
```
bastilian, line 39: We could instead of using an inline snapshot query the specific cells that should be shown with these arguments passed in. Maybe there is a list somewhere that his can even be based on and made dynamic in order to test any variation?

src/components/GeneralInfo/SystemCard/SystemCard.test.js
```diff
@@ -1,27 +1,34 @@
...
 
 const location = { pathname: 'some-path' };
-const removeLabelledBy = ({
-  'aria-labelledby': labelledBy,
-  'aria-describedby': describedby,
-  id: id,
-  ...restProps
-}) => restProps;
 
 jest.mock('react-router-dom', () => ({
```
bastilian, line 33: Unless it needs a specific location or params, I think we can remove this mock, since it should work fine when rendering with a router instance.

src/components/GeneralInfo/LoadingCard/LoadingCard.test.js
```diff
@@ -1,8 +1,10 @@
+import '@testing-library/jest-dom';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
-import toJson from 'enzyme-to-json';
+import { TestWrapper } from '../../../Utilities/TestingUtilities';
```
bastilian, line 5: I'm feeling a bit odd about `renderWithRouter` vs. `TestWrapper`. Both basically accomplish the same thing, but there is no real pattern to when to use which.

Maybe we can utilise one or the other and that consistently throughout to avoid confusion? 

Personally, I'd be more in favour of `TestWrapper` as it also provides other things. 
I'm also a bit iffy about wrapping `render`.

src/components/GeneralInfo/OperatingSystemCard/OperatingSystemCard.test.js
gkarat: I will replace these strings with array so that the tests are more readable. I initially departed from the idea to use inline snapshots there, but some issues arose and I ended up with having this weird string 😄 Now after another fresh look this makes more sense to use an array

src/components/GeneralInfo/dataMapper/dataMapper.test.js
gkarat: I think these tests are already straightforward and check the return value for each function. I am not sure querying some fields will improve the tests much. Though, I will leave a TODO note to try to implement dynamic tests and cover more cases

gkarat: It's convenient to use inline snapshots in this scenario when you assert data structures. If the implementation changes, the tests will catch the difference and if it's anticipated, it's very quick to update the snapshots.

src/components/InventoryTable/InventoryList.test.js
gkarat: onRefreshData, in this scenario and given the props, is called only when there is a change to sortBy prop: https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/InventoryTable/InventoryList.js#L46. This is why on the initial render it's not triggered.

src/Utilities/TagWithDialog.test.js
```diff
@@ -28,30 +29,42 @@ describe('EntityTable', () => {
...
+
+      await userEvent.click(
+        screen.getByRole('button', {
+          name: '10',
+        })
       );
-      wrapper.find('button').first().simulate('click');
       const actions = store.getActions();
-      expect(actions.length).toBe(2);
+      expect(actions.length).toBe(3);
```
bastilian, line 67: Just a note, nothing to do here.

These checks on the length of actions is odd. It probably will also fail each time a new test above is added that dispatches actions. In the future we should maybe replace this with something more deterministic.

gkarat: Yeah, good catch. I also used `shouldDispatch` together within this PR to assert on specific actions. I will replace it in the PF5/React 18 PR (based on this branch).

gkarat:
Follow-up: https://github.com/RedHatInsights/insights-inventory-frontend/pull/2141

src/components/GeneralInfo/SystemCard/SystemCard.test.js
```diff
@@ -74,263 +69,300 @@ describe('SystemCard', () => {
...
       );
+
       expect(
-        wrapper.find('SystemCardCore').first().instance().getAnsibleHost()
-      ).toBe('test-id');
+        screen.getByRole('definition', {
+          name: /ansible hostname value/i,
+        })
+      ).toHaveTextContent('test-id');
     });
```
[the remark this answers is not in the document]
gkarat: lol, these both tests make sense I guess. Only the description is bad 😄 the first one checks if data is displayed when provided thru `fqdn`, the second thru `id`. I will tweak the test descriptions

PR #2034, opened 2023-09-27
fix(RHIN-447): Update Staleness form validation

adonispuente:
The 'Never' string wasnt converted in the form validations, so error messages/validation wasnt working as it should.
This PR updates var names and adds that conversion + styling on error messages for the first dropdowns.

To test navigate to the staleness and cullings page 
play around with the dropdowns.
The dropdown to the left should never be allowed to go higher than the dropdown to its right.
Staleness < Stale warning < Culling


bastilian (approved):
Works for me! Thank you @adonispuente!

PR #2037, opened 2023-10-02
feat(RHIF-279: configurable, shared ImmutableDevices table

mkholjuraev:
This is the new table component that is shared through the fed-mod. It can be further extended and used to show an Immutable device table.

### DIFF
--- a/src/components/ImmutableDevices/helpers.js
+++ b/src/components/ImmutableDevices/helpers.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import {
   CheckCircleIcon,
   CubeIcon,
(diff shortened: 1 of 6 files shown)


src/components/ImmutableDevices/helpers.js
```diff
@@ -0,0 +1,57 @@
...
+) =>
+  dispatcherStatus === 'ERROR'
+    ? 'error'
+    : dispatcherStatus === 'UNRESPONSIVE'
+    ? 'unresponsive'
+    : deviceStatus === 'UPDATING'
+    ? 'updating'
+    : isUpdateAvailable
+    ? 'updateAvailable'
+    : 'upToDate';
```
bastilian, line 57: There is no way this is humanly readable. It might be better to use proper `if()...else` here, just to allow for better reading what is going on.

Maybe there is even a better way to match these device states?

mkholjuraev: Thank you for the proactive review. This is yet in Draft state. PTAL on this ready [PR](https://github.com/RedHatInsights/insights-inventory-frontend/pull/2030#event-10524769575) if you have free capacity.

mkholjuraev:
@RedHatInsights/team-interact this PR is now ready for a review!

bastilian (approved):
LGTM! Thank you @mkholjuraev!

PR #2038, opened 2023-10-03
fix(ESSNTL-5468): Check for ungrouped hosts in DetailRenderer

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5468, https://issues.redhat.com/browse/ESSNTL-5469, https://issues.redhat.com/browse/ESSNTL-5470.

The DetailRenderer component shows a header with system basic information. The component is shared across more apps (see the attached tickets), and there has been an issue validating permissions for ungrouped systems. For ungrouped hosts, the component was wrongly checking for general hosts:write permissions instead of using `null` in resource definition as a minimum required permission.

Now, when a host is not a part of any group, the header renders correct information if the user has enough permissions.

## How to test

The easiest way to test is to merge this PR and see it deployed in stage-preview :D Or to run some of the apps locally together with Inventory thru `LOCAL_API` settings.

### DIFF
--- a/src/components/InventoryDetail/DetailRenderer.js
+++ b/src/components/InventoryDetail/DetailRenderer.js
@@ -4,19 +4,17 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import AccessDenied from '../../Utilities/AccessDenied';
 import { hosts } from '../../api';
-import {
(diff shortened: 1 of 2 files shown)


AsToNlele (approved):
This fixes it, good job! :smile:  Tested on Advisor and Compliance

PR #2041, opened 2023-10-05
fix(RHINENG-1490): Add update RBAC check for staleness & typo/styling fixes

adonispuente:
https://issues.redhat.com/browse/RHINENG-1490 implements the new RBAC rules for this page.

### DIFF
--- a/src/components/InventoryHostStaleness/InventoryHostStaleness.js
+++ b/src/components/InventoryHostStaleness/InventoryHostStaleness.js
@@ -2,9 +2,21 @@
 import React from 'react';
 import HostStalenessCard from './HostStalenessCard';
 import { usePermissionsWithContext } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
-import { GENERAL_HOST_STALENESS_WRITE_PERMISSION } from './constants';
+import {
+  GENERAL_HOST_STALENESS_READ_PERMISSION,
+  GENERAL_HOST_STALENESS_WRITE_PERMISSION,
+} from './constants';
+import {
+  GENERAL_HOSTS_READ_PERMISSIONS,
+  GENERAL_HOSTS_WRITE_PERMISSIONS,
+} from '../../constants';
 
-const REQUIRED_PERMISSIONS = [GENERAL_HOST_STALENESS_WRITE_PERMISSION];
+const REQUIRED_PERMISSIONS = [
+  GENERAL_HOST_STALENESS_WRITE_PERMISSION,
+  GENERAL_HOST_STALENESS_READ_PERMISSION,
+  GENERAL_HOSTS_READ_PERMISSIONS,
+  GENERAL_HOSTS_WRITE_PERMISSIONS,
+];
 
 const InventoryHostStaleness = () => {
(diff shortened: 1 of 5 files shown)


adonispuente:
The test that are failing or not related to this PR as confirmed in the interact channel

mkholjuraev (commented):
Codewise looks good to me. I am not able to find an account without these perms. Nor I can not edit my roles. I will test this PR if you provide instructions.

src/routes/InventoryHostStaleness.js
```diff
@@ -9,11 +9,11 @@ import { usePermissionsWithContext } from '@redhat-cloud-services/frontend-compo
...
-import { GENERAL_GROUPS_READ_PERMISSION } from '../constants';
+import { GENERAL_HOSTS_READ_PERMISSIONS } from '../constants';
 
 const REQUIRED_PERMISSIONS = [
   GENERAL_HOST_STALENESS_READ_PERMISSION,
-  GENERAL_GROUPS_READ_PERMISSION,
+  GENERAL_HOSTS_READ_PERMISSIONS,
 ];
 
 const HostStaleness = () => {
```
gkarat, line 19: For the permissions check, you should use `true` for checkAll parameter (the second one); this is due to the fact that both permissions should be fulfilled:

```
  const { hasAccess: canReadHostStaleness } = usePermissionsWithContext(
    REQUIRED_PERMISSIONS,
    true,
    false
  );
```

gkarat: Another thing: I also think we have to change the `checkResourceDefinitions` (the third parameter) to true or just remove it (because it set by default to true): the reason is that according to miro graphs, the whole card should be visible only if users have inventory:hosts:read permission to all hosts. With ignoring RD, you are basically saying "it's sufficient if user has any hosts:read permission even when they are limited with RD, e.g. to some group).

gkarat (changes requested):

src/components/InventoryHostStaleness/InventoryHostStaleness.js
```diff
@@ -2,9 +2,21 @@
...
-const REQUIRED_PERMISSIONS = [GENERAL_HOST_STALENESS_WRITE_PERMISSION];
+const REQUIRED_PERMISSIONS = [
+  GENERAL_HOST_STALENESS_WRITE_PERMISSION,
+  GENERAL_HOST_STALENESS_READ_PERMISSION,
+  GENERAL_HOSTS_READ_PERMISSIONS,
+  GENERAL_HOSTS_WRITE_PERMISSIONS,
+];
 
 const InventoryHostStaleness = () => {
   const { hasAccess: canModifyHostStaleness } =
```
gkarat, line 22: You have to set `checkAll` (the second parameter): so that you make the logical AND operation and check that all of the permissions are fulfilled at a time

src/routes/InventoryHostStaleness.js
adonispuente: good catch, than you!

gkarat (approved):
Looks great @adonispuente . I have tested it locally, and the view is up to date with the mocks for different permissions. let's 🚢 it

PR #2042, opened 2023-10-06
fix(ESSNTL-4731): Detect 404 from Hosts

AsToNlele:
[ESSNTL-4731](https://issues.redhat.com/browse/ESSNTL-4731)

When an user visited a non existing page (eg. had 20 systems and displayed 20 per page, the second page would throw a 404) he would be faced with an error page.

Now the user is redirected to the first page and gets an error notification.

Had to use `useSearchParams` because there's an edge case when tags finish fetching first and page is set to 1 ignoring the url params.

How to test:
1. Go to Inventory Systems and change the page to a page number that doesn't exist
2. For some reason it could take several tries to get the error message. (Could be a router upgrade regression?)
3. After the error message you should be redirected to the first page and the table should correctly load.


bastilian:
@AsToNlele Mind rebasing this so it's just your changes in here?

gkarat:
@AsToNlele, the commit linter is failing (you'd need to update the commit messages according to conventions in order to fix that)

src/components/InventoryTable/InventoryTable.js
```diff
@@ -146,6 +149,18 @@ const InventoryTable = forwardRef(
         abortOnUnmount && controller.current.abort();
       };
     }, []);
+    const hasLoadEntitiesError =
+      error?.status === 404 &&
+      error?.type === 'LOAD_ENTITIES' &&
```
gkarat, line 154: I think you can import a constant with the LOAD_ENTITIES value from src/store/action-types.js

gkarat (approved):
LGTM! Thank you @AsToNlele. Tried it locally, and the table navigates to the first page. The applied filters are also preserved which is great 👍🏼

src/components/InventoryTable/InventoryTable.js
AsToNlele: Good idea, thank you!

PR #2043, opened 2023-10-06
chore(commitlint): Enable commitlint on all branches

gkarat:
Enable commitlint in all branches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ if: tag IS blank
 jobs:
   include:
     - stage: Lint
-      script: npm run lint && if [[ $TRAVIS_PULL_REQUEST == false && $TRAVIS_BRANCH =~ $(echo "^(master-stable|master|prod-beta|prod-stable)$") ]]; then commitlint-travis; fi
+      script: npm run lint && commitlint-travis
     - stage: Test
       script: npm run test && npm run test:ct
       after_success: npm run coverage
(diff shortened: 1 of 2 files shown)


bastilian:
I have seen in other repos that running commitlint on master can sometimes be iffy, when a PR gets squashed and merged and the PR title then commit message does not conform.
It requires then to amend the last commit and force push a new master.

I'm honestly not sure if that is good or bad though...

mkholjuraev:
> Let's see what happens, maybe it's good to run it on master. Thank you @gkarat!

Hmm, we wanted not to have to commit lint on the master branch, rather we [wanted](https://redhat-internal.slack.com/archives/C02T90PLJ5V/p1696599287638499) it to happen in the PR phase. @gkarat does it PR implement that thing?

bastilian:
@mkholjuraev This PR will enable running commitlint on the master branch, and it is not a bad thing. The only annoying thing will be if we aren't carful when writing the commit message when we squash and merge.

mkholjuraev:
@bastilian yup, we have already seen multiple times those annoying situations as you described. Devs that are not aware of the commit lint tend to merge commits that do not conform to the lint and they change locally, rebase, change commit history, and then force push. Later on, when we fully migrate to containers, if the commit history is changed it will lead to confusion as we faced several weeks ago with the insights dashboard app.

bastilian:
@mkholjuraev There is no way we can verify commit messages in master, without running them on master. And to keep the history in master proper, we will need to rewrite offending commit messages.

PR #2044, opened 2023-10-06
fix(THEEDGE-3610): add validation of host type before call devices

acosferreira:
Add validation for host before call edge devices view and avoid 404
Edge device
image

Traditional device
image


src/routes/InventoryDetail.js
```diff
@@ -134,11 +134,13 @@ const Inventory = () => {
       entity &&
       appList[osSlug]?.map((app) => {
         app.isDisabled = app.nonEdge && hostType === 'edge' ? true : false;
+        setHType(hostType);
```
mkholjuraev, line 137: I am wondering if there is any reason to set a state in a loop. If there is not, let's take this out.

mkholjuraev (changes requested):

mkholjuraev (approved):
LGTM! @acosferreira thank you! Let's first fix failing tests before merging

mkholjuraev:
@acosferreira Can you please rebase your PR with master locally and remove that commit by me :)

PR #2051, opened 2023-10-10
feat(RHIF-276): hide update method filter when edge parity is enabled

mkholjuraev:
Resolves: https://issues.redhat.com/browse/RHIF-276

Hides the system update method filter when edge parity feature is enabled.

To test:
1. visit inventory page
2. Verify that edge parity feature is disabled in stage env
3. Open filters dropdown
4. Observe that the filter is hidden
5. Enable the filter in stage env and verify its not hidden


mkholjuraev:
Putting this in Draft till I get a clear direction from PM.

PR #2054, opened 2023-10-11
fix(RHINENG-462-2): hide centos versions in os filter

johnsonm325:
Only inventory and tasks need to see centos versions in the operating system filter on the inventory table. All other apps will not be fetching centos systems, so they don't need them in the dropdown.

You can test this by running inventory with tasks and any other apps to make sure they show or are hidden.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -368,7 +368,7 @@ export const getOperatingSystems = async (params = [], showCentosVersions) => {
     ...params
   );
   if (!showCentosVersions) {
-    let newResults = operatingSystems?.results.filter(
+    const newResults = operatingSystems.results.filter(
       ({ value }) => !value.name.toLowerCase().startsWith('centos')
     );
     operatingSystems.results = newResults;
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -363,9 +363,18 @@ export function getAllTags(search, pagination = {}) {
   );
 }
 
-export function getOperatingSystems(params = []) {
-  return systemProfile.apiSystemProfileGetOperatingSystem(...params);
-}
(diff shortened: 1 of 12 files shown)


bastilian:
It looks and works as expected for me, but I think we'll need to take a closer look at the tests. I restarted Travis CI a second time now and it looks like there is maybe a race condition somewhere.

src/api/api.js
```diff
@@ -363,9 +363,18 @@ export function getAllTags(search, pagination = {}) {
...
 
-export function getOperatingSystems(params = []) {
-  return systemProfile.apiSystemProfileGetOperatingSystem(...params);
-}
+export const getOperatingSystems = async (params = [], showCentosVersions) => {
+  let operatingSystems = await systemProfile.apiSystemProfileGetOperatingSystem(
+    ...params
+  );
+  if (!showCentosVersions) {
+    let newResults = operatingSystems?.results.filter(
```
gkarat, line 371: This would error if `operatingSystems?.results` evaluates to undefined. `results` must be always present (this is the part of all of the API standardised responses).

gkarat, line 371:
```suggestion
    const newResults = operatingSystems?.results.filter(
```

gkarat:
https://github.com/RedHatInsights/insights-inventory-frontend/pull/2054/commits/ee127c5e70bde5b47b5e1734a81c3c57184e73e4 adds centos also to a couple of views in Inventory.

https://github.com/RedHatInsights/insights-inventory-frontend/pull/2054/commits/58537c31843608974b5d4a70b15e7b217828b262 should fix the tests

bastilian (approved):
LGTM! Thank you @johnsonm325!

PR #2057, opened 2023-10-12
feat(THEEDGE-3585): Allow immutable systems selection in group details empty state

ldjebran:
In the context of edgeParity, when creating a new group and selecting that group we enter the group details view, because it has no systems added yet an empty state is displayed and Add systems button is displayed also to systems, this PR allow to add conventional and immutable systems in that view.

- This should not affect the main add systems button when systems exists in group that's why edgeParityIsAllowed key is added.
- Two feature flags has be taken into considerations 'edgeParity.inventory-list' and 'edgeParity.inventory-groups-enabled'.
- If there is no immutable systems the immutable tab is not shown.
- Label is added in the header when edgeParity is enabled to show the overall number of the selected systems.
- The constraint is also taken into account also, when the immutable system has a group to disable adding system and show the warning message.

Note: The group details view has no immutable tab for the moment that will be added in another task.

FIXES: https://issues.redhat.com/browse/THEEDGE-3585

inventory-group-details-add-systems-empty-state-2023-10-12 15-18

### DIFF
--- a/src/components/InventoryGroupDetail/NoSystemsEmptyState.js
+++ b/src/components/InventoryGroupDetail/NoSystemsEmptyState.js
@@ -30,6 +30,7 @@ const NoSystemsEmptyState = ({ groupId, groupName }) => {
         setIsModalOpen={setIsModalOpen}
         groupId={groupId}
         groupName={groupName}
+        edgeParityIsAllowed={true}
       />
       <EmptyStateIcon
         icon={PlusCircleIcon}
--- a/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
+++ b/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
@@ -1,7 +1,17 @@
-import { Alert, Button, Flex, FlexItem, Modal } from '@patternfly/react-core';
+import {
+  Alert,
+  Button,
+  Flex,
+  FlexItem,
+  Label,
+  Modal,
+  Tab,
+  TabTitleText,
+  Tabs,
+} from '@patternfly/react-core';
 import { TableVariant } from '@patternfly/react-table';
 import PropTypes from 'prop-types';
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import {
   clearFilters,
@@ -16,12 +26,18 @@ import ConfirmSystemsAddModal from './ConfirmSystemsAddModal';
 import { useBulkSelectConfig } from '../../../Utilities/hooks/useBulkSelectConfig';
 import difference from 'lodash/difference';
 import map from 'lodash/map';
+import ImmutableDevicesView from '../../InventoryTabs/ImmutableDevices/EdgeDevicesView';
+import useFeatureFlag from '../../../Utilities/useFeatureFlag';
+import { PageHeaderTitle } from '@redhat-cloud-services/frontend-components/PageHeader';
+import { InfoCircleIcon } from '@patternfly/react-icons';
+import { inventoryHasEdgeSystems } from '../../../Utilities/edge';
 
 const AddSystemsToGroupModal = ({
   isModalOpen,
   setIsModalOpen,
   groupId,
   groupName,
+  edgeParityIsAllowed,
 }) => {
   const dispatch = useDispatch();
 
@@ -32,7 +48,6 @@ const AddSystemsToGroupModal = ({
   );
   const rows = useSelector(({ entities }) => entities?.rows || []);
 
-  const noneSelected = selected.size === 0;
   const total = useSelector(({ entities }) => entities?.total);
   const displayedIds = map(rows, 'id');
   const pageSelected =
@@ -55,7 +70,6 @@ const AddSystemsToGroupModal = ({
       );
     }
   );
-  const showWarning = alreadyHasGroup.length > 0;
 
   const handleSystemAddition = useCallback(
     (hostIds) => {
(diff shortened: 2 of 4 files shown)


src/components/InventoryGroupDetail/NoSystemsEmptyState.js
```diff
@@ -30,6 +30,7 @@ const NoSystemsEmptyState = ({ groupId, groupName }) => {
         setIsModalOpen={setIsModalOpen}
         groupId={groupId}
         groupName={groupName}
+        edgeParityIsAllowed={true}
```
mkholjuraev, line 33: Maybe it is better to provide a feature flag here? It will help us releasing the app without hesitations

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -92,14 +111,75 @@ const AddSystemsToGroupModal = ({
...
+  // overallSelectedKeys is the list of the conventional and immutable systems ids
+  const overallSelectedKeys = [...selected.keys(), ...selectedImmutableKeys];
+  // noneSelected a boolean showing that no system is selected
+  const noneSelected = overallSelectedKeys.length === 0;
+
+  const immutableDevicesAlreadyHasGroup = selectedImmutableDevices.filter(
+    (immutableDevice) => immutableDevice.deviceGroups?.length > 0
+  );
+  // show warning also when immutable systems had groups
+  showWarning = showWarning || immutableDevicesAlreadyHasGroup.length > 0;
```
bastilian, line 133: It might be good to wrap this into a `useMemo` and also avoid reassigning `showWarning` from line 77. Things like these have caused unexpected behaviour with React.

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -92,14 +111,75 @@ const AddSystemsToGroupModal = ({
...
+
+  const [hasImmutableSystems, setHasImmutableSystems] = useState(false);
+
+  useEffect(() => {
+    if (edgeParityEnabled) {
+      (async () => {
+        const result = await axios.get(
```
mkholjuraev, line 140: We should definitely keep this data somewhere on a higher level and share it across components where needed. Otherwise, this same request to check if there is a system is made in multiple places on a single page which is not optimal.

src/components/InventoryTabs/ImmutableDevices/EdgeDevicesView.js
```diff
@@ -0,0 +1,30 @@
...
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useDispatch } from 'react-redux';
+
+const ImmutableDevicesView = (props) => {
+  const dispatch = useDispatch();
+  const notificationProp = getNotificationProp(dispatch);
+  return (
+    <AsyncComponent
+      appName="edge"
+      module="./DevicesView"
```
mkholjuraev, line 18: For my knowledge what is the difference between the this fed module and this [one](https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/InventoryTabs/ImmutableDevices/EdgeDevicesTab.js)

src/components/InventoryGroupDetail/NoSystemsEmptyState.js
ldjebran: Feature flag option is not realisable here , as the AddSystemsToGroupModal component is used in two different places and we want to allow edge parity only when AddSystemsToGroupModal component is used in group details empty state

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -156,17 +253,34 @@ const AddSystemsToGroupModal = ({
...
-            variant={TableVariant.compact} // TODO: this doesn't affect the table variant
-            tableProps={{
-              isStickyHeader: false,
-              canSelectAll: false,
-            }}
-            bulkSelect={bulkSelectConfig}
-            initialLoading={true}
-            showTags
-          />
+          {edgeParityEnabled && hasImmutableSystems ? (
```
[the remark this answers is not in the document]
ldjebran: we cannot use that component as we want here basic inventory and devices view tables functionality , and HybridInventoryTabs is implementing too much to be reduced here.

src/components/InventoryTabs/ImmutableDevices/EdgeDevicesView.js
ldjebran: DevicesView is the basic component of Inventory without all actions features also inventory does not allow to select items , as it needs that for it's own functionality , we will need to deeply modify Inventory component to allow that , but DevicesView that is a basic component is enough for the current  implementation, that is also used for groups systems selection in edge-frontend

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
ldjebran: Thanks for pointing this,  will assign it only once then 
```js
const showWarning = alreadyHasGroup.length > 0 || immutableDevicesAlreadyHasGroup.length > 0;
```

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
ldjebran: What you mean "keep this data" ?

ldjebran: created a function inventoryHasEdgeSystems at src/Utilities/edge.js

ldjebran:
@mkholjuraev @bastilian updated and responded to comments, please revisit.

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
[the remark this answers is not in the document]
ldjebran: + HybridInventoryTabs is intended to be used as main page and using useLocation and navigate , the current component is used in a modal component.

ldjebran:
resolved conflicts

mkholjuraev (approved):
LGTM! Did not face issues while testing

ldjebran:
rebased

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -92,14 +106,75 @@ const AddSystemsToGroupModal = ({
     dispatch(clearFilters());
   };
 
+  const edgeParityEnabled =
```
bastilian, line 109: This might raise an error, because hooks are called conditionally. 
It should be more something like this:

```js

const edgeInventoryListEnabled = useFeatureFlag('edgeParity.inventory-list');
const edgeInventoryGroups = useFeatureFlag('edgeParity.inventory-groups-enabled');
const edgeParityEnabled = edgeParityIsAllowed && edgeInventoryListEnabled && edgeInventoryGroups;

```

ldjebran: Thanks @bastilian for pointing this, fixed

ldjebran:
Seems the automation was broken, as was not able to find npm.  
Updated commit message, to force automation to restart, as see a newer PR succeeded.

PR #2058, opened 2023-10-12
fix(THEEDGE-3610): add validation of host type before call devices

acosferreira:
Add validation for host before call edge devices view and avoid 404
Edge device
image

Traditional device
image


mkholjuraev (approved):
LGTM! For conventional systems, I do not see any req to edge API.

PR #2065, opened 2023-10-19
fix(RHIF-294): disable sorting on immutable status column

mkholjuraev:
Disables sorting on Status column in immutable devices table


Fewwy (approved):
LGTM

PR #2075, opened 2023-10-25
chore(ImmutableDevices): cover with cypress tests

mkholjuraev:
Covers new fed-mod Immutable devices with cypress tests

### DIFF
--- a/src/components/ImmutableDevices/ImmutableDevices.cy.js
+++ b/src/components/ImmutableDevices/ImmutableDevices.cy.js
@@ -1,0 +1,182 @@
+import React from 'react';
+import { Route, Routes } from 'react-router-dom';
+import ImmutableDevices from './ImmutableDevices';
+import {
+  featureFlagsInterceptors,
+  hostsInterceptors,
+} from '../../../cypress/support/interceptors';
+
+const defaultProps = {
+  mergeAppColumns: (columns) => columns,
+};
+
+// eslint-disable-next-line react/prop-types
+const MockRouter = ({ path = '/insights/inventory', ...props }) => (
+  <Routes>
+    <Route path={path} element={<ImmutableDevices {...props} />} />
+    <Route path={'*'} element={<div id="mock-detail-page" />} />
+  </Routes>
+);
+
+const getEntities =
+  (dataInject) => async (items, config, showTags, defaultGetEntities) => {
+    const result = await defaultGetEntities(items, config, showTags);
+    result.results.map(dataInject);
+    return result;
+  };
+
+const mountWithProps = (props) => {
+  return cy.mountWithContext(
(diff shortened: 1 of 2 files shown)


src/components/ImmutableDevices/ImmutableDevices.cy.js
```diff
@@ -0,0 +1,182 @@
...
+  });
+
+  it('renders without issues', () => {
+    mountWithProps(defaultProps);
+
+    cy.get('table[aria-label="Host inventory"]').should(
+      'have.attr',
+      'data-ouia-safe',
+      'true'
+    );
```
gkarat, line 56: I am not sure asserting data-ouia-safe is enough to understand whether the table renders without problems. The thing is, the parameter is never updated in fact: https://www.patternfly.org/2022.05/components/table/react-legacy#table - it has default value set to true for ouiaSafe. And we also don't provide this parameter explicitly to PfTable (see EntityTable.js) which makes data-ouia-safe basically useless at the moment.

src/components/ImmutableDevices/ImmutableDevices.cy.js
```diff
@@ -0,0 +1,182 @@
...
+
+const defaultProps = {
+  mergeAppColumns: (columns) => columns,
+};
+
+// eslint-disable-next-line react/prop-types
+const MockRouter = ({ path = '/insights/inventory', ...props }) => (
+  <Routes>
+    <Route path={path} element={<ImmutableDevices {...props} />} />
+    <Route path={'*'} element={<div id="mock-detail-page" />} />
```
gkarat, line 17: Do we really need the second route? I see it is used in one of the tests to check that browser navigated to a new URL. But can we just assert on `window.location.pathname`? This would help us remove extra code and just stick to `mountWithContext` function.

src/components/ImmutableDevices/ImmutableDevices.cy.js
gkarat, line 1: In my honest opinion, I think all of these scenarios can be moved to unit tests suite and tested with RTL. Can we do that?

gkarat (changes requested):

src/components/ImmutableDevices/ImmutableDevices.cy.js
mkholjuraev: Aha, will choose another selector.

src/components/ImmutableDevices/ImmutableDevices.cy.js
mkholjuraev: Wondering if having this extra code deserves to be kept. Because, instead of relying on bare window location, I will be sure that the next tab is open. What do you think?

PR #2078, opened 2023-11-01
fix(RHIF-72): Fix InventoryTable font size and actions button padding

gkarat:
Fixes https://issues.redhat.com/browse/RHIF-72.

- Changes the InventoryTable rows text to have font size 14 px.
- Fixes the actions (kebab) button styling so that it is equally distanced from both sides.

Before

Screenshot 2023-11-03 at 17 47 54

After

Screenshot 2023-11-03 at 17 45 56

### DIFF
--- a/src/components/InventoryTable/InventoryList.scss
+++ b/src/components/InventoryTable/InventoryList.scss
@@ -90,10 +90,6 @@
             font-size: var(--pf-global--FontSize--sm);
         }
 
-        * {
-            font-size: var(	--pf-global--FontSize--lg);
-        }
-
         .ins-composed-col {
             cursor: pointer;
         }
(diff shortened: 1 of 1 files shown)


AsToNlele (changes requested):
Thank you for the fix @gkarat! :smile:  The font size is finally consistent between the apps

Just a small nitpick, I believe the kebab on the right side should have 24px padding from the right side, not 16px. Feel free to correct me

gkarat:
You are right, the spacing for the default variant table is 24 px: https://v4-archive.patternfly.org/v4/components/table/design-guidelines#default-spacing. Will fix it now

gkarat:
@AsToNlele, it's fixed now

AsToNlele (approved):
Perfect, thank you @gkarat! :smile:

PR #2080, opened 2023-11-02
chore(ImmutableDevices): allow passing onRowClick function

mkholjuraev:
Enables consumer apps to pass their own onRowClick function. This will enable to have different behavior of system name column depending on the consumer.

### DIFF
--- a/src/components/ImmutableDevices/ImmutableDevices.js
+++ b/src/components/ImmutableDevices/ImmutableDevices.js
@@ -16,6 +16,7 @@ const ImmutableDevices = ({
   filterConfig,
   hideFilters,
   activeFiltersConfig,
+  onRowClick,
 }) => {
   const dispatch = useDispatch();
   const navigate = useNavigate();
@@ -40,8 +41,8 @@ const ImmutableDevices = ({
     return [...mergeAppColumns(filteredColumns), ...edgeColumns];
   };
 
-  const onRowClick = (_key, systemId) => {
-    navigate(`/insights/inventory/${systemId}?appName=vulnerabilities`);
+  const defaultOnRowClick = (_key, systemId) => {
+    navigate(`/insights/inventory/${systemId}`);
   };
 
   return (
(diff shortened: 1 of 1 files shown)


bastilian:
Wouldn't the `onRowClick` of the TitleColumn[0] (vie the entity) already allow to do what this is meant to allow? 

[0] https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/InventoryTable/TitleColumn.js#L12

bastilian:
> Do we want to tweak TitleColumn?

We always want to enhance what we have instead of introducing something new. We have had several occasions in the past where we simply wanted to change where the link navigates to. It seems to me it is time to enhance the `onRowClick` of the TitleColum.

mkholjuraev:
> It seems to me it is time to enhance the `onRowClick` of the TitleColum.

I am happy to improve. I am thinking about what would be the best way to go as TitleColumn is used in almost all places where InventoryTable is consumed. It is provided to default columns. Would I need to pass a new `onTitleColumnClick` function from top to down the TitleColumn component?

mkholjuraev:
@bastilian I investigated the scope of changes needed to improve the TitleColumn. It feels to me that it has large-scale changes and testing are needed. I would not want to be blocked with this while there are more important tasks waiting for edge-parity work. I can create a task for this improvement.

bastilian:
@mkholjuraev What changes would be "large-scale"? It would likely just need the addition of a `onRowClick` property to the entities provided to the InventoryTable.

mkholjuraev:
I am not able to find a clean way to pass onRowClick to TitleColumn. Can you clarify your idea? By large-scale, I mean passing this onRowClick would require me to make changes in multiple functions and test the change in multiple applications to make sure it does not affect them.

bastilian:
@mkholjuraev Is there no `getEntities` or fetch function that can add a `onRowClick` prop to each entity it returns?

mkholjuraev:
Putting in draft mode for  https://github.com/RedHatInsights/insights-inventory-frontend/pull/2087

PR #2083, opened 2023-11-03
chore(HybridInventoryTab): accept host counts instead of API request

mkholjuraev:
Improves the API requests scattered around components to check the presence of conventional and edge systems by sharing one request result through the react context provider.

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,4 +1,11 @@
-import React, { Suspense, lazy, useEffect, useMemo, useState } from 'react';
+import React, {
+  Suspense,
+  createContext,
+  lazy,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { Navigate, useRoutes } from 'react-router-dom';
 import { getSearchParams } from './constants';
 import RenderWrapper from './Utilities/Wrapper';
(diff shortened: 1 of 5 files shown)


bastilian:
This requires each applications to do the API requests themselves, isn't that actually scattering the API requests more?

mkholjuraev:
> This requires each applications to do the API requests themselves, isn't that actually scattering the API requests more?

Thank you for having a look. It seems I wrote a confusing description. By improving the scattered requests, I meant the API requests only inside/for the inventory app. As you can see now, we fetch the requests to check the presence of immutable and conventional systems in the first place where it is needed (Routes component) and save the info inside the react context. We then use the context in multiple other places where the info is needed. For example AddSystemsToGroupModal and InventoryTable, I expect such places will be more in the future. 

Regarding passing the new props to HybridTabsComponent, as it is used by other apps and they have their own conventional system from their API, they need to be able to pass their info, thus I have made a change to make it possible. Inventory API would give incorrect info if it is used. Let me know if this makes sense, I am ready for more discussion.

AsToNlele:
Each of the apps with empty state is doing this check already, maybe this could be somehow solved on chrome level?
IIRC this PR checks for the conventional/edge straight in the `Routes.js` instead of in the Conventional table when you open it. Probably a single shared hook would be the next step in the future? :smile:

AsToNlele (approved):
LGTM, works as described @mkholjuraev :smile:

mkholjuraev:
> Each of the apps with empty state is doing this check already, maybe this could be somehow solved on chrome level? IIRC this PR checks for the conventional/edge straight in the `Routes.js` instead of in the Conventional table when you open it. Probably a single shared hook would be the next step in the future? 😄

Indeed, I have the same idea since we worked on zero state. However, there were more-grained query filters in each app to decide the zero state making such a utility not perfect. But, we did not implement it due to time constraints. Thus, I did not push this idea hard. Also, none complained about it so far :)

PR #2086, opened 2023-11-09
feat(inventroyGroup): THEEDGE-3693- implement edge group expostion

mgold1234:
this commit expose edge group component from edge,
and give the ability if important customer log in to the system and navigate to group screen, he will see edge group list instead of inventory

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -16,10 +16,11 @@ import AsynComponent from '@redhat-cloud-services/frontend-components/AsyncCompo
 import ErrorState from '@redhat-cloud-services/frontend-components/ErrorState';
 import { inventoryHasEdgeSystems } from './Utilities/edge';
 import { inventoryHasConventionalSystems } from './Utilities/conventional';
-
+const InventoryOrEdgeView = lazy(() =>
+  import('./routes/InventoryOrEdgeComponent')
+);
 const InventoryTable = lazy(() => import('./routes/InventoryTable'));
 const InventoryDetail = lazy(() => import('./routes/InventoryDetail'));
-const InventoryGroups = lazy(() => import('./routes/InventoryGroups'));
 const InventoryHostStaleness = lazy(() =>
   import('./routes/InventoryHostStaleness')
 );
@@ -77,7 +78,7 @@ export const Routes = () => {
     { path: '/:inventoryId/:modalId', element: <InventoryDetail /> },
     {
       path: '/groups',
-      element: groupsEnabled ? <InventoryGroups /> : <LostPage />,
+      element: groupsEnabled ? <InventoryOrEdgeView /> : <LostPage />,
     },
     {
       path: '/groups/:groupId',
--- a/src/Utilities/hooks/useEdgeGroups.js
+++ b/src/Utilities/hooks/useEdgeGroups.js
@@ -1,0 +1,34 @@
+import { useEffect, useState } from 'react';
+import useFeatureFlag from '../useFeatureFlag';
+import { fetchEdgeEnforceGroups } from '../../api';
+
+const useEdgeGroups = () => {
+  const [data, setData] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const edgeParityInventoryGroupsEnabled = useFeatureFlag(
+    'edgeParity.inventory-groups-enabled'
+  );
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setIsLoading(true);
+      try {
+        if (edgeParityInventoryGroupsEnabled) {
+          const response = await fetchEdgeEnforceGroups();
+          const enforceEdgeGroups = response?.enforce_edge_groups;
+          setData(enforceEdgeGroups);
+        }
+      } catch (error) {
+        console.error(error);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    fetchData();
+  }, []);
+  return [data, isLoading];
+};
+
+export default useEdgeGroups;
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -399,3 +399,11 @@ export const fetchEdgeSystem = () => {
     console.log(err);
   }
 };
+
+export const fetchEdgeEnforceGroups = () => {
+  try {
+    return instance.get(`${EDGE_API_BASE}/device-groups/enforce-edge-groups`);
+  } catch (err) {
+    console.error(err);
+  }
+};
--- a/src/routes/InventoryOrEdgeComponent.js
+++ b/src/routes/InventoryOrEdgeComponent.js
@@ -1,0 +1,56 @@
+import React, { useEffect, useState } from 'react';
+import EdgeGroupsView from '../components/InventoryGroups/EdgeGroups';
+import InventoryGroups from '../components/InventoryGroups/InventoryGroups';
+import { inventoryHasConventionalSystems } from '../Utilities/conventional';
+import { inventoryHasEdgeSystems } from '../Utilities/edge';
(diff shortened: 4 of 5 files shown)


ldjebran (commented):
/lgtm

src/Utilities/hooks/useEdgeGroups.js
```diff
@@ -0,0 +1,24 @@
...
+import { fetchEdgeEnforceGroups } from '../../api';
+
+const useEdgeGroups = (value) => {
+  const [data, setData] = useState(value);
+  const edgeParityInventoryGroupsEnabled = useFeatureFlag(
+    'edgeParity.inventory-groups-enabled'
+  );
+
+  useEffect(() => {
+    if (edgeParityInventoryGroupsEnabled) {
```
adonispuente, line 12: This is more of a nitpick, but theres no else chain off of this.
Instead we can do something like 

 `edgeParityInventoryGroupsEnabled &&
  (async () => {
        const response = await fetchEdgeEnforceGroups();
        const enforceEdgeGroups = response?.enforce_edge_groups;
        setData(enforceEdgeGroups);
      })();
    } 
    `
    within the useEffect.

src/Utilities/hooks/useEdgeGroups.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react';
+import useFeatureFlag from '../useFeatureFlag';
+import { fetchEdgeEnforceGroups } from '../../api';
+
+const useEdgeGroups = (value) => {
+  const [data, setData] = useState(value);
```
adonispuente, line 6: This is more of a question, but why pass in a false boolean when data can be set in here? I only see it used in one place, and we're returning the value we put inside of it if the featureflag is false.  For that I think itd be better just to set the useState variable here to false, and only change it if its true, then return data.

src/Routes.js
```diff
@@ -53,6 +55,7 @@ export const Routes = () => {
   const edgeParityInventoryListEnabled = useFeatureFlag(
     'edgeParity.inventory-list'
   );
+  const edgeGroupsUse = useEdgeGroups(false);
```
mkholjuraev, line 58: It seems the hook argument false is not needed as the hook does not accept [arguments](https://github.com/RedHatInsights/insights-inventory-frontend/pull/2086/files#diff-0f486407b67781cb20ec63cea192aea4cdb84b5bbf0dec3b84feb56ee6ba4f28R5)

mkholjuraev (approved):
LGTM! Only one nitpick and good to be merged.

adonispuente (approved):
LGTM

src/Routes.js
```diff
@@ -53,6 +55,7 @@ export const Routes = () => {
   const edgeParityInventoryListEnabled = useFeatureFlag(
     'edgeParity.inventory-list'
   );
+  const edgeGroupsUse = useEdgeGroups();
```
gkarat, line 58: I am afraid not using any isLoading flag or similar will result in a bad UX. Imagine the following scenario: user visits /groups and `InventoryGroups` component will be _always_ loaded first since the flag is false. Then, hook does the request and evaluates to true - this makes the view re-render and switch to `EdgeGroupsView` instead.

I propose to also return `isLoading` from the hook indicating whether we can finally use the flag and render the correct view.

gkarat: Another problem: this hook and also the request will be always made on any view visited in Inventory. `src/Routes.js` is not the best place to put this logic and I suggest to granulate composition and move this logic to a separate component, not to keep it in `Routes`.

gkarat (changes requested):

src/Routes.js
```diff
@@ -68,6 +71,8 @@ export const Routes = () => {
     }
   }, []);
 
+  const GroupsComponents = edgeGroupsUse ? EdgeGroupsView : InventoryGroups;
```
gkarat, line 74: The flag indicates customers that has an enforced edge groups view. Could we rename the flag to better illustrate this fact? Like `enforceEdgeGroups` instead of `edgeGroupsUse`?

src/api/api.js
```diff
@@ -399,3 +399,11 @@ export const fetchEdgeSystem = () => {
     console.log(err);
   }
 };
+
+export const fetchEdgeEnforceGroups = () => {
+  try {
+    return instance.get(`${EDGE_API_BASE}/device-groups/enforce-edge-groups`);
+  } catch (err) {
+    console.log(err);
```
gkarat, line 407: What is the reason behind not using console.error here instead? I think if the request fails, it is a clear indication of a problem. Furthermore, Sentry won't catch this error because it watches the error messages in a browser.

gkarat (changes requested):

gkarat:
Can we also use feat instead of fix prefix for the PR title? It looks like a new feature, not a bug fix @mgold1234

src/Routes.js
mgold1234: @gkarat the hook will change to true only if we log in to system with special customer , and not with new request

gkarat: @mgold1234, network request is required to be complete first before the hook will change the return value: https://github.com/RedHatInsights/insights-inventory-frontend/pull/2086/files#diff-0f486407b67781cb20ec63cea192aea4cdb84b5bbf0dec3b84feb56ee6ba4f28R16. Thus, the whole Routes component will be triggered to re-render.

mgold1234: I added the isLoading , do you think that the correct place is in new component instead of Route

src/routes/InventoryOrEdgeComponent.js
```diff
@@ -0,0 +1,52 @@
...
+      console.log(e);
+    }
+  }, []);
+
+  const ViewComponent = enforceEdgeGroups ? EdgeGroupsView : InventoryGroups;
+  if (!isLoading) {
+    return (
+      <ViewComponent
+        hasConventionalSystems={hasConventionalSystems}
+        hasEdgeDevices={hasEdgeDevices}
```
gkarat, line 39: Do we really need these props here? Neither EdgeGroupsView nor InventoryGroups component accepts these two props. I tried to trace EdgeGroupsView federated module interface and found this https://github.com/RedHatInsights/edge-frontend/blob/master/src/Routes/Groups/Groups.js#L21: it only accepts `locationProp, navigateProp` props.

src/Utilities/hooks/useEdgeGroups.js
```diff
@@ -0,0 +1,35 @@
...
+      } catch (error) {
+        console.error(error);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    fetchData();
+  }, []);
+  console.log('enforceEdgeGroups', data);
```
gkarat, line 31:
```suggestion
```

src/Routes.js
```diff
@@ -16,10 +16,10 @@ import AsynComponent from '@redhat-cloud-services/frontend-components/AsyncCompo
 import ErrorState from '@redhat-cloud-services/frontend-components/ErrorState';
 import { inventoryHasEdgeSystems } from './Utilities/edge';
 import { inventoryHasConventionalSystems } from './Utilities/conventional';
+import InventoryOrEdgeView from './routes/InventoryOrEdgeComponent';
```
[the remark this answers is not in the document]
mgold1234: sure

src/routes/InventoryOrEdgeComponent.js
mgold1234: it doesnt work without that

gkarat (approved):
LGTM! Tested locally with both enforced and standard accounts

PR #2087, opened 2023-11-10
feat(TitleColumn): RHINENG-3112 - Allow passing in a link via href prop

bastilian:
This changes the `a` tag used to a proper react router `Link` component and allows entities to have a `href` prop that the Inventory should link the `display_name` (example in compliance[0]).

[0] https://github.com/RedHatInsights/compliance-frontend/pull/2012/files#diff-09ae52efeffb58d25e728921c02f75f65b9f249855e9eea936ce9e913d6997e1R145

### DIFF
--- a/src/components/ImmutableDevices/ImmutableDevices.cy.js
+++ b/src/components/ImmutableDevices/ImmutableDevices.cy.js
@@ -174,8 +174,8 @@ describe('ImmutableDevices', () => {
 
     cy.get('td[data-label="Name"]')
       .first()
-      .find('.ins-composed-col > :nth-child(2) > a')
-      .trigger('click');
+      .find('.ins-composed-col > div > a')
+      .click();
 
     cy.get('#mock-detail-page');
   });
--- a/src/components/ImmutableDevices/ImmutableDevices.js
+++ b/src/components/ImmutableDevices/ImmutableDevices.js
@@ -4,7 +4,6 @@ import React, { useEffect } from 'react';
 import { TableVariant } from '@patternfly/react-table';
 import { InventoryTable } from '../InventoryTable';
 import useFeatureFlag from '../../Utilities/useFeatureFlag';
-import { useNavigate } from 'react-router-dom';
 import { edgeColumns } from './columns';
 
 const ImmutableDevices = ({
@@ -22,7 +21,6 @@ const ImmutableDevices = ({
   ...props
 }) => {
   const dispatch = useDispatch();
-  const navigate = useNavigate();
   const inventoryGroupsEnabled = useFeatureFlag('hbi.ui.inventory-groups');
 
   const totalItems = useSelector(({ entities }) => entities?.total);
@@ -42,10 +40,6 @@ const ImmutableDevices = ({
     );
 
     return [...mergeAppColumns(filteredColumns), ...edgeColumns];
-  };
-
-  const onRowClick = (_key, systemId) => {
-    navigate(`/insights/inventory/${systemId}?appName=vulnerabilities`);
   };
 
   return (
@@ -73,7 +67,6 @@ const ImmutableDevices = ({
       getEntities={getEntities}
       filterConfig={filterConfig}
       activeFiltersConfig={activeFiltersConfig}
-      onRowClick={onRowClick}
       showTags
       onRefresh={onRefresh}
       actionsConfig={actionsConfig}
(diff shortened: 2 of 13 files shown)


src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.js
```diff
@@ -295,7 +295,6 @@ const ConventionalSystemsTab = ({
           ],
         }}
         bulkSelect={bulkSelectConfig}
-        onRowClick={(_e, id, app) => navigate(`/${id}${app ? `/${app}` : ''}`)}
```
bastilian, line 298: I'm not sure what this was meant to link to exactly. When the `onRowClick` is called it will receive an `event`, `id`, but the third argument is not an "app" or anything, but `isMetaKey` which determines if a certain keyboard key is pressed.

bastilian: Also, by default the inventory table will link/navigate to the systems details page in the inventory, like it did here in the passed in `onRowClick`. So this became obsolete.

src/components/ImmutableDevices/ImmutableDevices.js
```diff
@@ -40,10 +38,6 @@ const ImmutableDevices = ({
     return [...mergeAppColumns(filteredColumns), ...edgeColumns];
   };
 
-  const onRowClick = (_key, systemId) => {
-    navigate(`/insights/inventory/${systemId}?appName=vulnerabilities`);
```
bastilian, line 44: This became obsolete with using the `Link` in the `TitleColumn` and the react router oddnesses being resolved, not requiring to pass in a navigate ensuring it is referencing the correct router to navigate.

src/components/InventoryTable/EntityTable.js
```diff
@@ -70,14 +67,6 @@ const EntityTable = ({
     [loaded, columns, hasItems, rows, isExpandable]
   );
 
-  const defaultRowClick = (_event, key) => {
```
bastilian, line 73: Same as above.

src/components/InventoryTable/TitleColumn.js
```diff
@@ -24,32 +24,35 @@ const onRowClick = (event, key, { loaded, onRowClick: rowClick, noDetail }) => {
...
+ * @param {React.node} children React node with information that will be shown to user as column title.
  * @param {string} id inventory UUID, used to navigate to correct URL.
  * @param {*} item row data, holds every information from redux store for currecnt row.
  * @param {*} props additional props passed from `EntityTable` - holds any props passed to inventory table.
  */
-const TitleColumn = (data, id, item, props) => (
+const TitleColumn = ({ children, id, item, ...props }) => (
   <div className="ins-composed-col sentry-mask data-hj-suppress">
-    <div key="os_release">{item?.os_release}</div>
+    {item?.os_release && <div key="os_release">{item?.os_release}</div>}
```
bastilian, line 36: I don't think that we are showing/using this anywhere, or?

src/components/ImmutableDevices/ImmutableDevices.cy.js
```diff
@@ -166,8 +166,8 @@ describe('ImmutableDevices', () => {
 
     cy.get('td[data-label="Name"]')
       .first()
-      .find('.ins-composed-col > :nth-child(2) > a')
-      .trigger('click');
+      .find('.ins-composed-col > div > a')
+      .click();
```
bastilian, line 170: The `click()` was what fixed the failing Cypress test, which is odd, because I would expect `trigger('click')` to be the same and `click` would just be a shorthand.

src/components/InventoryTable/TitleColumn.js
johnsonm325: I'm not 100% sure what it even does. Would it provide another div in the title column with the os version along with the system name? If so, then I've never seen that.

bastilian: Yeah, it would do just that. I'll leave it in for now, but I think we can eventually remove it.

src/components/InventoryTable/InventoryTable.js
```diff
@@ -343,7 +341,6 @@ InventoryTable.propTypes = {
   tableProps: PropTypes.object,
   isRbacEnabled: PropTypes.bool,
   hasCheckbox: PropTypes.bool,
-  onRowClick: PropTypes.func,
```
gkarat, line 346: nit-pick: Since the prop is sent down to EntityTable eventually, I think it makes sense to leave it specified in propTypes here as an indication that InventoryTable generally supports this property.

gkarat (approved):
LGTM! Have tested it locally for preview and stable routes and didn't find any issues while navigating. Also thanks for updating the tests

PR #2088, opened 2023-11-10
fix(ImmutableDevices): image name should take to image details page

mkholjuraev:
This fixes existing bugs when users click on the image name on Immutable devices table. Before this change, it would take users to edge device detail page, after this change users will be taken to image details page in image-builder app.

To test:
1. run inventory and advisor applications
2. Open advisor recommendations and select **Decreased performance occurs when KVM and VMWare guests are not running with recommended tuned service configuration** which has immutable devices.
3. Click on Image name 
4. Observe that you are taken to image details page in image-builder app

### DIFF
--- a/config/cypress.webpack.config.js
+++ b/config/cypress.webpack.config.js
@@ -34,6 +34,7 @@ webpackConfig.module.rules.push({
         __dirname,
         './overrideChrome.js'
       ),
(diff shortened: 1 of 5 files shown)


AsToNlele (approved):
Works great, good job @mkholjuraev :smile:

PR #2093, opened 2023-11-14
fix(RHINENG-2247): Fix delete groups modal

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-2247.

This makes the delete groups modal check for all groups before allowing users to submit the change. The modal is now able to work with any number of groups selected.

## How to test

1. Go to /inventory/groups
2. Make sure you have more than 20 groups selected
3. If any of the groups is not empty, the modal should prevent you from submitting deletion. Otherwise, the modal should prompt you to acknowledge the change and enable the submit button
5. You can also check that in the network tab, there are more than one GET /groups requests: for big number of groups, the requests are now batched into 20 groups chunks. For example, if you select 30 groups, while opening the modal, there should be two GET requests made.

### DIFF
--- a/src/Utilities/hooks/useFetchBatched.js
+++ b/src/Utilities/hooks/useFetchBatched.js
@@ -18,6 +18,20 @@ const useFetchBatched = () => {
 
       return results;
     },
+    fetchBatchedInline: (fetchFunction, list, batchSize = 20) => {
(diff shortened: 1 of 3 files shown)


gkarat:
Something is wrong with tests. Putting to draft

AsToNlele (approved):
LGTM, good job @gkarat :smile:
Like the faker implementation

PR #2094, opened 2023-11-15
chore(ImmutableDevices): add extra props to immutable table

mkholjuraev:
This adds extra props to the ImmutableDevices module to make it more usable. Cypress tests are added.

### DIFF
--- a/src/components/ImmutableDevices/ImmutableDevices.js
+++ b/src/components/ImmutableDevices/ImmutableDevices.js
@@ -16,6 +16,9 @@ const ImmutableDevices = ({
   filterConfig,
   hideFilters,
   activeFiltersConfig,
+  tableActions,
+  onRefresh,
+  actionsConfig,
 }) => {
   const dispatch = useDispatch();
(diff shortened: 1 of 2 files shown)


src/components/ImmutableDevices/ImmutableDevices.js
```diff
@@ -16,6 +16,9 @@ const ImmutableDevices = ({
   filterConfig,
   hideFilters,
   activeFiltersConfig,
+  tableActions,
+  onRefresh,
+  actionsConfig,
```
bastilian, line 21: Why not add a `...props` and use it below with the `InventoryTable`?

```js
<InventoryTable
   showTags
   ...
   { ...props }
/>
```

Later on if we see that certain attributes are used all over the place with the ImmutableDevices component we can add them here with proper default.

mkholjuraev: Done.

bastilian (approved):
Approved! Thank you @mkholjuraev!

PR #2095, opened 2023-11-15
test: Use a different way to update window object in tests

gkarat:
This is an attempt to fix the coverage random error happening on Travis CI while executing component tests.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -65,25 +65,22 @@ Cypress.Commands.add('mountWithContext', (Component, options = {}, props) => {
 Cypress.Commands.add(
   'mockWindowChrome',
   ({ userPermissions } = { userPermissions: ['*:*:*'] }) => {
-    cy.window().then(
-      // one of the fec dependencies talks to window.insights.chrome
-      (window) => {
-        window.chrome = {
-          getApp: () => 'inventory',
-        };
-        window.insights = {
-          chrome: {
-            getUserPermissions: () => userPermissions,
-            auth: {
-              getUser: () => {
-                return Promise.resolve({});
-              },
-            },
-            getApp: () => 'inventory',
+    window.chrome = {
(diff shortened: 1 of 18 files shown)


cypress/support/commands.js
```diff
@@ -65,25 +65,22 @@ Cypress.Commands.add('mountWithContext', (Component, options = {}, props) => {
...
-        window.insights = {
-          chrome: {
-            getUserPermissions: () => userPermissions,
-            auth: {
-              getUser: () => {
-                return Promise.resolve({});
-              },
-            },
-            getApp: () => 'inventory',
+    window.chrome = {
```
bastilian, line 68: Would mocking all this with the `useChrome` override not make the use of the `window` object obsolete?

gkarat: This is true, it seems this property modification is already obsolete. I will try now to remove it

gkarat: done in the next commit

gkarat: Also, I didn't remove the window.insights mock because some parts of the code (but not Inventory's) still refers to window.insights.chrome object for some reason.
```
(uncaught exception)Error: Objects are not valid as a React child (found: TypeError: Cannot read properties of undefined (reading 'chrome')). If you meant to render a collection of children, use an array instead.
```

mkholjuraev: I think this should be remnants of the chrome older version which embedded the chrome into the window heavily. Maybe if we find what packages are still using that old version and update them, we can remove this final bit as well.

gkarat: Yeah, we can. But not sure it will contribute much to the issue with the tests I am trying to solve currently. I will keep you note in mind and fix in a separate PR.

Also merging this. Looks like at least 2 people took a look and there is almost no way to verify this PR, so let's see if it heals CI on master.

PR #2096, opened 2023-11-15
fix: Don't request for unknown staleness in tags

gkarat:
The "unknown" option for host staleness is deprecated. The inventory table crashed because a request to GET /tags was resulting in 400.


adonispuente (approved):
LGTM!

PR #2097, opened 2023-11-20
Add "No groups available" message to the filter

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-1628.

This PR adds some new tests to the useGroupFilter hook and also makes the filter render "No groups available" text when there are no groups available for user.

## How to test

1. Make sure you have no groups available (created) for your account
2. Navigate to /inventory and check that the new message is displayed

## Screenshots

Screenshot 2023-11-20 at 16 56 44


AsToNlele (approved):
Good job! :smile: 
Tested with and without groups, works as expected.
Thank you for the tests

PR #2100, opened 2023-11-20
Update Patchman API to v3

leSamo:
https://issues.redhat.com/browse/RHINENG-3395

Patchman v1 and v2 APIs will be turned off on 01/01/2024 so we need to update the version. The new version is backwards compatible (no parameters are removed/changed, only added).

### DIFF
--- a/src/api/systemIssues.js
+++ b/src/api/systemIssues.js
@@ -18,7 +18,7 @@ export const cves = async (systemId) => {
 
 export const patch = async (systemId) => {
   try {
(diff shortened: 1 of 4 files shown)


AsToNlele (approved):
LGTM :smile: Thanks for doing this ahead of time

Looks like you need to change your commit message like `fix(something): message`

gkarat:
I have spun the tests step (linting is erroring because of bad commit title tho). If it passes, I will squash and merge the PR, thanks @leSamo

PR #2101, opened 2023-11-21
fix(TitleTable): RHINENG-3766: group title page in case inventory user log in

mgold1234:
Fixes https://issues.redhat.com/browse/RHINENG-3766.

this commit fix the title of group page

Screenshot 2023-11-21 at 22 30 31


gkarat (approved):
LGTM! The title is back. Please let me add one new commit to this PR: I want to test this scenario to ensure we render the right content on the /groups route

PR #2102, opened 2023-11-22
fix(THEEDGE-3703): On immutable tab we can't see the add system option

acosferreira:
Fixes https://issues.redhat.com/browse/THEEDGE-3703.

As part of this ticket, iwe moved the list of immutable from federation module to insights code, adding the update actions inline and bulk selection based on image_set values.
We also match the mockup from #3581.
Now he have add system, update bulk selection, update inline, add/remove from group and tab list.

[THEEDGE-3705.webm](https://github.com/RedHatInsights/insights-inventory-frontend/assets/5039367/b4b2c434-730d-4635-b1b9-0255b069de2e)


src/components/GroupSystems/index.js
```diff
@@ -4,20 +4,24 @@ import React from 'react';
...
     (state) => state.groupDetail
   );
-
   return uninitialized || loading ? (
     <EmptyState>
       <EmptyStateBody>
         <Spinner />
       </EmptyStateBody>
     </EmptyState>
   ) : (data?.results?.[0]?.host_count || 0) > 0 ? (
```
gkarat, line 19: Can we double-check that data?.results?.[0]?.host_count also includes the immutable systems? If not, then we have to update the condition here too.

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,47 @@ const GroupTabDetailsWrapper = ({
...
+      groupId={groupId}
+      groupName={groupName}
+      hostType={hybridInventoryTabKeys.immutable.key}
+    />
+  );
+
+  const [component, setComponent] =
+    activeTab == hybridInventoryTabKeys.conventional.key
+      ? useState(conventionalSystemsContent)
+      : useState(immutableSystemsContent);
```
gkarat, line 51: Conditional hooks call is discouraged and leads to worse performance in React: https://react.dev/warnings/invalid-hook-call-warning#breaking-rules-of-hooks. I would like to have this updated in a way that we always call hooks in the same order.

gkarat (changes requested):

src/components/GroupSystems/index.js
acosferreira: the results not includes the immutable.
We need explicit pass the filter to get the total of immutable.

src/components/InventoryGroupDetail/GroupTabDetails.js
acosferreira: thanks!I've change it

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -6,13 +6,14 @@ import {
   TabTitleText,
   Tabs,
 } from '@patternfly/react-core';
-import React, { Suspense, lazy, useState } from 'react';
+import React, { Suspense, lazy, useEffect, useState } from 'react';
 import { hybridInventoryTabKeys } from '../../Utilities/constants';
 import GroupSystems from '../GroupSystems';
```
gkarat, line 11: It feels like you are importing the wrong component: did you mean
```suggestion
import GroupSystems from '../GroupSystems/GroupSystems';
```
instead? The reason is that you later use the imported `GroupSystems` in the `conventionalSystemsContent` variable. But the current import that you have will render `GroupSystemsWrapper` (https://github.com/RedHatInsights/insights-inventory-frontend/pull/2102/files#diff-272350e154e0572ba97a0176c9495f955365b3b500f6e5c752f8c8717d0f20d0R9) component which does another check/condition. It seems to be a duplicating/double-work: if user clicks on the Conventional tab, then Conventional GroupSystems should be rendered.

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,44 @@ const GroupTabDetailsWrapper = ({
...
+    />
+  );
+
+  const immutableSystemsContent = (
+    <GroupImmutableSystems
+      groupId={groupId}
+      groupName={groupName}
+      hostType={hybridInventoryTabKeys.immutable.key}
+    />
+  );
```
gkarat, line 46: I think these components have to be memoised with useMemo and dependency array set to [groupId, groupName]. Otherwise, the contents are re-calculated with each re-render.

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,44 @@ const GroupTabDetailsWrapper = ({
...
+
+  const [component, setComponent] = useState(conventionalSystemsContent);
 
+  useEffect(() => {
+    setComponent(null);
```
gkarat, line 51: Why do we need this intermediate set to null?

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,44 @@ const GroupTabDetailsWrapper = ({
...
+    if (activeTab == hybridInventoryTabKeys.conventional.key) {
```
gkarat, line 52:
```suggestion
    if (activeTab === hybridInventoryTabKeys.conventional.key) {
```

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,44 @@ const GroupTabDetailsWrapper = ({
...
+      setComponent(conventionalSystemsContent);
+    } else {
+      setComponent(immutableSystemsContent);
+    }
+  }, [activeTab]);
   const handleTabClick = (_event, tabIndex) => {
+    setComponent(null);
     setTab(tabIndex);
+    if (tabIndex == hybridInventoryTabKeys.conventional.key) {
```
gkarat, line 61:
```suggestion
    if (tabIndex === hybridInventoryTabKeys.conventional.key) {
```

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,44 @@ const GroupTabDetailsWrapper = ({
...
+  useEffect(() => {
+    setComponent(null);
+    if (activeTab == hybridInventoryTabKeys.conventional.key) {
+      setComponent(conventionalSystemsContent);
+    } else {
+      setComponent(immutableSystemsContent);
+    }
+  }, [activeTab]);
   const handleTabClick = (_event, tabIndex) => {
+    setComponent(null);
```
gkarat, line 59: The same question here

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -24,12 +25,44 @@ const GroupTabDetailsWrapper = ({
...
     setTab(tabIndex);
+    if (tabIndex == hybridInventoryTabKeys.conventional.key) {
+      setComponent(conventionalSystemsContent);
+    } else {
+      setComponent(immutableSystemsContent);
+    }
   };
 
   const [activeTabKey, setActiveTabKey] = useState(0);
```
gkarat, line 68: It feels like we duplicating the "active tab" logic here, whereas it is already implemented in a higher level component here https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/InventoryGroupDetail/InventoryGroupDetail.js#L50. It feels redundant/a lot of duplication going here. Can we reduce the tabs handling logic to only one component? Otherwise, this gets hard to understand and especially maintain very quick.

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -0,0 +1,367 @@
...
+    config,
+    showTags,
+    defaultGetEntities
+  ) => {
+    const updateInfo = await getUpdateInfo(groupId);
+    setDeviceData(updateInfo?.update_devices_uuids);
+    setDeviceImageSet(updateInfo?.device_image_set_info);
+    const mapDeviceIds = Object.keys(updateInfo?.device_image_set_info);
+    const customResult = await fetchImagesData({ devices_uuid: mapDeviceIds });
+    //##################################//
```
gkarat, line 137: What is the purpose of these comments?

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -0,0 +1,367 @@
...
+                    Update
+                  </ActionDropdownItem>
+                ),
+                style: {
+                  padding: 0, // custom component creates extra padding space
+                },
+              },
+            ],
+          }}
+          actionsConfig={{
```
gkarat, line 303: The buttons do not have a margin between each other.

Screenshot 2023-11-23 at 12 16 38

Can you please also double-check whether the button must be hidden under three-dots kebab button? Usually, only one action is available on top of the primary toolbar and other are hidden as secondary actions.

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -0,0 +1,367 @@
...
+            actions: [
+              [
+                <ActionButton
+                  key="add-systems-button"
+                  requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
+                    groupId
+                  )}
```
gkarat, line 310: Just asking for clarification: the inventory RBAC permissions do cover the edge systems too? So by saying I have "inventory:groups:read", I am able to read both conventional and immutable systems groups?

gkarat (changes requested):

src/components/GroupSystems/index.js
acosferreira: yes, that is the expected behavior.
If there the total of `conventional is 0 and immutable > 0` we should open the screen with focus in the immutable tab and the conventional will be empty.
empty

src/components/GroupSystems/GroupImmutableSystems.js
acosferreira: by the mockup definition should be 2 and the enable/disable is handle by the device selection and we send the update to the last version

src/components/GroupSystems/GroupImmutableSystems.js
acosferreira: yes, the rbac is being handled on our backend response as well

src/components/GroupSystems/GroupImmutableSystems.js
acosferreira: fixed the layout

src/components/InventoryGroupDetail/GroupTabDetails.js
acosferreira: @gkarat this screen has 2 components with tabs [Systems | Group Info] handled by ActiveTabKey and [Conventional | Immutable] that uses ActiveTab
the context between this 2 levels are different and lives on on details and the other on the tabs

gkarat (approved):
LGTM! Thank you @acosferreira

PR #2104, opened 2023-11-22
feat(RHINENG-1850): Filtering ungrouped hosts

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-1850.

This updates the `useGroupFilter` hook to support the no group option. The /inventory and /groups/%id (in the Add systems modal) tables are updated and now allow users to filter out hosts that are not a part of any group.

⚠️ The option is disabled by default: this means it will not appear in other applications by default because non-HBI services do not yet support this feature.

## How to test

1. Go to /inventory or /groups/%id (in the Add systems modal) and check that you can filter out ungrouped hosts
2. Try to reload the page (/inventory only) and make sure that the filter is still there.

## Screenshots

image

### DIFF
--- a/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
+++ b/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
@@ -160,6 +160,7 @@ const AddSystemsToGroupModal = ({
       initialLoading={true}
       showTags
       showCentosVersions
+      showNoGroupOption
     />
   );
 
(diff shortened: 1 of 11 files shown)


mkholjuraev (commented):
Codewise LGTM, @gkarat let's resolve the conflicts and I will go through again and test functionality.

gkarat:
@mkholjuraev, conflict is resolved ✅

mkholjuraev (approved):
LGTM! 'No group' filter works great. 

I have noticed that sometimes when I first use the group filter and try to filter by the second option in the dropdown and click on group title not checkbox, filter does not get triggered. This is not related to your PR though and I am not 100% sure if it happens to me only or everyone.

gkarat:
@mkholjuraev, yeah, this issue is known. I will fix it soon

PR #2105, opened 2023-11-22
fix(ESSNTL-5418): Trim group name before validate

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-5418.

## How to test

1. Go to /groups and click create a new group
2. Try to add a name of existing group and also add some whitespaces before/after the name
3. The modal should not allow you to create a group

### DIFF
--- a/src/components/InventoryGroups/Modals/CreateGroupModal.js
+++ b/src/components/InventoryGroups/Modals/CreateGroupModal.js
@@ -7,6 +7,15 @@ import { createGroup, validateGroupName } from '../utils/api';
 import { useDispatch } from 'react-redux';
 import awesomeDebouncePromise from 'awesome-debounce-promise';
 
+export const validate = async (value) => {
+  const results = await validateGroupName(value.trim());
+  if (results === true) {
+    throw 'Group name already exists';
+  }
+
+  return undefined;
+};
+
 const CreateGroupModal = ({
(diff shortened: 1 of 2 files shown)


src/components/InventoryGroups/Modals/CreateGroupModal.js
```diff
@@ -7,6 +7,15 @@ import { createGroup, validateGroupName } from '../utils/api';
 import { useDispatch } from 'react-redux';
 import awesomeDebouncePromise from 'awesome-debounce-promise';
 
+export const validate = async (value) => {
+  const results = await validateGroupName(value.trim());
+  if (results === true) {
+    throw 'Group name already exists';
+  }
+
+  return undefined;
```
mkholjuraev, line 16: Question: why would we need to return undefined explicitly?

mkholjuraev (approved):
LGTM! I have tried creating groups that had white space before and after, also with existing name. Modal did not allow me. Nice work!

src/components/InventoryGroups/Modals/CreateGroupModal.js
gkarat: This is the requirement of data-driven-forms. In order to mark that the validator succeeds, it shouldn't return nothing or just undefined 😄

PR #2107, opened 2023-11-23
fix(EntityTable): properly remove actions resolver when there is not …

mkholjuraev:
Properly removes actionsResolver when there is table is empty and 'No systems' empty state is shown. This fixes issues that happened with the advisor application when recommendation is disabled and then re-enabled the actionResolver was deleted unwantedly. Now, with the useMemo, it gets removed only when it is needed.

To test:
1. Run the app with advisor
2. Go to advisor recommendation detail page that has immutable systems
3. Disable recommendation using the actions dropdown in the header
4. Reenable the recommendation by using the 'Enable recommendation'. You can see it in the screenshot. This banner might not show up due to race condition in the API.
5. Observer that the table action in immutable systems table is always persent
6. Search for non existing systems so that you have empty state
7. observe that there is no action kebab in the table

### DIFF
--- a/src/components/InventoryTable/EntityTable.js
+++ b/src/components/InventoryTable/EntityTable.js
@@ -93,10 +93,15 @@ const EntityTable = ({
     direction: sortBy?.direction,
   };
 
-  delete tableProps.RowWrapper;
-  if (rows?.length === 0) {
-    delete tableProps.actionResolver;
-  }
+  const modifiedTableProps = useMemo(() => {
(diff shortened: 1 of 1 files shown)


gkarat (approved):
LGTM! Hardly could reproduce the bug TBH, but the PR fixes the issue. Thank you @mkholjuraev

mkholjuraev:
@gkarat thank you for the review. This was an edge case, but you know that @computercamplove is a pro to catch such cases :)

PR #2109, opened 2023-11-28
fix(RHINENG-3059): Fix padding issues for Inventory table

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-3059.

## Before

Screenshot 2023-11-28 at 13 56 13

Screenshot 2023-11-28 at 13 55 37

## After

Screenshot 2023-11-28 at 13 56 16

Screenshot 2023-11-28 at 13 55 41


bastilian:
I think RHINENG-3059 is obsolete and would bring back: https://issues.redhat.com/browse/RHINENG-2325 :)

gkarat:
@bastilian, ah, true, so do I need to drop the commit that deals with the bulb icon space around?

PR #2110, opened 2023-11-28
fix(RHINENG-3653): Replace Group Column text field

adonispuente:
This changes the text for when a system has no group from 'N/A' to 'No group' in grey wording. 

https://issues.redhat.com/browse/RHINENG-3653

### DIFF
--- a/src/store/entities.js
+++ b/src/store/entities.js
@@ -56,7 +56,12 @@ export const defaultColumns = () => [
     title: 'Group',
     props: { width: 10 },
     // eslint-disable-next-line camelcase
-    renderFunc: (groups) => (isEmpty(groups) ? 'N/A' : groups[0].name), // currently, one group at maximum is supported
+    renderFunc: (groups) =>
+      isEmpty(groups) ? (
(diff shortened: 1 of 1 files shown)


src/store/entities.js
```diff
@@ -56,7 +56,12 @@ export const defaultColumns = () => [
     title: 'Group',
     props: { width: 10 },
     // eslint-disable-next-line camelcase
-    renderFunc: (groups) => (isEmpty(groups) ? 'N/A' : groups[0].name), // currently, one group at maximum is supported
+    renderFunc: (groups) =>
+      isEmpty(groups) ? (
+        <div className="pf-v5-u-disabled-color-100">No group </div>
```
gkarat, line 61: It looks like in the ticket they explicitly asked for the 200 variant. It also gives more contrast compared to normal black text

```suggestion
        <div className="pf-v5-u-disabled-color-200">No group</div>
```

gkarat (approved):
LGTM, thanks @adonispuente !

PR #2115, opened 2023-12-05
fix(RHINENG-715): Reroute to groups page if group is deleted

adonispuente:
https://issues.redhat.com/browse/RHINENG-715 
Currently theres a bug where if you 
1. create group
2. open this group in 2 tabs
3. delete group in first tab
4. go to second tab and refresh page
The page would still load with the id. This PR reroutes to /groups if the api request fails. most/all of the functionality will not properly work if the request  doesnt work as is so this also avoids that issue.

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -9,7 +9,7 @@ const { config: webpackConfig, plugins } = config({
     https: true,
     useProxy: true,
     proxyVerbose: true,
-    env: `${process.env.ENVIRONMENT || 'stage'}-${
+    env: `${process.env.ENVIRONMENT || 'prod'}-${
       process.env.BETA ? 'beta' : 'stable'
     }`, // for accessing prod-beta start your app with ENVIRONMENT=prod and BETA=true
     appUrl: process.env.BETA
--- a/src/components/InventoryGroupDetail/GroupDetailHeader.js
+++ b/src/components/InventoryGroupDetail/GroupDetailHeader.js
@@ -66,6 +66,7 @@ const GroupDetailHeader = ({ groupId }) => {
             await getInventoryGroupDevicesUpdateInfo(groupId);
           setEdgeDeviceUpdateInfo(groupEdgeDevicesUpdateInfo);
         } catch (error) {
+          navigate('/groups');
           console.error(error);
         }
       })();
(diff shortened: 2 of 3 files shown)


src/components/InventoryGroupDetail/GroupDetailHeader.js
```diff
@@ -66,6 +66,7 @@ const GroupDetailHeader = ({ groupId }) => {
             await getInventoryGroupDevicesUpdateInfo(groupId);
           setEdgeDeviceUpdateInfo(groupEdgeDevicesUpdateInfo);
         } catch (error) {
+          navigate('/groups');
```
gkarat, line 69: I am not sure edge devices update info is the best endpoint to check if inventory group exists or not. It is very tied to the edge devices logic, and this also wouldn't work with edge feature flag set to false. 

When a group is deleted, GET /groups/%id should return an empty response, and this is the best source of knowledge to understand if a given group still exists or not. I suggest to look into https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/InventoryGroupDetail/InventoryGroupDetail.js#L55: here, in `data`, we receive the response after fetching GET /groups/%id. From this selector, you can also pick the key "fulfilled" and other flags: if the request is fulfilled, but the response has empty list => the group doesn't exist. I think we can combine this two facts and write a new useEffect in this component.

config/dev.webpack.config.js
```diff
@@ -9,7 +9,7 @@ const { config: webpackConfig, plugins } = config({
     https: true,
     useProxy: true,
     proxyVerbose: true,
-    env: `${process.env.ENVIRONMENT || 'stage'}-${
+    env: `${process.env.ENVIRONMENT || 'prod'}-${
```
mkholjuraev, line 12: We can remove this change.

src/components/InventoryGroupDetail/InventoryGroupDetail.js
```diff
@@ -70,6 +73,8 @@ const InventoryGroupDetail = ({ groupId }) => {
   }, [canViewGroup]);
 
   useEffect(() => {
+    fulfilled && data?.total === 0 && navigate('/groups');
```
mkholjuraev, line 76: I am wondering if this can cause the page to redirect to /groups when there is a group with zero system associated with it.

mkholjuraev (commented):
I do not fully understand the how to test this. Particularly, I am not sure what is meant in the second point (**open this group in 2 tabs**) in the PR description. happy to test once I understand how to.

src/components/InventoryGroupDetail/InventoryGroupDetail.js
gkarat: The data.total here stands for the request to GET /groups. If the list is empty, then there is no such group. Redirecting to /groups is correct in that case.

gkarat:
Closing in favour of https://github.com/RedHatInsights/insights-inventory-frontend/pull/2124

PR #2116, opened 2023-12-06
fix(RHINENG-1414): Sync create group modal schema with hints 

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-1414.

This makes the create group modal behaviour consistent with schema and displayed hints. Now underscore and hyphens are allowed to be in the beginning of the group name.

## How to test

1. Go to /groups
2. Make sure you are able to create group with name starting with underscore or hyphen.

### DIFF
--- a/src/components/InventoryGroups/Modals/CreateGroupModal.test.js
+++ b/src/components/InventoryGroups/Modals/CreateGroupModal.test.js
@@ -1,7 +1,12 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
 import { validateGroupName } from '../utils/api';
-import { validate } from './CreateGroupModal';
+import CreateGroupModal, { validate } from './CreateGroupModal';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
 
 jest.mock('../utils/api');
+jest.mock('react-redux');
 
 describe('validate function', () => {
   afterEach(() => {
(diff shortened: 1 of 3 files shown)


src/components/InventoryGroups/Modals/CreateGroupModal.test.js
```diff
@@ -1,25 +1,144 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
 import { validateGroupName } from '../utils/api';
-import { validate } from './CreateGroupModal';
+import CreateGroupModal, { validate } from './CreateGroupModal';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
 
 jest.mock('../utils/api');
+jest.mock('react-redux');
```
mkholjuraev, line 9: For my understanding: Why the whole react-redux is mocked?

mkholjuraev (approved):
LGTM!

src/components/InventoryGroups/Modals/CreateGroupModal.test.js
gkarat: This helps to mock useDispatch call in the modal component. Yep, this mock is very general, it can be enhanced in the future if we would need different behaviour for different function from the library. For now, as it serves its purpose for this test, we can leave it like that :)

PR #2117, opened 2023-12-06
feat(RHINENG-4614): Enable global filter on group details page

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-4614.

This extracts global filter handling to a separate hook and integrates it with the group details (/groups/%id) and also inventory (/inventory) pages.

## How to test

1. Go to any group and try to filter out systems by workloads, tags and other global parameter. The results should be consistent (apply tags with the global filter => the table should contain only hosts with this tag).
2. Also, since the /inventory view is also refactored, make sure the global filter works there as expected.

## Screenshots

image

### DIFF
--- a/config/overrideChrome.js
+++ b/config/overrideChrome.js
@@ -3,7 +3,7 @@ const chromeMock = {
   isBeta: () => false,
   appAction: () => {},
   appObjectId: () => {},
-  on: () => {},
+  on: () => () => {},
   getApp: () => 'inventory',
   getBundle: () => 'insights',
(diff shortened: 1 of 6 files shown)


mkholjuraev (approved):
LGTM! tested locally both group details and /inventory pages. I have verified API requests are working fine, URLs are applied and then I can open the same state of the UI in another tab using that URL. 

I have not spotted any issues apart from populating workloads into the URL. When I apply more than one workload, the second and third workloads are not applied to the URL. Maybe this is expected behaviour and in case this is an issue, most probably not related to this PR.

PR #2119, opened 2023-12-10
fix(useOperatingSystemFilter): Add missing redux state handling

bastilian:


AsToNlele (commented):
When I click on a minor version, I'm able to check it an filters are updated. But I'm unable to uncheck it
image

By clicking on the major version above I'm able to uncheck the minor one

src/components/filters/useOperatingSystemFilter.js
```diff
@@ -1,118 +1,95 @@
...
+      const fullSelections = filterSelectionToValue(newSelection);
+      const fullSelection = appendGroupSelection(newSelection, groups);
+
+      console.log(
+        'setValue',
+        newSelection,
+        groups,
+        fullSelection,
+        fullSelections
       );
```
gkarat, line 51: Let's 🔥 this

src/api/api.js
```diff
@@ -105,70 +102,64 @@ export const constructTags = (tagFilters) => {
...
-    let centosVersions = [];
-    let rhelVersions = [];
-    systemProfile.operating_system = {};
-
-    osFilterValues.forEach((filterValue) => {
-      if (filterValue.osName === 'RHEL') {
-        rhelVersions.push(filterValue.value);
-      }
+  const system_profile = {
+    ...(hostTypeFilter ? { host_type: hostTypeFilter } : {}),
```
gkarat, line 144: It doesn't seem to be an equal change. Before, if hostTypeFilter was undefined, we still provided 'nil' value. Won't your change lead to any errors? I am not sure what's the right method here, but just want to point out to this difference

src/components/InventoryTabs/ConventionalSystems/Utilities.test.js
```diff
@@ -0,0 +1,14 @@
...
+describe('calculateFilters', () => {
+  it('should returns a mutated SearchParams for filters', () => {
+    const searchParams = new URLSearchParams(location.search);
+
+    expect(calculateFilters(searchParams, []).toString()).toEqual('');
+  });
+  it('should returns a mutated SearchParams for filters', () => {
+    const searchParams = new URLSearchParams(location.search);
+
+    expect(calculateFilters(searchParams, []).toString()).toEqual('');
```
gkarat, line 12: Looks duplicate to the previous test. Also, can we add some more scenarios with the second argument not being empty? Or perhaps let's leave a TODO note to work on that later if it feels overwhelming

src/Utilities/hooks/useFetchOperatingSystems.js
```diff
@@ -0,0 +1,44 @@
...
+}) => {
+  const mounted = useRef(true);
+  const initialState = {
+    operatingSystems: [],
+    operatingSystemsLoaded: false,
+  };
+  const [data, setData] = useState(initialState);
+
+  const fetchOperatingSystems = useCallback(async () => {
+    if (typeof hasAccess === 'undefined') {
```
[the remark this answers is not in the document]
bastilian: typeof returns a string. 

Screenshot 2024-04-26 at 13 21 27

gkarat (commented):
Now the requests look good, the table doesn't error, but the URL still contains `?operating_system=8&operating_system=RHEL8.6...` if you select a group of OS version. Can we also fix the URL construction part?

upd: I see you are still pushing commits so please let me know once it's ready for review 😅

bastilian:
@gkarat I was about to ping you and @AsToNlele to say that everything should work now and just the bit with the URL is left. I am looking into it, and will try to filter out that bit. Some things are still a bit hackish. I also still need to look at other apps and test this with them.

src/components/filters/helpers.test.js
```diff
@@ -13,3 +19,196 @@ describe('Validates datepicker dates', () => {
...
+    ).toEqual(11);
+
+    expect(
+      appendGroupSelection(selection, operatingSystemsGroups)['CentOS-Linux-7'][
+        'CentOS-Linux-7-7.7'
+      ]
+    ).toEqual(true);
+    console.log(
+      appendGroupSelection(selection, operatingSystemsGroups)['Alma-Linux-8']
+    );
```
gkarat, line 109:
```suggestion
```

gkarat (approved):
LGTM! Thanks @bastilian. Please, take a look at the notes below. These are not something breaking the user flows/major bug, but we either need to address it in this PR or create a separate ticket to fix this later. I have tested it only with running inventory alone, checked other views like group details, etc., haven't spotted any errors, the filtering works as expected.

- The versions are now not ordered in the descending order (9.5, 9.3, 9.2, ...). I remember this was a part of the requirements when the filter was introduced. Would be nice to have this fixed (can be this or a separate PR though).
- When you select all versions under the same major version group (like clicking on RHEL 9), then refresh the page, the "RHEL 9" (group item) is not checked though all of the descendants are selected.

image

PR #2120, opened 2023-12-11
feat(THEEDGE-3758): show both tabs when adding systems to group

ldjebran:
Show both tabs when adding systems to group, when adding conventional systems, conventional tab should be selected by default and when adding immutable systems, immutable tab should be selected by default. 

FIXES: https://issues.redhat.com/browse/THEEDGE-3758

group-details-add-systems

### DIFF
--- a/src/components/GroupSystems/GroupSystems.js
+++ b/src/components/GroupSystems/GroupSystems.js
@@ -21,6 +21,7 @@ import { useBulkSelectConfig } from '../../Utilities/hooks/useBulkSelectConfig';
 import difference from 'lodash/difference';
 import map from 'lodash/map';
 import useGlobalFilter from '../filters/useGlobalFilter';
+import { hybridInventoryTabKeys } from '../../Utilities/constants';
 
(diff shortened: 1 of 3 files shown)


src/components/GroupSystems/GroupSystems.js
```diff
@@ -126,6 +126,8 @@ const GroupSystems = ({ groupName, groupId }) => {
           }}
           groupId={groupId}
           groupName={groupName}
+          edgeParityIsAllowed={true}
+          initialActiveTab={0}
```
mkholjuraev, line 130: I think we need to give a proper tab key instead of numbers considering initialActiveTab is passed from another component and it is not easily known what tab is active exactly.

mkholjuraev (commented):
@ldjebran thank you! Codewise looks good and works as described. However, let's add some coverage tests to make sure we avoid any possible future issues before merging.

mkholjuraev (approved):
LGTM! Works as described

PR #2112, opened 2023-11-30
fix(THEEDGE-3739): Fix filter and link to device details

acosferreira:
This PR fixes the filter and link on immutable tab for group details.
https://issues.redhat.com/browse/THEEDGE-3739

THEEDGE-3739

PR #2121, opened 2023-12-11
fix(THEEDGE-3756): when removing the latest immutable systems from group

ldjebran:
deviceview with post need devices uuids list to not be empty otherwise return http status 400, check the length of devices before calling the endpoint api.

FIXES: https://issues.redhat.com/browse/THEEDGE-3756

inventory-group-remove-lastest-edge-devices


ldjebran:
seems no more relevant after this PR changes https://github.com/RedHatInsights/insights-inventory-frontend/pull/2112
closing this PR.

PR #2122, opened 2023-12-12
fix(THEEDGE-3757): add validation to imageset

acosferreira:
Fix validation to avoid exception for navigate.
Check if deviceImageSet is filled before find the imae value
THEEDGE-3757

### DIFF
--- a/src/components/GroupSystems/GroupImmutableSystems.js
+++ b/src/components/GroupSystems/GroupImmutableSystems.js
@@ -187,7 +187,7 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
 
   //enable disable bulk update based on selection, must refactor
   useEffect(() => {
-    if (selected.size > 0) {
+    if (selected.size > 0 && deviceImageSet?.size > 0) {
       return () => {
         [...selected.keys()].map((s) => {
           const img = deviceImageSet[s];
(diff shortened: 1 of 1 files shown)


adonispuente:
I couldnt find the ticket referenced in the commit message, and as someone who isnt familiar with the groups page im not sure what im looking for in the video. Could you provide more insight into what this change is for? 
Or maybe @mkholjuraev or @gkarat have a better idea

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -187,7 +187,7 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
 
   //enable disable bulk update based on selection, must refactor
   useEffect(() => {
-    if (selected.size > 0) {
+    if (selected.size > 0 && deviceImageSet?.size > 0) {
```
LiorKGOW, line 190: Shouldn't `deviceImageSet` be added as a dependency to the useEffect's dependency array ? 🤔 
The condition could change when it is changed, which would effect the Effect

LiorKGOW (changes requested):

src/components/GroupSystems/GroupImmutableSystems.js
acosferreira: @LiorKGOW included the validation on useEffect but I kept the condition cause we should not get in this validation if there is no imageSet

LiorKGOW (dismissed):
Unfortunately I can't reproduce the bug that was mentioned on the ticket
Could someone have a look ?
@RedHatInsights/team-interact

LiorKGOW (commented):
Unfortunately I can't reproduce the bug that was mentioned on the ticket
Could someone have a look ?
@RedHatInsights/team-interact

LiorKGOW (approved):
Works fine to me 
Thank you @acosferreira

PR #1968, opened 2023-08-10
chore(ESSNTL-4076): Upgrade react-router-dom to version 6

mkholjuraev:
Resolves: https://issues.redhat.com/browse/ESSNTL-4076

Upgrades react-router to v6 and refactors all class components to function components

PR #2123, opened 2023-12-14
fix(THEEDGE-3764): fixes LoadingCard component

ldjebran:
A regression happened from latest changes (https://github.com/RedHatInsights/insights-inventory-frontend/pull/1968) when used in federated mode, preventing the model dialogs to open, as no way in federated mode to find the modelId. 

This PR revert the change related of using useParams in function "Clickable" and add the full pathname to the url, also needed as in federated mode the path was truncated for example when used in edge-management app the path was "/edge/ipv4", with this change the path is correct and looks like "/edge/inventory/9ea55e32-6e30-4288-bf08-0a877c7fd306/ipv4" 

FIXES: https://issues.redhat.com/browse/THEEDGE-3764

fixes-inventory-loading-card-component-2023-12-14 11-23

### DIFF
--- a/src/components/GeneralInfo/BiosCard/__snapshots__/BiosCard.test.js.snap
+++ b/src/components/GeneralInfo/BiosCard/__snapshots__/BiosCard.test.js.snap
@@ -1311,11 +1311,10 @@ exports[`BiosCard should render extra 1`] = `
                                           value="1 tests"
                                         >
                                           <Link
-                                            to="./undefined"
+                                            to="localhost:3000/example/path/undefined"
                                           >
                                             <a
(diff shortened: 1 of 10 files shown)


gkarat (approved):
Thanks @ldjebran for the PR. I tested it locally and it works fine. 

However, I think the reason why edge is having such issues is due to the fact that it hasn't yet migrated to react-router v6, uses its own Router and doesn't set react-router as singleton in it webpack config. This way, by sharing the same router instance across apps, the federated module will also respect the pathname and recognise route params. I recommend to look into this and reserve some time in the soonest future to avoid such bugs and workarounds.

PR #2124, opened 2023-12-14
fix(RHINENG-715): Redirect on non-existent group

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-715.

Redirect back to /groups if non-existent group is visited.

## How to test

1. Go to any group details page
2. Delete the group
3. Make sure you are redirected to /groups
4. Make sure if you navigate back to the group details page you are again redirected to /groups

### DIFF
--- a/src/components/InventoryGroupDetail/InventoryGroupDetail.js
+++ b/src/components/InventoryGroupDetail/InventoryGroupDetail.js
@@ -41,7 +41,9 @@ const InventoryGroupDetail = ({ groupId }) => {
   );
 
   const dispatch = useDispatch();
-  const { data, fulfilled } = useSelector((state) => state.groupDetail);
+  const { data, fulfilled, rejected, error } = useSelector(
+    (state) => state.groupDetail
+  );
   const navigate = useInsightsNavigate();
 
   const chrome = useChrome();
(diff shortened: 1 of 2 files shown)


src/components/InventoryGroupDetail/InventoryGroupDetail.js
```diff
@@ -110,6 +108,14 @@ const InventoryGroupDetail = ({ groupId }) => {
       }
     }
   }, [data]);
+
+  if (
+    (fulfilled && data.total === 0) || // group does not exist
+    (rejected && error.status === 400) // group name not correct
```
gkarat, line 114: This is added as extra condition because HBI API responds with 400 if the group ID is not of the 4 UUID format.

adonispuente (approved):
This is working as intended, im not seeing any problems. Its causing an iqe test to fail, details below. QA for inventory is off atm but it may be that we need to make changes or the test needs updating. However pending qa return this lgtm

An IQE test is failing 
10:44:19 ERROR tests/groups/test_groups.py::test_navigate_to_groups - iqe_host_invento...
 is failing.

gkarat:
Yeah, this seems to be happening on all PRs.

PR #2125, opened 2023-12-17
fix(RHINENG-2392): Show refresh link in group notifications

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-2392.

This adds a refresh link to the success notification when systems are added or removed from a group. The data is no longer reloaded automatically due to possible inconsistency in the results.

⚠️ When adding new systems, the view is still refreshed under submit. This will be fixed in a separate PR/ticket.

## How to test

1. Go to any group details page
2. Try to remove or add new systems
3. Check that the success notification has a refresh action and on click the page is reloaded with updated results

## Screenshots

image

### DIFF
--- a/src/components/GroupSystems/GroupImmutableSystems.js
+++ b/src/components/GroupSystems/GroupImmutableSystems.js
@@ -2,7 +2,6 @@ import { TableVariant } from '@patternfly/react-table';
 import PropTypes from 'prop-types';
 import React, { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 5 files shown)


adonispuente (approved):
Same IQE test is failing on this one , ERROR at setup of test_navigate_to_groups, but the PR itself looks good to me. I dont believe that failure is related to this PR. 

Just as a side note, we already reload the page when we add a system, do we want this notification to appear with the refresh option even though we already preform the action?

gkarat:
@adonispuente, the automatic refresh on adding systems will go away soon (see my caption in the description) because it still can provide inconsistent results 🙂

PR #2128, opened 2023-12-20
refactor(InventoryTable): Deprecate unused actions

gkarat:
This removes some actions definitions and action creators that are not used in Inventory and other applications.


LiorKGOW (approved):
I have checked and everything seems fine in my local environment after these changes, I agree we should remove those unused actions

Thank you @gkarat !

PR #2129, opened 2023-12-21
refactor: InventoryTable 

bastilian:


mkholjuraev:
@bastilian It seems it is better to be updated with the master before I review it. Because, there maybe a lot of changes that affect the behavior.

src/components/InventoryTable/InventoryTable.js
```diff
@@ -1,58 +1,26 @@
...
 import { useSearchParams } from 'react-router-dom';
 import { ACTION_TYPES } from '../../store/action-types';
 
-/**
- * A helper function to store props and to always return the latest state.
- * For example, EntityTableToolbar wraps OnRefreshData in a callback, so we need this
- * to get the latest props and not the props at the time of when the function is
- * being wrapped in callback.
- */
-export const inventoryCache = () => {
```
bastilian, line 31: The inventoryCache seems unnecessary, since we hold these params and props in many places already.

src/Utilities/Wrapper.js
```diff
@@ -3,13 +3,7 @@ import PropTypes from 'prop-types';
 import { usePermissionsWithContext } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
 import { GENERAL_HOSTS_READ_PERMISSIONS } from '../constants';
 
-const RenderWrapper = ({
-  cmp: Component,
-  isRbacEnabled,
```
bastilian, line 8: This prop is not really needed or used (please double check)

src/Routes.js
```diff
@@ -46,7 +44,6 @@ export const AccountStatContext = createContext({
 });
 
 export const Routes = () => {
-  const searchParams = useMemo(() => getSearchParams(), []);
```
bastilian, line 49: We shouldn't pass this in here. all components that use params lower in the tree should access theme vie the react router (params) hook

src/components/InventoryTable/InventoryList.js
```diff
@@ -1,72 +1,13 @@
...
 import './InventoryList.scss';
-import isEqual from 'lodash/isEqual';
 import AccessDenied from '../../Utilities/AccessDenied';
 
-const convertItem = ({ children, isOpen, ...item }) => item;
-
-/**
- * Component that works as a side channel for consumers to notify inventory of new data changes.
- */
-const ContextInventoryList = ({
```
bastilian, line 16: This is a redundant wrapper. Removing it will simplify the structure and should reduce the render burden.

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -237,35 +237,15 @@ const EntityTableToolbar = ({
...
-  const updateData = (config) => {
-    if (hasAccess) {
-      onRefreshDataInner(config);
-    }
-  };
-
-  /**
-   * Debounced `updateData` function.
-   */
-  const debouncedRefresh = useCallback(
```
bastilian, line 266: This is mainly the culprit as far as I remember, starting from line 247. 

These functions and callbacks were called when they shouldn't have been, due to the debouncing here. These changes in the table and the toolbar change that a little and consolidate everything to only call the ONE debounced refresh function in the EntityTable above.

src/components/InventoryTable/InventoryTable.js
```diff
@@ -117,128 +79,73 @@ const InventoryTable = forwardRef(
...
-              cachedProps.getEntities
-            )
-          );
+          dispatch(loadSystems(params, showTags, getEntities));
         }
       }
     };
 
-    const prevFilters = useRef(customFilters);
+    const debouncedRefresh = debounce((config) => onRefreshData(config), 800);
```
bastilian, line 134: this is the ONE debounced refresh function that should be called. And the only way that any refresh should be triggered.

src/components/InventoryTable/InventoryList.js
```diff
@@ -79,31 +20,20 @@ const InventoryList = React.forwardRef(
...
       </div>
     ) : (
-      <ContextInventoryList
-        {...props}
-        activeFilters={activeFilters}
-        onRefreshData={onRefreshData}
-      />
+      <Grid gutter="sm" className="ins-inventory-list">
+        <GridItem span={12}>
+          <InventoryEntityTable {...props} onRefreshData={onRefreshData} />
```
johnsonm325, line 25: It seems that removing the redundant wrapper out of here has broken the fetch when sorting has changed. @bastilian was your intention to pass `onRefreshData` to `EntityTable` and refresh the data when sorting changes, from within the component?

bastilian: @johnsonm325 Do you have the current changes somewhere I can take a look at them?

PR #2130, opened 2023-12-22
Feat(RHINENG-4796): Add Public IP & FQDN to Infrastructure Card in Host Details Page

LiorKGOW:
# Description

Add Public IP & FQDN to Infrastructure Card in Host Details Page
The necessary changes that address the issue described on the ticket

Associated Jira ticket: [#RHINENG-4796](https://issues.redhat.com/browse/RHINENG-4796) 

# How to test the PR

Ensure a host containing the properties `public_ipv4_addresses` (& `public_dns`) is presenting those properties on its Infrastructure Card in its Host Details page
(It could also be with two different hosts that each contain one property)

Expected Functionality: 
- Hosts that have a property, should present it.
- Hosts that do not have a property, should present 'Not available' instead.

To test, you can search for a host named `lior_test_ip-172-31-89-28.ec2.internal` on `insights-qa` account. 
This host contains the necessary `public_ipv4_addresses` property that should be presented on the Infrastructure Card.

### Please Read
*However, because we are not yet collecting the `public_dns`, we can not create a host that contain this property.
Because both of these properties are part of the same SystemProfile object that is received in the relevant API response, both properties have the same implementation in this PR, which should work.*

*Here is a link to the scheme of `SystemProfile` that is being returned:
https://github.com/lzap/inventory-schemas/blob/fe94ba7ab5561abf1e1d644645b0110d6e64fbc0/schemas/system_profile/v1.yaml#L389-L406*

*The API endpoint that receives SystemProfile that SHOULD contain both properties, but now containing only 1 because this host does not contain the `public_dns` property (see the screenshots attached):*

Screenshot from 2024-01-08 15-18-50

Screenshot from 2024-01-08 15-19-06

*Tested locally that when a host has also the `public_dns` property, it is being presented in the UI alongside the `public_ipv4_addresses` property. 
Attached screenshots of the changes and the UI:*

*Added code that is needed to ensure that the necessary values are presented when received*

Added code that is needed to ensure that the necessary values are presented when received

*The passed values are presented in the UI*

The passed values are presented in the UI

# Before the change

Host Details Page for a host that contains the `public_ipv4_addresses` property
Green - name of host so we can easily find it
Yellow - Infrastructure Card does not contain the mentioned properties

Before: Host Details Page for a host that contains the public_ipv4_addresses property

# After the change

Host Details Page for a host that contains the `public_ipv4_addresses` property
Green - name of host so we can easily find it (same host name)
Orange - New Public IP property that is showing on the Infrastructure Card because this host contains the necessary property
Yellow - New FQDN property is being presented on the card, but is showing Not Available for this host because it does not contain the necessary `public_dns` property

After: Host Details Page for a host that contains the public_ipv4_addresses property

# Checklist

- [X] Check the property `public_ipv4_addresses`
- [X] Check the property `public_dns` -> Need to create a host containing this prop (blocked) but tested locally (explained in the Please Read section)
- [X] update appropriate snapshots
- [X] Add screenshots

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -164,7 +164,9 @@ export const calculateSystemProfile = ({
     systemProfile['host_type'] = hostTypeFilter;
   }
 
-  return generateFilter({ system_profile: systemProfile });
+  return generateFilter({
+    system_profile: systemProfile,
+  });
 };
 
 export const filtersReducer = (acc, filter = {}) => ({
@@ -249,7 +251,7 @@ export async function getEntities(
           ),
         };
       } catch (e) {
-                console.error(e); // eslint-disable-line
+        console.error(e); // eslint-disable-line
       }
     }
 
--- a/src/components/GeneralInfo/GeneralInformation/__snapshots__/GeneralInformation.test.js.snap
+++ b/src/components/GeneralInfo/GeneralInformation/__snapshots__/GeneralInformation.test.js.snap
@@ -370,6 +370,18 @@ exports[`GeneralInformation custom components should not render BiosCardWrapper 
                       </dd>
                       <dt
                         class=""
+                        data-ouia-component-id="Public IP title"
+                      >
+                        Public IP
+                      </dt>
+                      <dd
+                        class=""
+                        data-ouia-component-id="Public IP value"
+                      >
+                        Not available
+                      </dd>
+                      <dt
+                        class=""
                         data-ouia-component-id="IPv4 addresses title"
                       >
                         IPv4 addresses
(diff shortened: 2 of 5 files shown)


LiorKGOW:
Thank you @gkarat for helping me with this PR ! 
It is now ready for review ✌🏼

gkarat (changes requested):
Everything looks good to me! However, I believe we better no to touch `value` for the IPv4, IPv6, and Interfaces/NICs fields. The problem is that if any of the arrays for this fields is empty, they will render "Not available addresses" text which is not correct. It must show "0 addresses" IIRC.

image

So let's just focus on the newly added fields (DNS, Public IP) and, if would be necessary in the future change the value logic for other fields in a separate PR.

LiorKGOW:
Thank you for the review @gkarat !  

I have removed the second commit like you requested 
The PR is now ready for another review 👍🏼

gkarat (approved):
LGTM! Thank you @LiorKGOW

PR #2133, opened 2024-01-07
Chore: Update README With 'Common Problems Users Might Encounter' Section

LiorKGOW:
### Description

* This PR comes to address common installation problems that user might encounter when they first install our application or update to the latest version

* This is just a suggestion of course, in case someone sees a better way to write this is more then welcome to suggest it or add another common problem that might be useful to mention in this section 😇 

### Explanation About the commits included

1. After a conversation with @mkholjuraev, we thought a README update might be needed with the case I mentioned in the first commit

2. The second commit fixes typos that were in the README in 'Release process change in stage env' section

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,3 +95,18 @@ These applications import `InventoryTable` component through federated modules:
 
 We use the new release process in stage environment that uses containers for deplyoment. Containerized Frontends are the new way to to build, package, and deploy the applicaitons. 
 
+## Common Problems You Might Encounter
+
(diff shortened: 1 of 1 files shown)


LiorKGOW:
@RedHatInsights/team-interact 

Do you guys think this is necessary? 
Do you have a better suggestion?

leSamo (approved):
Nice one, thanks for propagating information in visible places.

PR #2135, opened 2024-01-10
fix(GroupDetails): THEEDGE-3782 - Fixes Group details edge Update button

ldjebran:
When systems to update was selected the Update button was not updated as expected. 

Now the button is enabled when the selected systems are updatable and have the same image set, in all other cases the button is disabled. 

FIXES: https://issues.redhat.com/browse/THEEDGE-3782

[Screencast from 2024-01-10 13-24-37.webm](https://github.com/RedHatInsights/insights-inventory-frontend/assets/131553/6c3845c7-066d-4346-a1a0-703b3abc4e1d)


mkholjuraev (approved):
LGTM! @ldjebran thank you, works as expected.

PR #2136, opened 2024-01-11
add filter by type nil to avoid erroron active tab

acosferreira:
Add parameter on filter to return correct data for conventional devices


adeliaferreira (approved):
lgtm

mkholjuraev (approved):
LGTM! @acosferreira thank you!

PR #2137, opened 2024-01-11
fix(THEEDGE-3790): fix conventional filter when no host typ informed

acosferreira:
Add filter by host type as default to avoid list immutable and edge in the same list
image


omaciel (approved):
/lgtm

gkarat (approved):

PR #2142, opened 2024-02-01
fix(RHINENG-7744): fix status column default value in Immutable tabs

mkholjuraev:
Fixes: https://issues.redhat.com/browse/RHINENG-7744

Displays 'unknown' status for devices that are not properly uploaded.


gkarat (approved):
LGTM!

PR #2143, opened 2024-02-01
fix(RHINENG-7745): covert custom deep compare function to lodash isEq…

mkholjuraev:
…ual in useColumns

Fixes: https://issues.redhat.com/browse/RHINENG-7745

To test:
1. Run this PR with vulnerability app
2. Navigate to some CVE detail page
3. Switch between conventional and immutable tabs fast 
4. Repeat this multiple times
5. Observe that the table does not break

### DIFF
--- a/src/components/InventoryTable/hooks/useColumns.js
+++ b/src/components/InventoryTable/hooks/useColumns.js
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import { mergeArraysByKey } from '@redhat-cloud-services/frontend-components-utilities/helpers/helpers';
 import { defaultColumns } from '../../../store/entities';
-import { isEqual } from 'lodash';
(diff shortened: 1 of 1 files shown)


src/components/InventoryTable/hooks/useColumns.js
```diff
@@ -2,6 +2,7 @@ import { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import { mergeArraysByKey } from '@redhat-cloud-services/frontend-components-utilities/helpers/helpers';
 import { defaultColumns } from '../../../store/entities';
+import { isEqual } from 'lodash';
```
gkarat, line 5:
```suggestion
import isEqual from 'lodash/isEqual';
```

gkarat (approved):
LGTM! Just one minor change so that we import as least as possible from 'lodash'.

PR #2144, opened 2024-02-01
fix(THEEDGE-3793): Fix update option showed when all system are updated in a group

adeliaferreira:
According the ticket THEEDGE-3793, when we have a group with all the system "updated", we show on kebab menu the optiont o update as active. This is solved once a system with update pendent is added to the group.

The reason for this error is the deviceList been empty and it generate a false positive on the previous condition. This PR will fix the issue.

### DIFF
--- a/src/components/GroupSystems/GroupImmutableSystems.js
+++ b/src/components/GroupSystems/GroupImmutableSystems.js
@@ -131,7 +131,6 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
       enhancedConfig,
       showTags
     );
-
     const mapDeviceIds = mapDefaultData(defaultData.results);
     const updateInfo = await getUpdateInfo(groupId);
     setDeviceData(updateInfo?.update_devices_uuids);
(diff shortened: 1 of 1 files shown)


LightOfHeaven1994:
@adeliaferreira fyi, we are looking into failing tests

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -131,7 +131,6 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
       enhancedConfig,
       showTags
     );
-
     const mapDeviceIds = mapDefaultData(defaultData.results);
     const updateInfo = await getUpdateInfo(groupId);
     setDeviceData(updateInfo?.update_devices_uuids);
```
gkarat, line 136: nit-pick:
```suggestion
    setDeviceData(updateInfo?.update_devices_uuids || []);
```

since the server responds with `null` if there are no systems pending updates, I would recommend setting deviceData to [] so that it's more safe to work with this state variable across the component. For example, in that case, you'd rewrite the "Update" button condition (if it's disabled or not) to just `!deviceData.includes(row.id)`

gkarat (approved):
LGTM! just a minor suggestion, but feel free to merge it as it is

src/components/GroupSystems/GroupImmutableSystems.js
adeliaferreira: Changed for your suggestion and rebased the branch to help on merge.

Thanks for your review

PR #2155, opened 2024-03-01
fix(RHINENG-8372): Fix Group details not loading Edge devices

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-8372.

The Immutable tab was playing up because of the deviceData being undefined or null which was not treated as a separate edge case.

## How to test

1. Have a group with both conventional and immutable systems
2. Navigate to this group
3. Make sure you can switch between both immutable and conventional systems tabs without crash (which is reproducible on stage)


mkholjuraev (approved):
LGTM! opening a group detail page works for both conventional and immutable tabs

PR #2170, opened 2024-04-03
fix(RHINENG-8899): Update Staleness conditionals

adonispuente:
We missed an edge case where staleness should not be the same as stale warning. The form should not be submitable if an error is populated. 
The order should always be staleness < warning < deletion.
Warning should also never be the same as deletion 

Before:
Screenshot 2024-04-03 at 10 35 43 AM

After: 
Screenshot 2024-04-03 at 10 34 53 AM


computercamplove:
@adonispuente works as expected - not able to select same options from settings menu.
please ignore failed tests for filtering os in pr_check

AsToNlele (approved):
LGTM

PR #2171, opened 2024-04-08
fix(RHINENG-9239, RHINENG-9260): Keep filtering in Hybrid Inventory Tabs separate, filter reset when Add Systems modal closes

LiorKGOW:
## Description
This PR addresses issues: 
[#RHINENG-9239](https://issues.redhat.com/browse/RHINENG-9239)
[#RHINENG-9260](https://issues.redhat.com/browse/RHINENG-9260)

Inventory group's details page: Filtering of each tab should be kept separate and saved even when switching between tabs (Conventional and Immutable)
Also, this PR includes 2 commits that change state variable names (tab & activeTabKey) in GroupTabDetails component to improve code readability
4th commit - added Sebastian's suggestion 
5th commit - fixed an inconsistency between the different tabs on the "Add Systems" modal

This PR introduces a store instance for each tab, which holds it's own filters for it's InventoryTable. In order to create a new instance of the store for the InventoryTable, you need to provide InventoryTable component with the `isolateStore` prop to render it with a Provider Wrapper which holds a new store instance alongside with the InventoryTable.
I have removed the component state inside InventoryGroupDetail/GroupTabDetails.js and put the components inside the returned JSX of the component.
Regarding the Immutable tab inside the "Add Systems" Modal, the filters there were handled a bit differently, you can find small comments inside the 5th commit message which solves this inconsistency issue which was created in previous commits

### DIFF
--- a/src/components/GroupSystems/GroupImmutableSystems.js
+++ b/src/components/GroupSystems/GroupImmutableSystems.js
@@ -241,113 +241,112 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
           refreshTable={() => true}
         />
       )}
-      {!addToGroupModalOpen && (
-        <InventoryTable
-          columns={(columns) => mergeColumns(prepareColumns(columns))}
-          hideFilters={{ hostGroupFilter: true }}
-          // getEntities={entities}
-          getEntities={customGetEntities}
-          tableProps={{
-            isStickyHeader: true,
-            variant: TableVariant.compact,
-            canSelectAll: false,
-            actionResolver: (row) => [
-              {
-                title: (
-                  <ActionDropdownItem
-                    requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
-                      groupId
-                    )}
-                    noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
-                    onClick={() => {
-                      setCurrentSystem([row]);
-                      setRemoveHostsFromGroupModalOpen(true);
-                    }}
-                  >
-                    Remove from group
-                  </ActionDropdownItem>
-                ),
-              },
-              {
-                title: (
-                  <ActionDropdownItem
-                    isAriaDisabled={
-                      deviceData === null || !deviceData.includes(row.id)
-                    }
-                    requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
-                      groupId
-                    )}
-                    noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
-                    onClick={() => {
-                      setCurrentSystem([row]);
-                      navigate(`/insights/inventory/${row.id}/update`);
-                    }}
-                  >
-                    Update
-                  </ActionDropdownItem>
-                ),
-              },
+      <InventoryTable
+        isolateStore={true}
+        columns={(columns) => mergeColumns(prepareColumns(columns))}
+        hideFilters={{ hostGroupFilter: true }}
+        // getEntities={entities}
+        getEntities={customGetEntities}
+        tableProps={{
+          isStickyHeader: true,
+          variant: TableVariant.compact,
+          canSelectAll: false,
+          actionResolver: (row) => [
+            {
+              title: (
+                <ActionDropdownItem
+                  requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
+                    groupId
+                  )}
+                  noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
+                  onClick={() => {
+                    setCurrentSystem([row]);
+                    setRemoveHostsFromGroupModalOpen(true);
+                  }}
+                >
+                  Remove from group
+                </ActionDropdownItem>
+              ),
+            },
+            {
+              title: (
+                <ActionDropdownItem
+                  isAriaDisabled={
+                    deviceData === null || !deviceData.includes(row.id)
+                  }
+                  requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
+                    groupId
+                  )}
+                  noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
+                  onClick={() => {
+                    setCurrentSystem([row]);
+                    navigate(`/insights/inventory/${row.id}/update`);
+                  }}
+                >
+                  Update
+                </ActionDropdownItem>
+              ),
+            },
+          ],
+        }}
+        actionsConfig={{
+          actions: [
+            [
+              <div key="primary-actions" className="pf-v5-c-action-list">
+                <ActionButton
+                  key="add-systems-button"
+                  requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
+                    groupId
+                  )}
+                  noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
+                  onClick={() => {
+                    dispatch(clearEntitiesAction());
+                    setAddToGroupModalOpen(true);
+                  }}
+                  ouiaId="add-systems-button"
+                >
+                  Add systems
+                </ActionButton>
+                <ActionButton
+                  requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
+                    groupId
+                  )}
+                  noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
+                  key="update-systems-button"
+                  onClick={() => {
+                    setupdateDevice(true);
+                    handleUpdateSelected();
+                  }}
+                  ouiaId="update-systems-button"
+                  isAriaDisabled={!canUpdate}
+                >
+                  Update
+                </ActionButton>
+              </div>,
             ],
-          }}
-          actionsConfig={{
-            actions: [
-              [
-                <div key="primary-actions" className="pf-v5-c-action-list">
-                  <ActionButton
-                    key="add-systems-button"
-                    requiredPermissions={REQUIRED_PERMISSIONS_TO_MODIFY_GROUP(
-                      groupId
-                    )}
-                    noAccessTooltip={NO_MODIFY_GROUP_TOOLTIP_MESSAGE}
-                    onClick={() => {
-                      dispatch(clearEntitiesAction());
-                      setAddToGroupModalOpen(true);
-                    }}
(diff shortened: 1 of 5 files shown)


src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -206,42 +207,9 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
...
-          locationProp={useLocation}
-          notificationProp={notificationProp}
-          paramsProp={useParams}
-          updateModal={updateModal}
-          setUpdateModal={setUpdateModal}
-          refreshTable={() => true}
-        />
-      )}
-      {!addToGroupModalOpen && (
+  const InventoryTableWrapper = useMemo(
```
bastilian, line 210: Is there any specific reason to put it into a `useMemo`?

bastilian: It would also be better to put this into the `components/InventoryTable` and only wrap everything in a redux Provider if a `store` prop is passed.
Similar to what the `AsyncInventory` does: https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/modules/AsyncInventory.js#L24

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -23,83 +24,78 @@ const GroupTabDetailsWrapper = ({
...
+  const ConventionalTabTable = useMemo(
     () => (
       <GroupSystems
         groupName={groupName}
         groupId={groupId}
         hostType={hybridInventoryTabKeys.conventional.key}
       />
     ),
-    [groupId, groupName]
+    []
```
bastilian, line 44: Why do these need to be removed? And also if there is no specific reason for them to be in a `useMemo` we should maybe take them out and put them into the `return`.

src/components/GroupSystems/GroupImmutableSystems.js
LiorKGOW: > Is there any specific reason to put it into a `useMemo` ?

I have put this in a `useMemo` because without it, this value would get calculated on every render of the component which would create another instance of the store

bastilian: @LiorKGOW I think in this case you'd want to put the result of `getStore()` into a `useMemo`, but not the whole component.

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -23,83 +24,78 @@ const GroupTabDetailsWrapper = ({
...
+  const [hybridInventoryTabKey, setHybridInventoryTabKey] = useState(
+    hybridInventoryTabKeys.conventional.key
+  );
 
   const { hasAccess: canViewHosts } = usePermissionsWithContext(
     REQUIRED_PERMISSIONS_TO_READ_GROUP_HOSTS(groupId)
   );
-  const conventionalSystemsContent = useMemo(
+
+  const ConventionalTabTable = useMemo(
```
bastilian, line 36: These should be removed and not memoize rendered components. See https://react.dev/reference/react/useMemo#memoizing-individual-jsx-nodes

src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -109,8 +105,12 @@ const GroupTabDetailsWrapper = ({
...
         </PageSection>
       </Tab>
-      <Tab eventKey={1} title="Group info" aria-label="Group info tab">
-        {activeTabKey === 1 && ( // helps to lazy load the component
+      <Tab
+        eventKey={groupTabKeys.groupInfo}
+        title="Group info"
+        aria-label="Group info tab"
+      >
+        {groupTabKey === groupTabKeys.groupInfo && ( // helps to lazy load the component
```
bastilian, line 113: > // helps to lazy load the component

How?

LiorKGOW: Though you are right, this is another set of Tabs 
We need to save the filters only for the "Hybrid Inventory Tabs" (the inner Tabs component), according to the ticket

In case we need to also keep the filtering when switching to the "Group Info" tab, the fix would probably be what you mentioned here. But, at the moment, it was not mentioned on the ticket

src/components/GroupSystems/GroupImmutableSystems.js
LiorKGOW: I have changed the PR according to this comment 
Thank you @bastilian !

LiorKGOW:
Hey @bastilian I have went over your comments and updated the PR accordingly
It is ready for another review ✌🏼

src/components/InventoryTable/InventoryTable.js
```diff
@@ -349,4 +357,21 @@ InventoryTable.propTypes = {
   showNoGroupOption: PropTypes.bool, // group filter option
 };
 
-export default InventoryTable;
+const InventoryTableWrapper = forwardRef(
+  ({ needIsolatedStore = false, ...props }, ref) => {
+    const Wrapper = needIsolatedStore ? Provider : React.Fragment;
+    const store = needIsolatedStore ? useMemo(() => getStore(), []) : undefined;
```
bastilian, line 363: This is calling a hook conditionally, it should rather be this and call the hook an each render, otherwise React may throw an error about this and the change of hooks called: 
```js
const store = useMemo(() => {
  return needIsolatedStore ? getStore() undefined
}, [needIsolatedStore]);
```

src/components/InventoryTable/InventoryTable.js
```diff
@@ -349,4 +357,21 @@ InventoryTable.propTypes = {
...
+    return (
+      <Wrapper {...(store ? { store } : {})}>
+        <InventoryTable {...props} ref={ref} />
+      </Wrapper>
+    );
+  }
+);
+
+InventoryTableWrapper.propTypes = {
+  needIsolatedStore: PropTypes.bool,
```
bastilian, line 374: tbh. `isolateStore` might be better.

LiorKGOW: I will change it 👍🏼

src/components/InventoryTable/InventoryTable.js
LiorKGOW: I have added it 👍🏼

src/components/InventoryTable/InventoryTable.js
```diff
@@ -349,4 +357,21 @@ InventoryTable.propTypes = {
   showNoGroupOption: PropTypes.bool, // group filter option
 };
 
-export default InventoryTable;
+const InventoryTableWrapper = forwardRef(
+  ({ needIsolatedStore = false, ...props }, ref) => {
+    const Wrapper = needIsolatedStore ? Provider : React.Fragment;
```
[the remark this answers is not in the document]
LiorKGOW: I have added it 👍🏼

LiorKGOW:
Thank you for the quick review @bastilian !
I have updated the PR with your suggestions 
It is ready for another review ✌🏼

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -241,113 +241,112 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
...
-                      setCurrentSystem([row]);
-                      navigate(`/insights/inventory/${row.id}/update`);
-                    }}
-                  >
-                    Update
-                  </ActionDropdownItem>
-                ),
-              },
+      <InventoryTable
+        isolateStore={true}
```
bastilian, line 245: With boolean JSX props you can also just add the prop and it will default to `true`, like the `showTags` prop.

bastilian:
@LiorKGOW The code looks good to me, however, while testing I noticed an inconsistency. On the Systems tab both the Conventional and the Immutable Systems persist the filters properly, but in the modal when it closes and the filters should not persist the Immutable tab still keeps them after closing the modal:



https://github.com/RedHatInsights/insights-inventory-frontend/assets/7757/c6382fe4-33dc-4dfa-a108-57ef0dea3178

src/components/InventoryTable/InventoryTable.js
```diff
@@ -349,4 +357,24 @@ InventoryTable.propTypes = {
   showNoGroupOption: PropTypes.bool, // group filter option
 };
 
-export default InventoryTable;
+const InventoryTableWrapper = forwardRef(
+  ({ isolateStore = false, ...props }, ref) => {
+    const store = useMemo(
+      () => (isolateStore ? getStore() : undefined),
```
gkarat, line 363: Do we want to limit the reducers to only those we use with InventoryTable? Something we discussed earlier. But it's more of "nice to have", so if you don't want to work on this, just leave the "TODO" note

LiorKGOW:
> @LiorKGOW The code looks good to me, however, while testing I noticed an inconsistency. On the Systems tab both the Conventional and the Immutable Systems persist the filters properly, but in the modal when it closes and the filters should not persist the Immutable tab still keeps them after closing the modal:
> 
>  immu.mp4

@bastilian that is the expected functionality :) 

Please have a look at this ticket:
[#RHINENG-4108](https://issues.redhat.com/browse/RHINENG-4108?focusedId=24476181&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24476181) was the original ticket (one which has a bit more info on this matter) 
Because #RHINENG-4108 was already released, I have opened [this ticket](https://issues.redhat.com/browse/RHINENG-9260) only for the immutable tab (for the second issue that Zhamilya mentioned in the comments)

src/components/GroupSystems/GroupImmutableSystems.js
LiorKGOW: Oh I haven't known that
I will add it 👍🏼 
Thanks !

src/components/InventoryTable/InventoryTable.js
LiorKGOW: I think we will work on it in a different PR
I'll add a todo note like you suggested @gkarat

LiorKGOW: What do you think about this note? 
`// TODO: Create a partial store containing only relevant Slices & Reducers relevant to the InventoryTable`
@gkarat

LiorKGOW:
@bastilian I have added the code that fixes the inconsistency problem you mentioned, you can find it in the last commit 👍🏼

LiorKGOW:
rebased on top of latest master and repushed

johnsonm325 (approved):

gkarat (commented):
There is one regression: if URL contains any filters/pagination, it's no longer applied for conventional tab. I will try to find out what caused it and fix it

PR #2174, opened 2024-04-11
fix(RHINENG-3920): Handle Groups 403

AsToNlele:
[RHINENG-3920](https://issues.redhat.com/browse/RHINENG-3920)

Previously when the user didn't have `inventory:groups:read` permission he would get this error.
image

Now this request is only made when user has `inventory:groups:read` permission.

Test without sufficient permissions:
1. Make sure your account doesn't have `inventory:groups:read`. If you don't have such account, you can use `acelakov-groups-test` and the usual password
2. Go to Inventory > Systems
3. Check the groups filter, it should be empty and no request should be made to hosts.

Test with sufficient permissions:
1. User `acelakov-groups` has the correct permissions
2. Go to Inventory > Systems
3. Check the groups filter and it should be populated

### DIFF
--- a/src/Utilities/hooks/useFetchBatched.js
+++ b/src/Utilities/hooks/useFetchBatched.js
@@ -10,7 +10,6 @@ const useFetchBatched = () => {
 
       const results = resolve(
         [...new Array(pages)].map(
-          // eslint-disable-next-line camelcase
           (_, pageIdx) => () =>
             fetchFunction(filter, { page: pageIdx + 1, per_page: batchSize })
         )
@@ -27,6 +26,27 @@ const useFetchBatched = () => {
             fetchFunction(
               list.slice(batchSize * pageIdx, batchSize * (pageIdx + 1))
             )
+        )
+      );
+
+      return results;
+    },
+    pageOffsetfetchBatched: (
+      fetchFunction,
+      total,
+      filter,
+      batchSize = 50,
+      pageOffset = 0
+    ) => {
+      const pages = Math.ceil(total / batchSize) || 1;
+
+      const results = resolve(
+        [...new Array(pages)].map(
+          (_, pageIdx) => () =>
+            fetchFunction(filter, {
+              page: pageOffset + pageIdx + 1,
+              per_page: batchSize,
+            })
         )
       );
 
--- a/src/components/filters/useGroupFilter.js
+++ b/src/components/filters/useGroupFilter.js
@@ -4,6 +4,8 @@ import useFetchBatched from '../../Utilities/hooks/useFetchBatched';
 import { HOST_GROUP_CHIP } from '../../Utilities/index';
 import { getGroups } from '../InventoryGroups/utils/api';
 import SearchableGroupFilter from './SearchableGroupFilter';
+import { GENERAL_GROUPS_READ_PERMISSION } from '../../constants';
+import { usePermissionsWithContext } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
 
 export const groupFilterState = { hostGroupFilter: null };
 export const GROUP_FILTER = 'GROUP_FILTER';
@@ -36,19 +38,42 @@ export const buildHostGroupChips = (selectedGroups = []) => {
     : [];
 };
 
+const REQUIRED_PERMISSIONS = [GENERAL_GROUPS_READ_PERMISSION];
+
 const useGroupFilter = (showNoGroupOption = false) => {
-  const { fetchBatched } = useFetchBatched();
+  const { pageOffsetfetchBatched } = useFetchBatched();
   const [fetchedGroups, setFetchedGroups] = useState([]);
   const [selectedGroupNames, setSelectedGroupNames] = useState([]);
 
+  const { hasAccess } = usePermissionsWithContext(
+    REQUIRED_PERMISSIONS,
+    true,
+    false
+  );
+
   useEffect(() => {
     const fetchOptions = async () => {
+      if (!hasAccess) return;
+
       const firstRequest = !ignore
-        ? await getGroups(undefined, { page: 1, per_page: 1 })
+        ? await getGroups(undefined, { page: 1, per_page: 50 })
         : { total: 0 };
-      const groups = !ignore
-        ? await fetchBatched(getGroups, firstRequest.total)
-        : [];
+
+      const groups =
+        !ignore && firstRequest.total > 50
+          ? await pageOffsetfetchBatched(
+              getGroups,
+              firstRequest.total - 50,
+              {},
+              50,
+              1
+            )
+          : [];
+
+      if (firstRequest.total > 0) {
+        groups.push(firstRequest);
+      }
+
       !ignore && setFetchedGroups(groups.flatMap(({ results }) => results));
     };
 
@@ -59,7 +84,7 @@ const useGroupFilter = (showNoGroupOption = false) => {
     return () => {
       ignore = true;
     };
-  }, []);
+  }, [hasAccess]);
 
   const chips = useMemo(
     () => buildHostGroupChips(selectedGroupNames),
--- a/src/components/InventoryTable/InventoryTable.test.js
+++ b/src/components/InventoryTable/InventoryTable.test.js
@@ -62,6 +62,14 @@ const renderTable = (store, props) => {
 
 jest.mock('../../Utilities/useFeatureFlag');
 
+jest.mock(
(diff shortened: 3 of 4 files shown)


AsToNlele:
Ignore the PR build, looks like its broken now

src/components/filters/useGroupFilter.js
```diff
@@ -46,9 +46,10 @@ const useGroupFilter = (showNoGroupOption = false) => {
       const firstRequest = !ignore
         ? await getGroups(undefined, { page: 1, per_page: 1 })
         : { total: 0 };
-      const groups = !ignore
-        ? await fetchBatched(getGroups, firstRequest.total)
-        : [];
+      const groups =
+        !ignore && firstRequest.total > 0
+          ? await fetchBatched(getGroups, firstRequest.total)
```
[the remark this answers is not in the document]
bastilian: The `p-all` library used by the promise queue hook has a `stopOnError` option that we can use. https://www.npmjs.com/package/p-all#stoponerror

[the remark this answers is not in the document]
bastilian: hm. I just saw the default is true, so it should already do that. <minecraft-villagersound>hm</minecraft-villagersound>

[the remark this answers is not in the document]
AsToNlele: @bastilian thanks for having a look!

I was playing a bit with it and I'm able to cancel the rest. Did you mean to also skip the `firstRequest` and do it all in one batch? We would need to somehow pass up the `total` so we know how many requests it will need and since we're doing 2 concurrent requests I'm not sure how it would be done.

I think we could optimize it a different way. Let's make the `firstRequest` get the first 50, if `total > 0` then we can merge it with the `groups` and the batch will be made only if there's more than 50 total systems. Since we fetched the first 50, we can start the `fetchBatched` from 51..100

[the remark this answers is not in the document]
AsToNlele: and there won't be a need for any cancellations

AsToNlele:
Made a new `useCustomFetchBatched` hook and not edited the original once, since it's not the same. Includes a new parameter which can offset the pages.

Previously the first `groups` request was made to check if we have the permission and fetched only one group. Now the first request gets the first 50 systems, and later is merged to the rest.

In the end we are making one less request, because we use the first requests data and offset the batch request pages by one page.

src/Utilities/hooks/useFetchBatched.js
```diff
@@ -36,3 +36,34 @@ const useFetchBatched = () => {
...
+          (_, pageIdx) => () =>
+            fetchFunction(filter, {
+              page: pageOffset + pageIdx + 1,
+              per_page: batchSize,
+            })
+        )
+      );
+
+      return results;
+    },
```
johnsonm325, line 67: Maybe I'm missing the point of this, but could we not rename this as `pageOffsetFetchBatched` and include it here: https://github.com/RedHatInsights/insights-inventory-frontend/blob/2cedc36f31e0966b4ef869d0cf56d0f2fd448f77/src/Utilities/hooks/useFetchBatched.js#L34C1-L35

Then we could import it as `const { pageOffsetFetchBatched } = useFetchBatched();`

johnsonm325 (commented):
Not sure if this warrants a "Request changes" or not. Want your feedback first.

I'm testing functionality.

src/Utilities/hooks/useFetchBatched.js
bastilian: +1 We should rather extend the existing hook, than create a new one that looks almost exactly the same just with one minor difference. 

On that note, the existing hook already has two very similar functions that it returns, which we should also refactor.

src/components/InventoryGroups/utils/api.js
```diff
@@ -23,7 +23,16 @@ export const getGroups = (
     parameters.length > 0 ? '?' + parameters : ''
   }`;
 
-  return instance.get(path);
+  try {
+    return await instance.get(path);
+  } catch (e) {
+    if (e?.status === 403) {
+      return {
+        total: 0,
```
bastilian, line 31: Why return this here? 

Would this not just silence the issue?

I'm still for failing the whole batch when one request fails, regardless of what error it is. And the catching of those errors should be higher up and if one of the errors was a 403, it should direct the user to a login page.

AsToNlele: In this case the 403 is a legit response from the api, the first request checks if user has permissions to groups. In case he has not, then no more requests are made and you get the total as 0

src/Utilities/hooks/useFetchBatched.js
AsToNlele: Ok I'll extend the existing hook

AsToNlele:
Moved it to the original hook

AsToNlele:
Updating snapshots has changed all single quotes to double quotes in the test file and linting would fail, pretty funny

src/components/InventoryGroups/utils/api.js
bastilian: How can you know it is the/a first request?

AsToNlele:
TODO: Hide column and groups filter as well

src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.js
```diff
@@ -156,6 +158,15 @@ const ConventionalSystemsTab = ({
     setAddHostGroupModalOpen
   );
 
+  const REQUIRED_PERMISSIONS = [GENERAL_GROUPS_READ_PERMISSION];
+
+  const { hasAccess: hasGroupsAccess } = usePermissionsWithContext(
+    REQUIRED_PERMISSIONS,
+    true
+  );
```
gkarat, line 166: I think the call to usePermissionsWithContext would also need the third argument which is responsible for checking resource definitions. If we provide the third argument as `false`, then your function would check for any `inventory:groups:read` permissions including those that resource definitions (= narrowed to some specific groups only):

https://github.com/RedHatInsights/frontend-components/blob/master/packages/utils/doc/RBAC.md#resource-definitions

src/components/filters/useGroupFilter.js
```diff
@@ -36,19 +38,38 @@ export const buildHostGroupChips = (selectedGroups = []) => {
...
 
+const REQUIRED_PERMISSIONS = [GENERAL_GROUPS_READ_PERMISSION];
+
 const useGroupFilter = (showNoGroupOption = false) => {
-  const { fetchBatched } = useFetchBatched();
+  const { pageOffsetfetchBatched } = useFetchBatched();
   const [fetchedGroups, setFetchedGroups] = useState([]);
   const [selectedGroupNames, setSelectedGroupNames] = useState([]);
 
+  const { hasAccess } = usePermissionsWithContext(REQUIRED_PERMISSIONS, true);
```
gkarat, line 48: The same thing as from the previous comment would apply here

src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.js
AsToNlele: thanks, thats perfect

AsToNlele:
reverted hiding the column and filter, keeping it as it was, [slack discussion](https://redhat-internal.slack.com/archives/C02T90PLJ5V/p1715789356364939?thread_ts=1715760963.070589&cid=C02T90PLJ5V)

now the 403 is not happening at all

when the user has no groups permissions - the groups filter is empty

otherwise groups that the user has groups:read permission are shown

src/Utilities/hooks/useFetchBatched.js
```diff
@@ -30,6 +30,28 @@ const useFetchBatched = () => {
...
+      total,
+      filter,
+      batchSize = 50,
+      pageOffset = 0
+    ) => {
+      const pages = Math.ceil(total / batchSize) || 1;
+
+      const results = resolve(
+        [...new Array(pages)].map(
+          // eslint-disable-next-line camelcase
```
mkholjuraev, line 46: Super nitpick: we do not need this eslint comment ^

mkholjuraev (commented):
While testing with sufficient permissions, I see the 'No groups' option in groups filter dropdown in both master and the PR. Maybe, this is out of the PR scope, but I believe if a user has no perms, then this option does not make much sense.

AsToNlele:
@mkholjuraev I agree, I'd hide the filter completely but it was decided against it

mkholjuraev (approved):
LGTM! @AsToNlele thank you for all the work!

PR #2175, opened 2024-04-15
fix: Prop types, unused props, replaced deprecated Main component

AsToNlele:
While working on inventory I was annoyed by the amount of prop type/other errors. Fixed some of them.

Some components don't use certain props so I excluded them from the spread.

`Main` component from `frontend-components` was deprecated for a while.

How to test:
Make sure the green section is the same like on stage
image

### DIFF
--- a/src/components/InventoryDetail/ApplicationDetails.js
+++ b/src/components/InventoryDetail/ApplicationDetails.js
@@ -13,6 +13,7 @@ import {
   REPORTER_RHSM_CONDUIT,
   REPORTER_RHSM_PROFILE_BRIDGE,
 } from '../../Utilities/constants';
+import { PageSection } from '@patternfly/react-core';
 
 /**
  * Component that renders tabs for each application detail and handles clicking on each item.
@@ -120,7 +121,7 @@ const ApplicationDetails = ({
                 >
                   {item.name === currentApp && (
                     <Suspense fallback={Spinner}>
-                      <section className="pf-v5-c-page__main-section">
+                      <PageSection>
                         {isEmptyState(currentApp) ? (
                           <NotConnected />
                         ) : (
@@ -130,7 +131,7 @@ const ApplicationDetails = ({
                             {...item}
                           />
                         )}
-                      </section>
+                      </PageSection>
                     </Suspense>
                   )}
                 </TabContent>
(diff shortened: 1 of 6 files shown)


src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -96,15 +96,26 @@ const ApplicationDetails = ({
...
-                />
-              ))}
+              {activeTabs?.map(
+                (
+                  {
+                    systemProfilePrefetched,
+                    component,
+                    nonEdge,
+                    tabRef,
+                    ...item
```
gkarat, line 106: What's the goal of spreading the keys that are not used? For example, you explicitly extract `systemProfilePrefetched` or `component` key, but don't utilise it anywhere further down the code.

AsToNlele: Nvm just removed the item spread, the props weren't used

src/routes/InventoryTable.js
```diff
@@ -48,7 +47,7 @@ const Inventory = (props) => {
       <PageHeader className="pf-m-light">
         <PageHeaderTitle title="Systems" />
       </PageHeader>
-      <Main>
+      <section className="pf-v5-l-page__main-section pf-v5-c-page__main-section">
```
bastilian, line 50: Should/Could we maybe use the `PageSection` component here?

https://www.patternfly.org/components/page#pagesection

AsToNlele: yup, replaced it on more places

AsToNlele:
Not sure that this error is fixable from Inventory, the `Tabs` component wants these props but the `button` at the bottom of the three doesn't.
image

src/components/InventoryDetail/ApplicationDetails.js
```diff
@@ -13,6 +13,7 @@ import {
   REPORTER_RHSM_CONDUIT,
   REPORTER_RHSM_PROFILE_BRIDGE,
 } from '../../Utilities/constants';
+import { PageSection } from '@patternfly/react-core/dist/esm';
```
gkarat, line 16: nit-pick:
```suggestion
import { PageSection } from '@patternfly/react-core';
```

gkarat (approved):
LGTM!

PR #2176, opened 2024-04-15
feat(RHINENG-9249): create bifrost toggle

mkholjuraev:
Resolves: https://issues.redhat.com/browse/RHINENG-9249

This creates a new toggle to switch between bootc table view and hybrid inventory tabs. This toggle should only be shown to accounts that has bootc images. 

To test:
1. Run the PR
2. Open the app with an account that has bootc images (ping me for an account with bootc)
3. Observe the new toggle is present and by default hybrid tabs are shown
4. Click on the bootc toggle on the right
5. Observe that you see a mock bootc table
6. Open the app with another account that does not have any bootc image
7. Observe that the toggle is not shown


computercamplove:
@mkholjuraev 
1. for what this feature flag **hbi.ui.bifrost** - when it was decided to use? 
2. since Systems page require filtering and this filtering works only with  "hbi.api.xjoin-disable" is enabled for specific account - account without this feature flag gets 500 and page is not loading, i don't see check for this flag
`	https://stage.foo.redhat.com:1337/api/inventory/v1/hosts?filter[system_profile][bootc_status][booted][image_digest][is]=not_nil&per_page=1`
image

3. toggle works as expected in account that has image-mode systems and feature flag "hbi.api.xjoin-disable" is enabled

mkholjuraev:
@computercamplove to your questions:
1. The frontend has an agreement to put a feature flag in every new feature
2. I did not know since yesterday evening's meeting about this backend flag. As it is backend-related, should it be used in frontend???
3. It was said yesterday that this flag was not in prod. I thought the backend getting rid of it, do I understand correctly that

computercamplove:
> @computercamplove to your questions:
> 
>     1. The frontend has an agreement to put a feature flag in every new feature
> 
>     2. I did not know since yesterday evening's meeting about this backend flag. As it is backend-related, should it be used in frontend???
> 
>     3. It was said yesterday that this flag was not in prod. I thought the backend getting rid of it, do I understand correctly that

1. thank you for explanation, i was asking because Bifrost team decided that they don't need UI flag check - user don't have image mode systems so they can't see all UI changes related to Bifrost (for context https://redhat-internal.slack.com/archives/C06EDRT4KB6/p1712666815175369). But it's fine to have - in this case we will have enabled in both environments  to keep automated tests running. If there will any issue we will disable it in production via your flag. 
2. i guess so, without this flag check filter url is not working on all accounts as you said yesterday (and as Dan explained this filter doesn't work with xjoin and works only when Xjoin is disabled - Xjoin disabled only for specific account, not on whole stage/prod, HBI team still works on migration), if you want we can clarify it again via thread with Chris/Dan.
3. "hbi.api.xjoin-disabled" is on prod (i just need to add account number to it so we can test it later)

computercamplove:
here clarification
Filtering image-mode systems without Xjoin is expected to be ready by Summit
for context
https://redhat-internal.slack.com/archives/CQFKM031T/p1713273872452199

johnsonm325:
@mkholjuraev We'll need to handle errors. The api for checking for bootc returned a 500 and the page just loads and never errors
image

mkholjuraev:
@johnsonm325 it seems this continuous loading is because of this [handler](https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/Routes.js#L75) that only considers 403 errors. I think this was to solve some issues. @gkarat I am wondering if we can generalise that handler to other error types as well?

gkarat:
@mkholjuraev, yeah, the 403 handler was introduced here https://github.com/RedHatInsights/insights-inventory-frontend/pull/2138, you can also find some details why it was done. You can for sure extend this block and treat other codes if it's necessary.

src/routes/InventoryPage.js
```diff
@@ -0,0 +1,45 @@
...
+  );
+};
+
+Inventory.defaultProps = {
+  initialLoading: true,
+  notificationProp: PropTypes.object,
+};
+Inventory.propTypes = {
+  isImmutableTabOpen: PropTypes.bool,
+};
```
johnsonm325, line 44: Can these be removed?

src/routes/InventoryComponents/InventoryPageHeader.js
```diff
@@ -0,0 +1,51 @@
...
+      aria-label="Bifrost"
+      isSelected={mainContent === pageContents.bifrost.key}
+      onChange={() => changeMainContent(pageContents.bifrost.key)}
+    />
+  </ToggleGroup>
+);
+
+const InventoryPageHeader = (toggleProps) => {
+  const isBifrostEnabled = useFeatureFlag('hbi.api.disable-xjoin');
+  const { hasBootcImages } = useContext(AccountStatContext);
```
johnsonm325, line 33: Is it worth it to provide `isBifrostEnabled` via the AccountStatContext as well?

johnsonm325 (approved):
I think this looks good enough to approve.

Just fix conflicts and address the small things I've pointed out.

src/routes/InventoryPage.js
mkholjuraev: Good call, will remove them

src/routes/InventoryComponents/InventoryPageHeader.js
mkholjuraev: The context is used to hold information about an account. I believe the flag would not be a proper thing to add.

PR #2177, opened 2024-04-15
feat(runtimes): add Runtimes Processes card to GeneralInformation

aptmac:
Hi! This PR adds a card to the General Information on the Inventory Systems page to display information from runtimes-inventory.

This "Application Services Proccesses" card resides in the [insights-runtimes-frontend](https://github.com/RedHatInsights/insights-runtimes-frontend) repository, and at the moment is only deployed in Stage (but will be looking to get into prod soon). The runtimes-inventory API returns application processes data given a hostname, so the card will need to be passed the `fqdn` from General Information. If no information exists or nothing is returned from the runtimes-inventory endpoint, the card will not display.

As for testing this functionality, there are two ways that I use at the moment:

1. Webpack custom proxy + locally hosted mock data

The insights-runtimes-frontend repo contains a /mock folder that holds a json file containing a test response from insights-runtimes-inventory that can be tested against, see: https://github.com/RedHatInsights/insights-runtimes-frontend/blob/master/mock/instances.json

You can run a mock json-server in insights-runtimes-frontend by running: `npm run mock-instances`, and this json will be served at localhost:3000.

Over in insights-inventory-frontend, you can add some extra custom proxy to the dev webpack (I'll add a diff below) to make it use that json-server for information, and run `npm run mock-server` && `npm run start:mock`.

```
diff --git a/config/dev.webpack.config.js b/config/dev.webpack.config.js
index 0add63c..75ea99d 100644
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -26,8 +26,8 @@ const { config: webpackConfig, plugins } = config({
   ...(process.env.MOCK && {
     customProxy: [
       {
-        context: ['/api/inventory/v1/groups'], // you can adjust the `context` value to redirect only specific endpoints
-        target: 'http://localhost:4010', // default prism port
+        context: ['/api/inventory/v1/groups', '/api/inventory/v1/hosts'], // you can adjust the `context` value to redirect only specific endpoints
+        target: 'http://127.0.0.1:4010', // default prism port
         secure: false,
         changeOrigin: true,
         pathRewrite: { '^/api/inventory/v1': '' },
@@ -35,6 +35,15 @@ const { config: webpackConfig, plugins } = config({
           proxyReq.setHeader('x-rh-identity', 'foobar'); // avoid 401 errors by providing neccessary security header
         },
       },
+      {
+        context: ['/api/runtimes-inventory-service/v1/instances'],
+        target: 'http://localhost:3000',
+        secure: false,
+        changeOrigin: true,
+        pathRewrite: {
+          '^/api/runtimes-inventory-service/v1/instances': 'instances',
+        },
+      },
     ],
   }),
 });
```

2. Uploading mock host + reports to stage

In insights-runtimes-inventory, I've added a folder containing a couple of scripts that can be used to create mock reports for sending to Ephemeral or Stage environments for testing the frontend, see: https://gitlab.cee.redhat.com/insights-runtimes/insights-runtimes-inventory/-/blob/main/mock/mock.md?ref_type=heads

Once these reports are in Stage, you can view them with a proxied insights-inventory-frontend with `npm run start:proxy`.

Example UI:
inventory-frontend-pr-2

### DIFF
--- a/src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
+++ b/src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
@@ -194,7 +194,7 @@ class GeneralInformation extends Component {
                   </GridItem>
                 )}
 
-                {this.props.showRuntimesProcesses && (
+                {this.props.showRuntimesProcesses && entity.fqdn && (
                   <GridItem>
                     <AsyncComponent
                       appName="runtimes"
--- a/src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
+++ b/src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
@@ -193,6 +193,16 @@ class GeneralInformation extends Component {
                     />
                   </GridItem>
                 )}
+
+                {this.props.showRuntimesProcesses && (
+                  <GridItem>
+                    <AsyncComponent
+                      appName="runtimes"
+                      module="./RuntimesProcessesCard"
+                      hostname={entity.fqdn}
+                    />
+                  </GridItem>
+                )}
               </Grid>
             </GridItem>
             {children}
@@ -227,6 +237,7 @@ GeneralInformation.propTypes = {
         name: PropTypes.string,
       }),
     }),
+    fqdn: PropTypes.string,
   }),
   openedModal: PropTypes.string,
   loadSystemDetail: PropTypes.func,
@@ -271,6 +282,7 @@ GeneralInformation.propTypes = {
   systemProfilePrefetched: PropTypes.bool,
   showImageDetails: PropTypes.bool,
   isBootcHost: PropTypes.bool,
+  showRuntimesProcesses: PropTypes.bool,
 };
 GeneralInformation.defaultProps = {
   entity: {},
@@ -285,6 +297,7 @@ GeneralInformation.defaultProps = {
   CollectionCardWrapper: false,
   systemProfilePrefetched: false,
   showImageDetails: false,
+  showRuntimesProcesses: false,
 };
 
 const GeneralInformationComponent = (props) => {
--- a/src/components/SystemDetails/GeneralInfo.js
+++ b/src/components/SystemDetails/GeneralInfo.js
@@ -16,6 +16,9 @@ const GeneralInfoTab = (props) => {
   const enableEdgeInventoryListDetails = useFeatureFlag(
     'edgeParity.inventory-list'
   );
+  const enableRuntimesInventoryCard = useFeatureFlag(
+    'runtimes.inventory-card.enabled'
+  );
 
   return (
     <GeneralInformation
@@ -24,6 +27,7 @@ const GeneralInfoTab = (props) => {
       showImageDetails={
         enableEdgeImageDetails && enableEdgeInventoryListDetails && isEdgeHost
       }
+      showRuntimesProcesses={enableRuntimesInventoryCard}
     />
   );
 };


src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -193,6 +193,16 @@ class GeneralInformation extends Component {
                     />
                   </GridItem>
                 )}
+
+                {this.props.showRuntimesProcesses && (
```
mkholjuraev, line 197: Do all hosts have fqdn associated? and if not should we check `fqdn` presence as well not to load the module when it is unnecessary?

mkholjuraev (commented):
Codewise LGTM! I have tested the PR locally by running both apps and mocked server. I get the 'Created' as invalid, but I think this is due to mocked data. If this is indeed related to the mocked data, not the implementation, we can proceed with the PR.

image

aptmac:
> Codewise LGTM! I have tested the PR locally by running both apps and mocked server. I get the 'Created' as invalid, but I think this is due to mocked data. If this is indeed related to the mocked data, not the implementation, we can proceed with the PR.
> 
> image

Thanks for the review!

My first guess is that something went wrong when formatting the timestamp using `Date(instance.created).toLocaleString()` at: https://github.com/RedHatInsights/insights-runtimes-frontend/blob/master/src/utils/utils.ts#L30.

But it also could have been the way the mock data generated the timestamp in the `generate_mock_reports.py` script at https://gitlab.cee.redhat.com/insights-runtimes/insights-runtimes-inventory/-/blob/main/mock/generate_mock_reports.py?ref_type=heads#L23. 

I'll take a look at this tomorrow morning. Would there happen to be an error in the browser dev console? That might give me a better indication if it was typescript side or the mock data generation side.

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
aptmac: Good idea, taking a quick look at the open api spec it looks like fqdn is not a required parameter in host inventory so a check sounds like a good safeguard.

computercamplove:
@aptmac could you please provide jira-card, i didn't know about this change and no automation test is prepared for it for now.

aptmac:
> I get the 'Created' as invalid, but I think this is due to mocked data. If this is indeed related to the mocked data, not the implementation, we can proceed with the PR.

@mkholjuraev I think I might have found out why it came back as Invalid, would you be able to check something for me in the dev console of whatever browser you're using? 

Could you be able to check what is returned by:

`new Date("2024-02-21T20:14:29.586839Z").toLocaleString()`

and if that comes back with 'Invalid Date', could you try:

`new Date("2024-02-21T20:14:29.586Z").toLocaleString()`

When parsing the date string the JavaScript `new Date()` is likely coming back as NaN, so the locale string is 'Invalid Date'. Looking at the Date documentation ([link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format)) it looks like the format it's expecting is `YYYY-MM-DDTHH:mm:ss.sssZ`, but the default return from our Java backend with the type ZonedDateTime is `YYYY-MM-DDTHH:mm:ss.ssssssZ` with 3 extra millisecond digits (as seen with the test data), I'm wondering if that's what is breaking here.

mkholjuraev:
Here is the output that you requested, let me know if anything else needed: 

image

aptmac:
The runtimes-frontend has included a change to better handle the incoming date strings, and is deployed in Stage. I also rebased this PR on-top of master to include the latest upstream changes.

mkholjuraev (approved):
LGTM! @aptmac thank you! Let me know if anything else is needed.

aptmac:
@mkholjuraev Hey! If you get the time to take a look at merging this it would be greatly appreciated. I rebased on-top of the latest changes and the Travis build looks to be okay, ci-int not so much but that looks to be down across the board here.

mkholjuraev:
@aptmac sorry, I missed you ping. Merging

PR #2178, opened 2024-04-22
feat(RHINENG-9248): create bootc tree view

johnsonm325:
To test:
- Log into account with systems with bootc data
- Go to /insights/inventory url
- Click the toggle that is right-aligned in the Systems page header

From this point, you should see a table with expandable rows. Those expanded rows include specific hash data for each image name. The system counts  for the hash commits for the nested table should add up to equal the system counts for the parent image.

Please also ensure that the numbers in the system column line up for the table and nested table.

### DIFF
--- a/src/Utilities/constants.js
+++ b/src/Utilities/constants.js
@@ -1,9 +1,4 @@
-import React from 'react';
 import { createContext } from 'react';
-import {
-  Skeleton,
-  SkeletonSize,
-} from '@redhat-cloud-services/frontend-components/Skeleton';
 
 export const TEXT_FILTER = 'hostname_or_id';
 export const TEXTUAL_CHIP = 'textual';
@@ -28,10 +23,9 @@ export const INVENTORY_TOTAL_FETCH_URL_SERVER = '/api/inventory/v1/hosts';
 export const INVENTORY_TOTAL_FETCH_EDGE_PARAMS =
   '?filter[system_profile][host_type]=edge&page=1&per_page=1';
 export const INVENTORY_TOTAL_FETCH_CONVENTIONAL_PARAMS = '?page=1&per_page=1';
-export const INVENTORY_TOTAL_FETCH_BIFROST_PARAMS =
-  '?filter[system_profile][bootc_status][booted][image_digest][is]=not_nil&per_page=1';
 export const INVENTORY_FETCH_BIFROST_PARAMS =
   '?filter[system_profile][bootc_status][booted][image_digest][is]=not_nil';
+export const INVENTORY_TOTAL_FETCH_BIFROST_PARAMS = `${INVENTORY_FETCH_BIFROST_PARAMS}&per_page=1`;
 export function subtractDate(days) {
   const date = new Date();
   date.setDate(date.getDate() - days);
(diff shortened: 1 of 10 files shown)


src/Utilities/constants.js
```diff
@@ -25,6 +30,8 @@ export const INVENTORY_TOTAL_FETCH_EDGE_PARAMS =
 export const INVENTORY_TOTAL_FETCH_CONVENTIONAL_PARAMS = '?page=1&per_page=1';
 export const INVENTORY_TOTAL_FETCH_BIFROST_PARAMS =
   '?filter[system_profile][bootc_status][booted][image_digest][is]=not_nil&per_page=1';
+export const INVENTORY_FETCH_BIFROST_PARAMS =
```
mkholjuraev, line 33: If you feel up to small improvements, we can combine the above 2 constants together. This would be useful in case backend folks change the filter syntax.

src/Utilities/constants.js
```diff
@@ -268,3 +275,36 @@ export const hybridInventoryTabKeys = {
     url: '/manage-edge-inventory',
   },
 };
+
+export const LOADING_BIFROST_TABLE = [
```
mkholjuraev, line 279: This works just fine. However, I think we can also use the SkeletonTable from fec-components package as it was used here: https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/src/components/InventoryTable/EntityTable.js#L11C32-L11C88.

src/routes/InventoryComponents/BifrostTable.js
```diff
@@ -1,7 +1,88 @@
...
+      );
+      return isExpanding
+        ? [...otherExpandedImageNames, image.image]
+        : otherExpandedImageNames;
+    });
+
+  const isImageExpanded = (image) => expandedImageNames.includes(image.image);
+
+  return (
+    <Table aria-label="Simple table">
```
mkholjuraev, line 32: nitpick: let's use more meaningful aria-label.

mkholjuraev (commented):
Looks nice! One thing I noticed, should we also implement the filter and the pagination as in the [mockups](https://www.figma.com/file/izOiimeI3i1a44dKEaFsUf/Systems-in-Bifrost?type=design&node-id=579-30873&mode=design&t=gtEUIQFec7qzQuhd-0)?

beav:
> Looks nice! One thing I noticed, should we also implement the filter and the pagination as in the [mockups](https://www.figma.com/file/izOiimeI3i1a44dKEaFsUf/Systems-in-Bifrost?type=design&node-id=579-30873&mode=design&t=gtEUIQFec7qzQuhd-0)?

IMO I'd only add this in if it was easy. If it isn't, we can put in an RFE to add it.

mkholjuraev:
> IMO I'd only add this in if it was easy. If it isn't, we can put in an RFE to add it.

In case, the number of bootc images is not huge, we can add the filter and pagination later, for sure. Let's go ahead and approach the work as you suggest then!

mkholjuraev (approved):
LGTM! Just a few comments that can be useful.

mkholjuraev (approved):
I went through the code and re-tested again, everything seems fine.

src/routes/InventoryComponents/BifrostPage.js
```diff
@@ -0,0 +1,77 @@
...
+      ];
+
+      setLoaded(true);
+      setBootcImages(updated);
+    };
+
+    fetchBootcImages();
+  }, []);
+
+  return <BifrostTable bootcImages={bootcImages} loaded={loaded} />;
```
bastilian, line 74: The `loaded` could be used here and not passed down, and either show the `BifrostTable` or the `SkeletonTable` and make the BifrostTable cleaner.

PR #2180, opened 2024-04-24
feat(RHINENG-9686): Add system table toggle details

johnsonm325:
Add "View by" text to the left of system table toggle on the Systems table page header. Each toggle button should have tooltips.


mkholjuraev (approved):
Works as expected! The only thing I noticed is that after these changes toggle buttons have unexpected border radius. It is not visible unless you look closely. It would maybe better to fix this before merging!

image

PR #2182, opened 2024-04-25
feat(RHINENG-9709): Add available image to bootc card

dkuc:
Adds available image and available image hash.
image

### DIFF
--- a/src/components/GeneralInfo/BootcImageCard/BootcImageCard.js
+++ b/src/components/GeneralInfo/BootcImageCard/BootcImageCard.js
@@ -22,6 +22,13 @@ export const BootcImageCard = ({ handleClick, extra }) => {
         ...[{ title: 'Booted Image Digest', value: bootc.bootedImageDigest }],
         ...[{ title: 'Staged Image', value: bootc.stagedImage }],
         ...[{ title: 'Staged Image Digest', value: bootc.stagedImageDigest }],
+        ...[{ title: 'Available Image', value: bootc.availableImage }],
+        ...[
+          {
+            title: 'Available Image Digest',
+            value: bootc.availableImageDigest,
+          },
+        ],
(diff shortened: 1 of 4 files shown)


bastilian:
@dkuc Can we please add a test for this component? I would recommend the `BiosCard` test as a base, as it is the most basic.

src/components/GeneralInfo/BootcImageCard/BootcImageCard.js
```diff
@@ -22,6 +22,13 @@ export const BootcImageCard = ({ handleClick, extra }) => {
         ...[{ title: 'Booted Image Digest', value: bootc.bootedImageDigest }],
         ...[{ title: 'Staged Image', value: bootc.stagedImage }],
         ...[{ title: 'Staged Image Digest', value: bootc.stagedImageDigest }],
+        ...[{ title: 'Available Image', value: bootc.availableImage }],
+        ...[
```
bastilian, line 26: It is not necessary to wrap each item in an array and spread it into the array, with `...[ ]`. Only the result of `extra.map()` needs to be spread, since map returns an array.

dkuc: Agreed, just following the conventions and style of the other code in the repo.

dkuc:
> @dkuc Can we please add a test for this component? I would recommend the `BiosCard` test as a base, as it is the most basic.

Done.

src/components/GeneralInfo/BootcImageCard/BootcImageCard.js
bastilian: @dkuc Though, never hesitate to break conventions if they aren't right. :)

dkuc:
Just fyi, I don't have merge permissions.

PR #2183, opened 2024-04-25
fix(RHINENG-9248): fetch non-edge hosts

johnsonm325:
For the new bootc image table, we request the non-image based system count to include at the bottom of the table. This PR modifies the api call to request to filter by non-edge based systems so we get a count of systems that matches the system table.


dkuc (approved):

mkholjuraev (approved):
LGTM!

PR #2184, opened 2024-04-25
feat(RHINENG-9685): image/package mode icons to inv table

johnsonm325:
This PR adds Image and Package mode icons to the Inventory table. We use the api to request the `bootc_status` from `system_profile` and if `bootc_status` exists, we show the image mode icon. If it doesn't exist, we show the package mode icon.

Hovering over each will provide a popover with a description of the image and package modes, respectively.


mkholjuraev:
Icons look good. I have a question, should those icons be displayed to accounts that have no bootc images? There was a requirement that no bootc table should be shown to those accounts.

computercamplove:
fyi ignore my retest command,  PR check is not working

mkholjuraev (approved):
LGTM!

PR #2185, opened 2024-04-26
fix(RHINENG-9686): border-radius in toggle group

johnsonm325:
This PR fixes the issue with the rounded edges of the ToggleGroupItems when a Tooltip is wrapped around the items. Adding the `triggerRef` removes the `<div>` that breaks the component.

### DIFF
--- a/src/routes/InventoryComponents/InventoryPageHeader.js
+++ b/src/routes/InventoryComponents/InventoryPageHeader.js
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useRef } from 'react';
 import PropTypes from 'prop-types';
 import {
   Split,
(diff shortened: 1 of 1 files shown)


mkholjuraev (approved):
This does fix the issues with the border radius. But for my understanding can you tell me how? What does `triggerRef={() => document.getElementById('view-by-systems-toggle')}` do?

bastilian:
+1 to what @mkholjuraev is asking. I also wonder if this could, and maybe even should use a proper React ref.

johnsonm325:
@mkholjuraev @bastilian This is part of the PF5 upgrade. without the `triggerRef`, there is an extra `<div>` between the `Tooltip` and child layer. Adding the `triggerRef` removes that `<div>` layer, allowing the styles to work properly.

This is the PF way to fix this issue.

bastilian:
@johnsonm325 I think this should use the `useRef` and the `ref` prop and not use `document.getElementById`. with useRef React will maintain the reference, using `document.getElementById` might return an instance of the element that is already outdated.

johnsonm325:
@mkholjuraev @bastilian you can take another look.

mkholjuraev (approved):
Now, it looks even better. I retested and the border radius is fine.

PR #2194, opened 2024-05-09
update components to pf5

adonispuente:


fec.config.js
```diff
@@ -1,13 +1,122 @@
...
+      {
+        'react-router-dom': {
+          singleton: true,
+          import: false,
+          version: packageJson.dependencies['react-router-dom'],
+          requiredVersion: '>=6.0.0 <7.0.0',
+        },
+      },
+    ],
+    root: resolve(__dirname, './'),
```
bastilian, line 30: Is this really needed?

fec.config.js
```diff
@@ -1,13 +1,122 @@
...
-  debug: true,
-  useProxy: true,
-  proxyVerbose: true,
-  /**
-   * Change to false after your app is registered in configuration files
-   */
+  appName,
+  appUrl: `/${bundle}/${appName}`,
+  useProxy: process.env.PROXY === 'true',
   interceptChromeConfig: false,
```
bastilian, line 12: This is not needed and might actually cause issues with prod.

bastilian:
@adonispuente We should add `.cache` to the `.gitignore` and not commit the contents of this directory.

package.json
```diff
@@ -157,28 +158,28 @@
...
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelint.json",
-    "prod": "NODE_ENV=production webpack serve --config config/dev.webpack.config.js",
     "server:ctr": "node src/server/generateServerKey.js",
-    "start": "NODE_ENV=development webpack serve --config config/dev.webpack.config.js",
-    "start:proxy": "PROXY=true NODE_ENV=development webpack serve --config config/dev.webpack.config.js",
-    "start:proxy:beta": "BETA=true PROXY=true NODE_ENV=development webpack serve --config config/dev.webpack.config.js",
-    "travis:build": "NODE_ENV=production webpack --config config/prod.webpack.config.js",
+    "start": "fec dev",
+    "start:proxy": "PROXY=true fec dev",
+    "travis:build": "NODE_ENV=production webpack --config fec.webpack.config.js",
```
bastilian, line 177: We can remove this task too. and just replace `travis:build` with just `build` in `travis:verify`

package.json
```diff
@@ -187,7 +188,9 @@
     "start:mock": "MOCK=true npm run start:proxy",
     "start:mock:beta": "MOCK=true npm run start:proxy:beta",
     "prepare": "husky install",
-    "commitlint": "commitlint --from HEAD~1 --to HEAD --verbose"
+    "commitlint": "commitlint --from HEAD~1 --to HEAD --verbose",
+    "postinstall": "ts-patch install",
+    "patch-hosts": "fec patch-etc-hosts"
```
bastilian, line 193: I don't think we need to add this. If we would we would want to add it to all apps, and we probably don't want that. 

If someone wants to run the patch hosts script they can do it via `npx fec patch-etc-hosts` as well.

pr_check.sh
```diff
@@ -10,7 +10,7 @@ export WORKSPACE=${WORKSPACE:-$APP_ROOT} # if running in jenkins, use the build'
 export IMAGE_TAG=$(git rev-parse --short=7 HEAD)
 export GIT_COMMIT=$(git rev-parse HEAD)
 export APP_ROOT=$(pwd)
-export NODE_BUILD_VERSION=15
+export NODE_BUILD_VERSION=18
```
bastilian, line 13: We should also bump the version for travis while we are at it. Do watch out bumping node on travis will also require bumping the distro. Take a look at the compliance travis config.

src/components/InventoryGroups/Modals/ConfirmSystemsAddModal.js
```diff
@@ -1,7 +1,7 @@
 import { FormSpy, useFormApi } from '@data-driven-forms/react-form-renderer';
 import { Button, Flex, Icon } from '@patternfly/react-core';
 import ExclamationTriangleIcon from '@patternfly/react-icons/dist/js/icons/exclamation-triangle-icon';
-import warningColor from '@patternfly/react-tokens/dist/esm/global_warning_color_100';
+import warningColor from '@patternfly/react-tokens';
```
bastilian, line 4: This should be `import { warningColor } from '@patternfly/react-tokens';`

package.json
adonispuente: are you saying to replace with `travis:verify : travis build` or fec:build?

package.json
```diff
@@ -157,28 +158,28 @@
...
-    "travis:build": "NODE_ENV=production webpack --config config/prod.webpack.config.js",
+    "start": "fec dev",
+    "start:proxy": "PROXY=true fec dev",
+    "travis:build": "NODE_ENV=production webpack --config fec.webpack.config.js",
     "travis:verify": "npm-run-all travis:build lint test",
     "verify": "npm-run-all build lint test",
     "verify:local": "npm-run-all build lint test:local test:ct",
     "nightly": "npm run travis:verify",
-    "start:federated": "fec static --config config/dev.webpack.config.js",
+    "start:federated": "fec static --config fec.webpack.config.js",
```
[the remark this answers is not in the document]
adonispuente: ahh yeah, this was when I was trouble shooting and forgot to remove

fec.config.js
adonispuente: its been there for quite some time, I kept it simply because I wasnt confident it wouldnt affect another app

package.json
bastilian: > replace `travis:build` with just `build` in `travis:verify`

Below the `travis:verify` calls `travis:build`. This can be replaced with just `build` and the `travis:build` task can be removed.

pr_check.sh
adonispuente: theres a difference in script sections and the dist key/value but overall the only other difference is the string/int in node version but travis is smart enough to read that. What else would you think should mimic compliance? currently inventory travis has a lot of if statements in there. We could remove that overall

bastilian: ... I am not sure what you mean... :D These three lines from the compliance travis config should be in the `.travis.yml` here to get node 18: https://github.com/RedHatInsights/compliance-frontend/blob/master/.travis.yml#L5-L7

adonispuente: Seems my eyes were broken, added lol

.travis.yml
```diff
@@ -1,8 +1,9 @@
 language: node_js
 notifications:
   email: false
+dist: focal
```
gkarat, line 4: Is this change really needed? I can see other apps migrated to fec without changing this option like patchman https://github.com/RedHatInsights/patchman-ui/blob/master/.travis.yml

fec.config.js
```diff
@@ -1,13 +1,121 @@
...
-  /**
-   * Change to false after your app is registered in configuration files
-   */
-  interceptChromeConfig: false,
-  /**
-   * Add additional webpack plugins
-   */
+  plugins: [
+    new webpack.DefinePlugin({
+      IS_DEV: false,
```
gkarat, line 15: I would set this boolean value to `process.env.NODE_ENV === 'development'`. Because otherwise with your change there will be never IS_DEV=true when working for example locally on stage where we would need redux logger and some other enhancers enabled in dev environment.

src/components/InventoryHostStaleness/BaseDropDown.js
gkarat, line 1: These changes seem irrelevant to the PR goal to update fec config. Why we need it here?

fec.config.js
bastilian: +1 However, I think in the future we should replace this. It's used to add or not add the `redux-logger`, We can do that differently, similar to how we do it elsewhere.

.travis.yml
bastilian: It is needed for node `v18` otherwise it won't be able to install this version due to missing glibc libraries.

fec.config.js
bastilian: I'd try to remove it, so we keep the config at really the bare minimum.

src/components/InventoryHostStaleness/BaseDropDown.js
adonispuente: without these changes, the styling in this section is improper. the icons become black.

src/components/ImmutableDevices/helpers.js
```diff
@@ -12,11 +12,13 @@ import {
...
-import successColor from '@patternfly/react-tokens/dist/esm/global_success_color_100';
-import infoColor from '@patternfly/react-tokens/dist/esm/global_info_color_100';
-import activeColor from '@patternfly/react-tokens/dist/esm/global_active_color_100';
+import {
+  dangerColor,
+  successColor,
+  warningColor,
+  infoColor,
+  activeColor,
+} from '@patternfly/react-tokens';
```
bastilian, line 21: It seems that webpack now complain about these. I think the imports are named like the filenames before, like `global_success_color_100` and `global_warning_color_100`.

fec.config.js
bastilian: While testing the PR I tried running it with `root` and it appeared to work fine.

package.json
```diff
@@ -187,8 +186,9 @@
     "start:mock": "MOCK=true npm run start:proxy",
```
bastilian, line 186: I think we need to copy the part for the mock routes[0] over as well, to not break this here. I'm not sure if this is still in use by someone, but let's be safe. I thin the cypress tests may use this.

[0] https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/config/dev.webpack.config.js#L26-L39

src/components/ImmutableDevices/helpers.js
adonispuente: If I dont use this current path, the only other path I can find is the old `dist/esm/colorNameHere`

bastilian: no, no. what I mean is that the import names, like `activeColor` or `successColor` have to be something like `global_success_color_100`. the `from '@patternfly/react-tokens'` is correct.

If you look at the Terminal when webpack compiled, you can see a wall of possible exports from '@patternfly/react-tokens'.

src/components/InventoryTable/EntityTable.js
```diff
@@ -2,7 +2,6 @@ import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { selectEntity, setSort } from '../../store/actions';
 import { useDispatch, useSelector } from 'react-redux';
-import { TableGridBreakpoint, TableVariant } from '@patternfly/react-table';
```
bastilian, line 5: There should be a least the `TableVariant` exported from the `/deprecated` module, maybe even the `TableGridBreakpoint`. If it is not found we may need to add an entry to the transforms.

adonispuente: Seems wed have to go the transfromer route. The only possible exports from this are Table, TableBody, TableHeader

adonispuente: Though I dont think this is causing any issues now. These are essentially just strings

fec.config.js
bastilian: Oh, I just saw that I mean "without". 🙈

Meaning we should remove this `root` prop.

src/components/InventoryHostStaleness/__test__/InventoryHostStaleness.test.js
```diff
@@ -47,12 +47,6 @@ describe('Table Renders', () => {
 
     await userEvent.click(screen.getByRole('button', { name: 'Edit' }));
 
-    expect(
-      screen.getAllByRole('button', { name: /options menu/i })
-    ).toHaveLength(3);
-    screen
-      .getAllByRole('button', { name: /options menu/i })
-      .forEach((button) => expect(button).toBeEnabled());
```
bastilian, line 55: Why was this needed to be removed?

src/components/InventoryTable/EntityTable.js
bastilian: We should change these back and add the proper entries to the fec config.

bastilian: Actually. They don't need any entry `import { TableGridBreakpoint, TableVariant } from '@patternfly/react-table';` works just fine. 

I'm curious why these had to change in the first place? :)

src/components/GroupsTable/GroupsTable.cy.js
```diff
@@ -295,7 +295,7 @@ describe('selection and bulk selection', () => {
 
   it('can select none', () => {
     selectRowN(1);
-    cy.get(PT_BULK_SELECT).contains('1 selected').click(); // open selection dropdown
+    cy.get(PT_BULK_SELECT).contains('1 selected').click().click();
```
bastilian, line 298: ... `.click().click();`? ... when the battery on my TV remote is almost empty I tend to push harder and harder too.... but the better solution is always to change the batteries...

src/components/GroupSystems/GroupSystems.cy.js
```diff
@@ -309,7 +310,7 @@ describe('selection and bulk selection', () => {
   });
 
   it('can select and deselect systems', () => {
-    const middleRow = Math.ceil(DEFAULT_ROW_COUNT / 4);
+    const middleRow = Math.ceil(DEFAULT_ROW_COUNT / 7.5);
```
bastilian, line 313: red-sus-among-us_bGhsa2aUmZqaraWkpJRnZWltrWdlaW0-3351645577

src/components/InventoryHostStaleness/BaseDropDown.js
```diff
@@ -1,7 +1,14 @@
...
-import { Select, SelectOption } from '@patternfly/react-core/deprecated';
+import {
+  Button,
+  Flex,
+  FlexItem,
+  MenuToggle,
+  Popover,
+  Select,
+  SelectOption,
+} from '@patternfly/react-core';
```
bastilian, line 11: Let's try to revert this. While we do get a component NOT from `'@patternfly/react-core/deprecated'` it may not be the same and behave differently. 

Reverting to the old imports will probably also not require any changes below.

src/components/InventoryTable/EntityTable.js
adonispuente: Reverting, but we had talked about this on stand up and decided that there shouldnt be two different instances for imports on the table. See as their strings though I see no problem with this

src/components/GroupSystems/GroupSystems.cy.js
adonispuente: lol cypress was giving an error where if it couldnt actually see the item being selected on the viewport, it would fail. So i increased the value so the item selected would be higher up/cypress would always see it

src/components/GroupsTable/GroupsTable.cy.js
adonispuente: Truthfully I thought it looked nicer than the actual cypress syntax, sending it up now

src/components/InventoryHostStaleness/__test__/InventoryHostStaleness.test.js
adonispuente: This shouldnt have been sent up, the names arent there anymore but I just find the buttons a different way now.

src/components/InventoryHostStaleness/BaseDropDown.js
```diff
@@ -35,6 +42,34 @@ const BaseDropdown = ({
...
+  };
+
+  const toggle = (toggleRef) => {
+    const getNameByValue = (value) => {
+      for (const obj of dropdownItems) {
+        if (obj.value === value) {
+          return obj.name;
+        }
+      }
+    };
```
bastilian, line 56: `.filter` ;)

bastilian:
@adonispuente We also need to remove the transforms in the babel config here: https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/babel.config.js#L54-L105

gkarat:
@adonispuente, I think we've got one merge conflict after the recent push to master. Can you please take a look?

mkholjuraev (approved):
LGTM! Conflicts seem to be minor. Thus approving. If you feel another eye is needed, feel free to ping

aptmac:
Hi! I've been keeping tabs on this PR while working on something similar for the insights-runtimes-frontend repo. Just a couple of things I've come across:

1. I've noticed that the `fec.config.js` here is missing the custom proxy when running `npm run start:mock`, so the prism mock server isn't used.

2. I've also noticed (and something I'm looking into at the moment) is that trying to specify a port number isn't working when running a LOCAL_APPS setup. `npm run start -- --port=8004` for example doesn't take the `8004` into consideration, and for me at least it starts up on port `8002`.

adonispuente:
@aptmac I took a look around some other applications and it seems that this is the case for the majority. 
npm run start:mock isnt set up in the inventory repo itself, and the port issue is also in other apps as well. I think that has to be resolved in the package itself and then delivered to all the apps via package update or another means

aptmac:
> @aptmac I took a look around some other applications and it seems that this is the case for the majority. npm run start:mock isnt set up in the inventory repo itself, [..]

It used to be in the `dev.webpack.config.js` see: [here](https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/config/dev.webpack.config.js#L26), FWIW I've added it into the runtimes-frontend repo the same way it was in the old webpack and it's been working, see: https://github.com/RedHatInsights/insights-runtimes-frontend/pull/16/files#diff-d0f06d7cd724a6add60ee1bac8183344dc66faae9dac3ee6ee920f5b7ac3a88eR18

adonispuente:
@aptmac thank you! Im not against it, but now that I see the other converted apps are ran the same way im thinking maybe we wanted to remove the mock option for some reason, or if there was just going to be a short period before the new set up can handle it 'directly'.
Nice fix though, im gonna ask around and see if theres a reason, but if not then I can add that in, or at least look into how to bake it into the package

aptmac:
Was just snooping around the frontend-components repo and found https://github.com/RedHatInsights/frontend-components/pull/2007, looks like that might address the LOCAL_APPS in the future, just a heads up!

ezr-ondrej:
do I read it correctly image build for this have failed?

mkholjuraev:
> do I read it correctly image build for this have failed?

Yes, it seems the image is not built. We will have a look why

src/components/InventoryGroups/Modals/ConfirmSystemsAddModal.js
```diff
@@ -1,7 +1,7 @@
 import { FormSpy, useFormApi } from '@data-driven-forms/react-form-renderer';
 import { Button, Flex, Icon } from '@patternfly/react-core';
 import ExclamationTriangleIcon from '@patternfly/react-icons/dist/js/icons/exclamation-triangle-icon';
-import warningColor from '@patternfly/react-tokens/dist/esm/global_warning_color_100';
+import { warningColor } from '@patternfly/react-tokens';
```
bastilian, line 4: This should have been changed too, like the ones above. it is now raising an error when building: 
Screenshot 2024-06-26 at 13 42 54

PR #2196, opened 2024-05-14
fix(RHINENG-8550): Fix the last seen filter not being checked

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-8550.

This resolves one of the issues started happening after PF5 migration: the last seen filter values were not marked as checked when user selected any of the values.

It slightly changes how we build filter configuration for the Last seen filter: the value was not marked as selected, because PrimaryToolbar (from frontend-components) expects `value` to be string, whereas we were passing an object which should have broken selection.

## How to test

- At /inventory, make sure the filter works correctly: the selected option is checked, the table is updated and URL has a filter parameter. After refresh, the filter must be applied as well (to the exception of "custom" option).
- On any other page where InventoryTable is used, test if the filter works there as well.


AsToNlele (commented):
Tested on Inventory and works great! :smile: 

In Patch > Advisories the publish date filter has the "All" value not checked
Not sure if it's connected, it's working in Vulnerability
image

AsToNlele (approved):
LGTM, thanks for clarifying! :smile:

PR #2197, opened 2024-05-16
chore: Opt out from semantic-release and change commitlint behaviour

gkarat:
As we have not found any bigger benefit from using releases and updating CHANGELOG, I propose to dump the support of semantic-release-related packages. Also, commitlint is now executed only in non-PR runs only.


bastilian (approved):
LGTM! Thank you @gkarat!

PR #2204, opened 2024-05-29
fix(RHINENG-9504): Return missing system stale and warning values

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-9504.

After https://github.com/RedHatInsights/insights-inventory-frontend/pull/2189, we have removed some of the values that should have persisted. This PR is making sure we don't unnecessarily remove some values for staleness warning and deletion.


mkholjuraev (approved):
LGTM! I am able to save the changes and see the new options

PR #2205, opened 2024-05-30
fix(api): RHINENG-10393 - Prepend existing system_profile filters

bastilian:
This fixes the global filter on the Systems and the systems table Group Details page.


gkarat (approved):
LGTM! thanks @bastilian, let's wait for CI to pass and we can merge it

PR #2207, opened 2024-05-31
Enables the back the commit to implement system type filter and also changes how image and package modes differ 

mkholjuraev:
Resolves: https://issues.redhat.com/browse/RHINENG-10029

This PR is intended to change how image mode and package modes are checked in the code base to show the icons in the conventional systems table and also in the inventory detail page. Previously, the UI checked only system_profile.bootc_state. This resulted in incorrect results in the table. Now it uses system_profile.bootc_status.booted.image_digest.

The PR enables the system type filter. The commit was reverted when the mobile issues happened during the summit and we had to do an immediate release.


gkarat (approved):
LGTM 🥇  Thanks @mkholjuraev

gkarat:
If you could rebase these two commits, this would be awesome. This way we can make sure two distinct changes are in two separate commits. Thought the revert one has the naming that can make commitlint fail on master so I advise to change the commit message before rebasing

mkholjuraev:
@gkarat thank you for the review! I will make sure I rebase the commits instead of squashing

PR #2208, opened 2024-05-31
feat(RHINENG-358): Add subscription info to inv details

johnsonm325:
Description of problem: Unable to review what subscription information per host is uploaded to Insights

Steps to Reproduce:

1. Navigate to https://console.redhat.com/insights/inventory/
2. Click on any system from the list and view its information under 'General information' tab

Actual results: Unable to review the Subscription information of a system, the info page of a system is lacking subscription relevant information

Expected results: The system info page has also subscription information (service-level, purpose role, purpose usage, purpose addons)

Use case:

- Add system purpose Role and Usage to some of my servers
- Run the Inventory Sync from Satellite 6.9+
- Go to Insights Inventory to review if the Satellite uploaded the correct values

### DIFF
--- a/src/__mocks__/selectors.js
+++ b/src/__mocks__/selectors.js
@@ -68,3 +68,11 @@ export const collectInfoTest = {
   insights_client_version: 'test-client',
   insights_egg_version: 'test-egg',
 };
+
+export const subscriptionsTest = {
+  system_purpose: {
+    usage: 'Development',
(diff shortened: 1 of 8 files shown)


leSamo (approved):
The code looks good, I was not able to find a system which has any subscription, but I believe its okay. I'd prefer not to use `connect` but rather `useDispatch`, but I understand that it's done for consistency with other cards.

PR #2211, opened 2024-06-06
fix(RHINENG-8684): Icon locator for image mode systems is not visible in Inspect

computercamplove:
Currently` aria-label` is not visible for image mode icon is not visible in Inspect
image

with this change 'aria-label` should be visible in Inspect 
image


adonispuente (approved):
LGTM

PR #2212, opened 2024-06-07
fix(RHINENG-358): update selectors

johnsonm325:
In the [previous PR](https://github.com/RedHatInsights/insights-inventory-frontend/pull/2208), we checked for systemProfile.system_purpose to find subscription info. But, we should be checking entity.facts instead.


computercamplove:
i added smoke test for it - it passed, thank you!

adonispuente (approved):
LGTM

PR #2217, opened 2024-06-25
fix(RHINENG-10413): Fix no entities found

johnsonm325:
On groups table, when you enter a filter that returns no groups, even if workspaces is enabled, it shows groups instead. This passes the correct "entities" to the NoEntitiesFound component depending on if the workspaces feature flag is enabled.


InsightsDroid:
## :exclamation: Outdated dependencies error
* Installed version of @redhat-cloud-services/frontend-components-config (5.1.1) is older than required version (6.0.0)

Fewwy (approved):

PR #2224, opened 2024-07-01
chore(Staleness): Add OUIA IDs for dropdowns

gkarat:
This adds OUIA IDs to the staleness dropdown (both on the conventional and immutable tabs).


src/components/InventoryHostStaleness/BaseDropDown.js
```diff
@@ -61,6 +62,7 @@ const BaseDropdown = ({
           width: '200px',
         }}
         status={!isFormValid && 'danger'}
+        data-ouia-component-id={ouiaId}
```
leSamo, line 65: Nitpick: `MenuToggle` does have `ouiaId` prop

leSamo (approved):

PR #2225, opened 2024-07-02
fix(RHINENG-10825): Use correct slug to the conversion analysis

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-10825.

Use correct (newly updated) slug for the conversion analysis task. The change covers the label and popover at /inventory and the host details page alert.

## How to test

1. Go to /inventory
2. Filter out CentOS Linux hosts and find the label next to the host display name
3. Click on the label and then on the button to get redirected to Tasks
4. The correct task modal must be open
5. Do the same for /inventory/:id page for some CentOS Linux host and the alert that should prompt the same action


mkholjuraev (approved):
LGTM!

PR #2228, opened 2024-07-10
refactor: Fix build warnings, circular dependencies

gkarat:
This aims to address multiple warnings we had during webpack build.


Fewwy (approved):
Tested it locally, LGTM

PR #2229, opened 2024-07-22
fix(RHINENG-10786): Remove occurrence of /preview

jiridostal:
Fixes https://issues.redhat.com/browse/RHINENG-10786.


gkarat (approved):
LGTM. I would wait for ci.int to pass first

PR #2230, opened 2024-07-25
chore: Use default electron browser when running test:ct

gkarat:
No jira.


gkarat:
This hasn't ended up well: we have an error in Travis
```
Cannot read properties of null (reading 'getWebSocketDebuggerUrl')
```
when switched to electron. I will try to take a deeper look soon

bastilian:
@gkarat Can we try to update cypress to the latest version?

src/components/GroupsTable/GroupsTable.cy.js
```diff
@@ -58,7 +58,8 @@ const checkSorting = (label, order, dataField) => {
   if (order === 'ascending') {
     cy.get(header).find('button').click();
   } else {
-    cy.get(header).find('button').click().click();
+    cy.get(header).find('button').click();
+    cy.get(header).find('button').click();
```
gkarat, line 62: `dblclick` does two clicks almost immediately, it doesn't really work here due to a slight delay in the table. So leaving it as just two separate `click` calls here

PR #2233, opened 2024-08-06
fix(RHINENG-10975): remove unused items variable in customGetEntities

mkholjuraev:
Resolves: https://issues.redhat.com/browse/RHINENG-10975

Please test this thoroughly as I found it hard to reproduce locally.

To test:
1. Navigate to /workspaces list page
2. On stage environment find some workspace with immutable systems added
3. Navigate to its detail page
4. Switch to immutable tab
5. Observe that the page functions as expected and you only see those added immutable systems in the table

### DIFF
--- a/src/components/GroupSystems/GroupImmutableSystems.js
+++ b/src/components/GroupSystems/GroupImmutableSystems.js
@@ -100,7 +100,7 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
   ) => {
     const enhancedConfig = enhancedEdgeConfig(groupName.toString(), config);
     const defaultData = await defaultGetEntities(
-      items,
+      null,
       enhancedConfig,
       showTags
     );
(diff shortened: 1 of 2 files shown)


src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -100,7 +100,7 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
   ) => {
     const enhancedConfig = enhancedEdgeConfig(groupName.toString(), config);
     const defaultData = await defaultGetEntities(
-      items,
+      null,
```
gkarat, line 103: Do we want to send null? Is it necessary when we could just fix the Reference error by simply declaring `let items = []` before this function call?

src/components/GroupSystems/GroupImmutableSystems.js
```diff
@@ -109,21 +109,19 @@ const GroupImmutableSystems = ({ groupName, groupId, ...props }) => {
...
-      items = rowInfo.map(({ id }) => id);
 
       return {
         results: mergeArraysByKey([defaultData.results, rowInfo]),
         total: customResult?.data?.total,
       };
     } else {
       return {
-        results: mergeArraysByKey([defaultData.results]),
+        results: mergeArraysByKey([]),
```
gkarat, line 124: Why this is changed?

src/store/inventory-actions.js
```diff
@@ -101,6 +101,8 @@ export const loadEntities = (
         hideFilters: config.hideFilters,
       }))
       .catch((error) => {
+        //Somehow this catch block hides prior JS errors. Log is intended to be aware of them
+        console.log(error);
```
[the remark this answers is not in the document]
mkholjuraev: BTW, you are right

src/components/GroupSystems/GroupImmutableSystems.js
mkholjuraev: We do not have to pass anything. Because, this is not needed at all, rather it may result in non-immutable systems to appear.

gkarat (approved):
LGTM, thanks @mkholjuraev

PR #2234, opened 2024-08-07
fix(InventoryTable): RHINENG-9871 - Correct export resource

bastilian:
This corrects the resource name for the export service.

**How to test:**

1) Open Systems page (for conventional systems)
2) Trigger an export, either CSV/JSON
3) Verify that the export request was submitted and a "prepare export" notification appeared.

PR #2236, opened 2024-08-09
fix(InventoryTable): RHINENG-11904 -  Preserve a passed exportConfig

bastilian:
Some applications using the `InventoryTable` provide an export of systems, in these cases we should not override the passed export config for the toolbar.

**How to test:**

1) Ensure the `hbi.export-data` feature flag is enabled
2) Open an application that provides a custom export (Vulnerability, Compliance for example) on the "Systems" page
3) Verify that the CSV/JSON export is possible
4) Open the Inventory Systems page and verify that the export is still available there too.

### DIFF
--- a/src/components/InventoryTable/EntityTableToolbar.test.js
+++ b/src/components/InventoryTable/EntityTableToolbar.test.js
@@ -843,22 +843,61 @@ describe('EntityTableToolbar', () => {
 
   describe('export', () => {
     const formats = ['json', 'csv'];
+    const format = formats[Math.floor(Math.random() * formats.length)];
+    const axiosPostMock = jest.fn(async () => {});
 
-    it('should call the exports API', async () => {
-      const format = formats[Math.floor(Math.random() * formats.length)];
-      const axisoPostMock = jest.fn(async () => {});
+    beforeEach(() => {
       useAxiosWithPlatformInterceptors.mockImplementation(() => {
         return {
-          post: axisoPostMock,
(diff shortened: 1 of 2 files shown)


src/components/InventoryTable/EntityTableToolbar.test.js
```diff
@@ -843,22 +843,61 @@ describe('EntityTableToolbar', () => {
 
   describe('export', () => {
     const formats = ['json', 'csv'];
+    const format = formats[Math.floor(Math.random() * formats.length)];
+    const axisoPostMock = jest.fn(async () => {});
```
mkholjuraev, line 847: Super minor
```suggestion
    const axiosPostMock = jest.fn(async () => {});
```

mkholjuraev (commented):
Now, export exists in other apps and inventory. But, I encountered another issue in inventory. When I tried exporting, I got `{"message":"Payload does not match Configured Exports","code":406}` from the API and export failed

src/components/InventoryTable/EntityTableToolbar.test.js
bastilian: Nice catch!

bastilian:
@mkholjuraev The issue you saw when submitting the export will be fixed via https://github.com/RedHatInsights/insights-inventory-frontend/pull/2234.

mkholjuraev (approved):
LGTM!

PR #2237, opened 2024-08-13
chore: do run commit-lint in both master and pull requests

mkholjuraev:
This PR makes inventory consistent with other apps when it comes to how commit lint is run. With this change, commit lint is run for all commits on master and also for all pull requests commits.


bastilian (approved):
LGTM! Thank you @mkholjuraev!

PR #2240, opened 2024-08-14
fix(RHINENG-4792): Don't show filter chip for whitespace

leSamo:
https://issues.redhat.com/browse/RHINENG-4792

Fixed the following bug:
1. Go to Systems page
2. Type " " (space) inside the search text field
3. Filter chip appears, but it shouldn't


bastilian (approved):
LGTM! Thank you @leSamo!

PR #2243, opened 2024-08-16
test: Fix flaky HybridInventory.cy.js test

gkarat:
We don't pass `data-ouia-safe` to InventoryTable and it's set to true by default. Though, in the flaky test, for some reason, `data-ouia-safe` was missing completely, which makes this attribute somewhat unreliable to use: https://app.travis-ci.com/github/RedHatInsights/insights-inventory-frontend/jobs/625131895#L6568. HybridInventory.cy.test should use other way to indicate the table readiness.

### DIFF
--- a/src/routes/InventoryComponents/HybridInventory.cy.js
+++ b/src/routes/InventoryComponents/HybridInventory.cy.js
@@ -41,11 +41,7 @@ const waitForTable = (waitNetwork = false) => {
   }
 
   // indicating the table is loaded
-  cy.get('table[aria-label="Host inventory"]').should(
-    'have.attr',
-    'data-ouia-safe',
-    'true'
-  );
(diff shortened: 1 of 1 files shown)


mkholjuraev (commented):
Should not we fix passing `data-ouia-safe` to InventoryTable or give it a default value instead?

gkarat:
@mkholjuraev, it's not an easy task to program `data-ouia-safe` properly: you should account for all data to be finished fetching, consider if all the animations are complete, etc. When, in the context of the tests, we have another indicator of table loading complete like a first row (or any other from the fixtures data) to be rendered. I think it's an easier solution and not requiring additional work in the source code (not the intention of this PR).

mkholjuraev (approved):
LGTM!

PR #2244, opened 2024-08-20
feat(InventoryTable): RHINENG-9872 - Implement export download notification

bastilian:
This adds polling for exports and downloading. as a temporary solution till the export service itself can notify the users.

**How to test:**

1) Open the Inventory Systems page (with the export feature enabled)
1.1) Use the rhn-engineering-sgraessl account to not stress the backend to much and make sure the export/download works.
2) Click the export and choose a format.
3) Verify that a "pending notification" was triggered.
4) Verify that a "downloading export notification" was triggered and the desired export has been downloaded.


src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -256,12 +255,6 @@ const EntityTableToolbar = ({
       ...activeFilters,
       ...customFilters,
     },
-    onSuccess: () => {
```
bastilian, line 259: This is moved into the hook itself to consolidate really everything related to export in the toolbar.

src/components/InventoryTable/EntityTableToolbar.test.js
```diff
@@ -842,18 +842,6 @@ describe('EntityTableToolbar', () => {
   });
 
   describe('export', () => {
-    const formats = ['json', 'csv'];
```
bastilian, line 845: At this level we don't need to be this specific I think. So this is moved and the hook itself is now tested with a random format.

src/components/InventoryTable/EntityTableToolbar.test.js
```diff
@@ -916,21 +910,11 @@ describe('EntityTableToolbar', () => {
...
-        name: 'inventory-export',
-        sources: [
-          {
-            application: 'urn:redhat:application:inventory',
-            resource: 'urn:redhat:application:inventory:export:systems',
-            filters: {},
-          },
-        ],
-      });
+      await waitFor(() => expect(axiosPostMock).toHaveBeenCalled());
```
mkholjuraev, line 917: Why do we now prefer a test just asserts only the function to be called. Is not it better to have additional assertion on the function call arguments?

src/components/InventoryTable/hooks/useInventoryExport/helpers.js
```diff
@@ -0,0 +1,32 @@
...
+    {
+      application: EXPORT_SERVICE_APPLICATON,
+      resource: EXPORT_SERVICE_RESOURCE,
+      filters,
+    },
+  ],
+});
+
+// TODO Extend the function in fec to allow downloading a link and not just a blob and replace this.
+export const downloadFile = async (
```
mkholjuraev, line 19: Can we use the shared function https://github.com/RedHatInsights/frontend-components/blob/1599246a16e70d78401e70628797d5fef24f4fae/packages/utils/src/helpers/helpers.ts#L14

src/components/InventoryTable/hooks/useInventoryExport/useExportDownloadCheck.js
```diff
@@ -0,0 +1,58 @@
...
+        setCheckForDownload(undefined);
+        onError?.(error);
+      }
+    },
+    [list, onDownloadAvailable, checking]
+  );
+
+  useEffect(() => {
+    if (checkForDownload) {
+      const downloadCheck = setInterval(async () => {
```
mkholjuraev, line 40: Just a note: [Server side events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) would be probably perfect for this need, instead of polling the server. But, I know that is already late :)

mkholjuraev (approved):
LGTM! Worked like a charm. I was able to export both Json and Csv files, the notifications popped up as expected.

src/components/InventoryTable/hooks/useInventoryExport/helpers.js
bastilian: At the moment the shared function only supports "downloading" blobs and wouldn't work correctly here. yet. Once we updated the shared function to work by just passing a URL we can use it here as well.

src/components/InventoryTable/hooks/useInventoryExport/useExportDownloadCheck.js
bastilian: I could be wrong, but I think that server side events would only work if the server would sent events.

src/components/InventoryTable/EntityTableToolbar.test.js
bastilian: This assertion is simply only to ensure that clicking triggered the right action. What is passed and how is tested on the hook itself.

src/components/InventoryTable/hooks/useInventoryExport/useExportDownloadCheck.js
bastilian: Also, eventually, hopefully, 🤞, this check will go away and the Export service itself will trigger a notification for the download (via websockets).

src/components/InventoryTable/hooks/useInventoryExport/helpers.js
mkholjuraev: Aha, let's keep this in mind and make it shared function later. I am sure that it will be used somewhere in the future.

src/components/InventoryTable/EntityTableToolbar.test.js
bastilian: The test could also just be to assert the export button is available, because most of the implementation for the exporting itself is contained within the hook, and the only thing that the toolbar really does is include or not include the export based on `enableExport`.

PR #2245, opened 2024-08-21
fix(RHINENG-9836) duplicate fetch on systems table

johnsonm325:


doc/props_table.md
```diff
@@ -100,24 +100,6 @@ Example in [Patchman UI](https://github.com/RedHatInsights/patchman-ui/blob/mast
...
-*boolean*
-
-When `true`, the table is in loading state on mount until `entities.loaded` is set to `false` (and from that point, `loaded` is the only determinator.). Use when users can go back to already loaded table, this prop ensures that there will be no change from `loaded` > `loading` > `loaded`.
-
-## ignoreRefresh
-
-*boolean = true*
-
-On the initial mount and when items/sortBy are changed, the inventoryTable ignores `onRefresh` prop. By setting the prop to false, you can control this behavior.
-
```
gkarat, line 120: These parameters can be used by other applications that use InventoryTable. Have we checked if this is not a breaking change?

bastilian: As far as I remember when making the original changes, these props were wither already not used or the changes removed the behaviour these props would enable and made them obsolete. 

The behaviours these props enabled were mostly workarounds and unintuitive. For example the `autoRefresh` is kinda redundant, because we should always refresh the table when customFilters change, otherwise we have the contents and filters out of sync. This was most likely a "fix" for a bug that caused the table not to get updated automatically. 

However, in any case we should test these changes with some/all apps and maybe do a quick GitHub search for these specifically.

src/components/InventoryTable/InventoryList.test.js
```diff
@@ -50,7 +50,12 @@ describe('InventoryList', () => {
     render(
       <MemoryRouter>
         <Provider store={store}>
-          <InventoryList ref={ref} onRefreshData={onRefreshData} loaded />
+          <InventoryList
+            ref={ref}
+            onRefreshData={onRefreshData}
+            loaded
+            hasAccess={true}
```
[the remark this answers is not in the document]
johnsonm325: @bastilian It's because [I've changed](https://github.com/RedHatInsights/insights-inventory-frontend/pull/2245/files#diff-41dbf31730868a3d85a1277faa8a678ec8afb3b2a9f3605913837e182d9974beR37) default hasAccess to false, which is the correct thing to do until you fetch privileges.  It seems like a great security measure. Do you agree?

PR #2246, opened 2024-08-22
fix(RHINENG-12079): Update alert text when updating Ansible hostname

computercamplove:


AsToNlele (approved):
LGTM, thank you! 😄
I've restarted both tests, let's wait for them to pass

PR #2247, opened 2024-08-30
chore: Remove unused dependencies

gkarat:
No jira.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,8 +20,6 @@
     "@unleash/proxy-client-react": "^3.5.0",
     "awesome-debounce-promise": "^2.1.0",
     "classnames": "^2.3.1",
-    "graphql": "^15.6.0",
-    "html-webpack-plugin": "^5.5.0",
     "lodash": "^4.17.21",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -122,8 +115,7 @@
     "stylelint-scss": "^3.21.0",
     "ts-patch": "^3.1.2",
     "typescript": "^5.4.5",
-    "webpack-bundle-analyzer": "^4.5.0",
-    "whatwg-fetch": "^3.6.2"
+    "webpack-bundle-analyzer": "^4.5.0"
```
bastilian, line 118: I think we can also remove the analyser stuff. If we configure this we should do it somewhere central to include in all apps.

gkarat: Yeah, forgot about this one

bastilian:
Do we use the "html-webpack-plugin" package? I think that may be remnant of some previous webpack config.

bastilian (approved):
LGTM! Everything appears to still work as expected! Thank you @gkarat!

PR #2249, opened 2024-09-10
fix(InventoryTable): RHINENG-9872 - Make download more reliable

bastilian:
This changes the export check for downloads to use the `/status` endpoint of the exports to check for the completion. The download check now also recognises "failed" exports and will abort the download (check) and trigger an error notification. And the filename provided in the content disposition header is used for the downloaded filed if provided.

**How to test:** 

1) Ensure the export feature is enabled for the inventory
2) Visit the Systems page
3) Click the "export button" and choose either CSV or JSON to download
4) Verify the export was properly triggered and the export was downloaded on success
4.1) There is no easy way to verify the error handling. But exports on accounts with many systems may still fail at the moment.


mkholjuraev (approved):
LGTM! Codewise looks robust as usual. I tried locally, download happens without issues and the title is taken from the header. I was not able to tests the failing state. But, I expect it works.

PR #2251, opened 2024-09-19
chore(tests): Update coverage config

bastilian:
This should fix the coverage reporting to codcov.

**How to test:**

* Run `rm -rf node_modules && npm install && npm run travis:verify`
* The HTML reports in coverage, coverage-cypress, and coverage-jest to make sense.


config/cypress.webpack.config.js
```diff
@@ -20,32 +20,40 @@ const { config: webpackConfig, plugins } = config({
...
   resolve: {
-    alias: {
-      '@redhat-cloud-services/frontend-components/useChrome': resolve(
-        __dirname,
-        './overrideChrome.js'
-      ),
-      '../useChrome': resolve(__dirname, './overrideChrome.js'),
+    fallback: {
+      stream: require.resolve('stream-browserify'),
+      zlib: require.resolve('browserify-zlib'),
```
gkarat, line 56: Could you provide a brief background on why this is added?

gkarat:
With the current changes, I can see that the coverage reports are generated fine 👍 and it looks like both cypress and jest tests are included in the coverage

bastilian:
@gkarat I want to wait for https://github.com/RedHatInsights/frontend-components/pull/2066 to be merged, as `react-intl` should not be a direct dependency.

config/cypress.webpack.config.js
bastilian: I just took what was in compliance configured. I'll try and see if and why this would be required.

.gitignore
```diff
@@ -15,20 +15,20 @@ dist
...
-coverage
 sitemap.xml
 
 # cypress
 cypress/downloads
 cypress/screenshots
 cypress/videos
 
-codecov*
+codecov
```
LightOfHeaven1994, line 27: Do we need this `codecov`?

LightOfHeaven1994 (approved):
Generally LGTM, great work! found only 1 minor thing

.gitignore
bastilian: We don't need it. Sorta. it's just so that the `codcov` binary doesn't show up as a change in git. I'll remove it to align with other apps.

PR #2253, opened 2024-09-19
fix(RHINENG-2632): Separate Satellite and Discovery data collectors

kruai:
This PR addresses [RHINENG-2632](https://issues.redhat.com/browse/RHINENG-2632). It separates the Satellite and Discovery data collectors. This change is needed ASAP because incoming host payloads now use the "satellite" and "discovery" values rather than "yupana".


chris1984 (approved):
Looks fine to me, from an outside person looking at it.

bastilian (approved):
LGTMT! Thank you @kruai!

PR #2268, opened 2024-10-17
feat(RHINENG-14072): Add sentry integration for inventory

adonispuente:
This adds the ability to upload source maps from inventory into sentry. It also adds the pointer to tell sentry which project to send the error too.

Testing is a bit tricky, as currently you need to run chrome-prod in tandem with inventory to test the actual error. You will need an authKey present (reach out to me for the file).

in inventory, run` SENTRY_AUTH_TOKEN=IWILLGIVETHISTOYOU  npm run static`,
Create an error
You can create an easy error by adding` <button onClick={() => methodDoesNotExist()}/>` anywhere and clicking the button.

In chrome youd need to have a prod account ->
change webpack.config.js file env var too '`prod-stable`' ->
add this to routes` ...(true && { '/apps/inventory': { host: 'http://localhost:8003', }, })`,->
and run npm run dev.

actually go ahead and cause the error by pressing the button. 
Observer that the error then goes to the inventory-rhel project and has valuable data inside.

Here is an example: https://red-hat-it.sentry.io/issues/6002273624/?project=4508083145408512&referrer=onboarding-first-event-indicator


bastilian (approved):

PR #2271, opened 2024-10-21
fix(RHINENG-8385): Make tag filter search box size look correct

LightOfHeaven1994:
Before:
image

After:
image

### DIFF
--- a/src/components/InventoryTable/InventoryList.scss
+++ b/src/components/InventoryTable/InventoryList.scss
@@ -18,6 +18,10 @@
     .ins-c-chip-filters .pf-v5-c-chip-group:not(:last-of-type) { margin-right: var(--pf-v5-global--spacer--xs); }
 
     .ins-c-tagfilter { width: initial; }
+
+    .ins-c-tagfilter.pf-v5-c-menu-toggle {
+        height: 36px;
+    }
 }
 
 .ins-c-table__empty {


src/components/InventoryTable/InventoryList.scss
```diff
@@ -18,6 +18,10 @@
     .ins-c-chip-filters .pf-v5-c-chip-group:not(:last-of-type) { margin-right: var(--pf-v5-global--spacer--xs); }
 
     .ins-c-tagfilter { width: initial; }
+
+    .ins-c-tagfilter.pf-v5-c-menu-toggle {
+        height: 36px;
```
mkholjuraev, line 23: I think it is better to use some PF variables instead of 36px. Because, this may work for certain screen resolution, but not for others.

mkholjuraev (changes requested):

src/components/InventoryTable/InventoryList.scss
mkholjuraev: Also, the issue should be related to something not properly handled with regards to the tag filter dropdown. As this issue happens to tag filter only, not others.

LightOfHeaven1994: @mkholjuraev thanks for the review! I agree with you. Change has to be done on FEC side so closing this PR

PR #2273, opened 2024-10-23
fix(RHINENG-13857): Make tags aligned with table header

LightOfHeaven1994:
Before:
image

After:
image


leSamo (approved):
Thank you, this bothered me for a really long time :smile:

PR #2275, opened 2024-10-24
fix(RHINENG-11788): Add reg assistant link to sys details

johnsonm325:
The details page for a system not registered with Insights is not pointing to the main source of information for registering the system. We should include a link to 'Registration Assistant'.

To test, navigate to inventory and click on the name of a system that is not registered (yellow icon). A link to the registration assistant should be provided as this is the main location to find information on how to register your system. The banner at the top of the page should include "How to register with insights-client?", and it should link to Registration Assistant app.


LightOfHeaven1994 (approved):
lgtm!

PR #2276, opened 2024-10-25
fix(RHINENG-12428): Make header look correct

LightOfHeaven1994:
Fix items order

Before:
image

After:
image


adonispuente (approved):
LGTM, thank you!

PR #2279, opened 2024-10-29
fix(RHINENG-12953): strip whitespace edit display and ansible hostname

johnsonm325:
It is possible to set both Display name and Ansible hostname to a space character, ' '. Not only this makes the UI confusing, it makes you unable to click the host in the Systems page: you have to click Edit, change the name and only then you can open the host details.

Trailing whitespace is cleaned on some places, but not consistently; display name "test   " is stored as such.

To test, find a system and click the system to view its details. Edit the display name and ansible hostname values.

This PR disables the Save button if the name hasn't changed, or is empty, including only whitespaces. Also, if you add whitespace before or after the string, it will trim the whitespace on both sides.

### DIFF
--- a/src/components/GeneralInfo/TextInputModal/TextInputModal.js
+++ b/src/components/GeneralInfo/TextInputModal/TextInputModal.js
@@ -1,7 +1,15 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
-import { Button, Modal, TextInput } from '@patternfly/react-core';
+import {
+  Button,
+  FormHelperText,
+  HelperText,
+  HelperTextItem,
+  Modal,
+  TextInput,
+  ValidatedOptions,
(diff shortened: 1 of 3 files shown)


adonispuente:
Functionality seems to work as intended.
Currently you can save the name if the name is exactly the same, but you add spaces on either side. If the goal is to also prevent sending a request if the name is the same then that needs to be altered. 

This doesnt have to be part of this PR, but while work is being done on it, currently for validation/errors it just gives you the error icon on the input but with no messaging prompt, so the user has no idea why they cant save.

johnsonm325:
@adonispuente It's fine if the user is allowed to submit with the old name and whitespace. It doesn't break the experience.

I have added helper text.

adonispuente (approved):
Works as intended, LGTM

PR #2282, opened 2024-10-31
fix(RHINENG-13293): Fix undefined values on staleness page while loading

adonispuente:
https://issues.redhat.com/browse/RHINENG-13293
On page load, setIsLoading would fire BEFORE all of the calls were completed because these are all asychronous functions returning promises. They were starting and setIsLoading would fire without them being completed. 

Each call needs to be awaited to ensure the async functions are completed before firing off the setIsloading(false);
To test just navigate to staleness and deletion page and observe the drop downs on page load. They should only load in with data, and never display undefined.

Before: 
Screenshot 2024-10-31 at 2 57 02 PM

After: (just an example, data can be different depending on the account.
Screenshot 2024-10-31 at 2 59 59 PM

### DIFF
--- a/src/components/InventoryHostStaleness/HostStalenessCard.js
+++ b/src/components/InventoryHostStaleness/HostStalenessCard.js
@@ -178,9 +178,9 @@ const HostStalenessCard = ({ canModifyHostStaleness }) => {
   };
 
   const batchedApi = async () => {
-    edgeSystemCheck();
-    fetchApiStalenessData();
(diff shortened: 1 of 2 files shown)


adonispuente:
Id like to ignore the code climate failure. Its angry cause I reused the render function but youre not supposed to use render inside beforeEach or beforeAll.

johnsonm325 (approved):

PR #2301, opened 2025-01-13
feat(InventoryTable): RHINENG-15246 - Reword export dropdown items

bastilian:
Changes the text of the export dropdown according to the Jira.

**Before:**

Screenshot 2025-01-21 at 12 35 46

**After:**
Screenshot 2025-01-21 at 12 40 57

### DIFF
--- a/cypress/support/component.js
+++ b/cypress/support/component.js
@@ -1,3 +1,8 @@
 import '@patternfly/patternfly/patternfly.scss';
 import '@cypress/code-coverage/support';
 import './commands';
+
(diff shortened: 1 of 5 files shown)


johnsonm325:
@bastilian failing cypress tests, as you can see. The same test files are failing locally as what I see from Travis. But one of the specific tests and its details are different. This is what I get locally. Do we know if these are simply flaky?

image

johnsonm325 (approved):
This looks good. Seems like the tests are flaky and certainly not related to this.

PR #2313, opened 2025-01-14
fix(sentry): Update sentrys workflow and always inject debugIDs

adonispuente:
Updates flow to always inject debugID, and update the workflow file to remove overriding of release in sentry, and map against debug IDs properly


LightOfHeaven1994 (approved):
LGTM, thank you!

PR #2323, opened 2025-01-17
fix(RHINENG-15122): sorting kept on refresh

AsToNlele:
[RHINENG-15122](https://issues.redhat.com/browse/RHINENG-15122)

Sorting should persist between refreshes

Test without any parameters and as well with some sorting applied


johnsonm325 (approved):
The code looks good to me. Either create some tests here, or create a card and put it in Future sprints to create tests.

AsToNlele:
added cypress tests

johnsonm325:
@AsToNlele I've run the tests locally twice and I get the same failing tests each time, but they are different from the failing tests in Travis. Not sure if this is flaky tests for what.

image

AsToNlele:
uh not again, everything passes for me

AsToNlele:
I checked previous commits and they have the same issue, I'll investigate

PR #2338, opened 2025-01-28
fix(sentry): Update sentry flow

adonispuente:
This also removes the issue where the plugin is initialized locally and sends up debugIDs for local env on every build
**Workflow Behavior:**

**Locally (without variables):**
The Sentry plugin won't build as SENTRY_AUTH_TOKEN/ENABLE_SENTRY is not defined.

**GitHub push to master (automatic trigger):**
The plugin won't include the authToken because commit_hash is not provided in push events.

**Manually triggered with commit_hash:**
The plugin includes the authToken and tracks the build with Sentry.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,12 +1,15 @@
-name: sentryConfig
+name: sentryInit
 
 on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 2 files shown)


bastilian (approved):

PR #2339, opened 2025-01-28
fix(RHINENG-15627): Images table persistent loading

johnsonm325:
A regression happened where when you load the images table, it would remain in a persistent loading state. This PR fixes that issue.

To test:
- go to `/inventory` and click the images button for "View by images"

See that the table constantly shows loading state, but the fetch happens.


LightOfHeaven1994 (approved):
LGTM, issue is fixed

PR #2346, opened 2025-01-30
fix(RHINENG-15664): modal links support trailing slash

AsToNlele:
[RHINENG-15664](https://issues.redhat.com/browse/RHINENG-15664)

Issue was we weren't expecting a trailing slash in the url.
Added some tests

How to test:
1. Go to Inventory Detail page > Infrastructure Card > IPv4 address link (Other cards could be checked as well, but they use the same component)
2. Make sure the modal opens
3. Now add a `/` to the end of the URL
4. Make sure the modal still works

### DIFF
--- a/src/Utilities/sharedFunctions.js
+++ b/src/Utilities/sharedFunctions.js
@@ -68,3 +68,5 @@ export const loadSystems = (options, showTags, getEntities) => {
 
   return loadEntities(limitedItems, config, { showTags }, getEntities);
 };
+
+export const removeTrailingSlash = (path) => path.replace(/\/$/, '');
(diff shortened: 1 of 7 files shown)


AsToNlele:
ah, accidentally fixed some other tests as well

bastilian:
@AsToNlele Do you know where the "extra slash" comes from? We should rather try to figure out how a user would get to the URL mentioned in the Jira and make sure that it doesn't add an unnecessary slash.

AsToNlele:
@bastilian user is redirected from [Subscriptions service](https://github.com/search?q=repo%3ARedHatInsights%2Fcuriosity-frontend%20insights%2Finventory&type=code)

AsToNlele:
This approach is fine, I'll make a PR to Subscriptions service as well and see if they're willing to change it.

bastilian:
@AsToNlele If the links are fixed, do we still need this PR?

PR #2352, opened 2025-02-05
fix(RHINENG-9868): Fix custom last seen filter to option default date

leSamo:
https://issues.redhat.com/browse/RHINENG-9868


bastilian (approved):
LGTM! Thank you for fixing this @leSamo!

PR #2353, opened 2025-02-06
feat(RHINENG-11178): add workspace to system detail card

xmicha82:
Added Workspace to SystemCard in System detail page

Steps to test:
1. Go to System detail page
2. Observe System properties card - Workspace field
    - Not available on a system not assigned to a workspace
image
    - link to a workspace on a system assigned to a workspace 
image

Before: 
image

### DIFF
--- a/src/components/GeneralInfo/LoadingCard/LoadingCard.js
+++ b/src/components/GeneralInfo/LoadingCard/LoadingCard.js
@@ -18,6 +18,7 @@ import {
   SkeletonSize,
 } from '@redhat-cloud-services/frontend-components/Skeleton';
 import { Link, useLocation } from 'react-router-dom';
+import InsightsLink from '@redhat-cloud-services/frontend-components/InsightsLink';
 
 const valueToText = (value, singular, plural) => {
   if ((value || value === 0) && singular) {
@@ -42,6 +43,14 @@ export const Clickable = ({ value, target, plural, singular, onClick }) => {
       onClick({ value, target });
     }
   }, [modalId, target]);
+
+  if (target?.[0] === '/') {
+    return (
+      <InsightsLink to={target} app="inventory">
+        {valueToText(value, singular, plural)}
+      </InsightsLink>
+    );
+  }
 
   return (
     <Link to={`${pathname}/${target}`}>
@@ -120,7 +129,7 @@ const LoadingCard = ({
                             <Skeleton size={size || SkeletonSize.sm} />
                           )}
                           {!isLoading &&
-                            (onClick && value ? (
+                            ((onClick || target?.[0] === '/') && value ? (
                               <div>
                                 <Clickable
                                   onClick={onClick}
--- a/src/components/GeneralInfo/SystemCard/SystemCard.js
+++ b/src/components/GeneralInfo/SystemCard/SystemCard.js
@@ -79,6 +79,7 @@ class SystemCardCore extends Component {
       hasHostName,
       hasDisplayName,
       hasAnsibleHostname,
+      hasWorkspace,
       hasSAP,
       hasSystemPurpose,
       hasCPUs,
@@ -89,6 +90,7 @@ class SystemCardCore extends Component {
       extra,
     } = this.props;
     const { isDisplayNameModalOpen, isAnsibleHostModalOpen } = this.state;
+
     return (
       <Fragment>
         <LoadingCard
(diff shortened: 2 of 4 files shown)


src/components/GeneralInfo/LoadingCard/LoadingCard.js
```diff
@@ -33,7 +33,14 @@ const valueToText = (value, singular, plural) => {
   return value || 'Not available';
 };
 
-export const Clickable = ({ value, target, plural, singular, onClick }) => {
+export const Clickable = ({
+  value,
+  target,
+  fullTarget,
```
bastilian, line 39: What's `fullTarget` for? Why do we need this?

src/components/GeneralInfo/SystemCard/SystemCard.js
```diff
@@ -159,6 +161,18 @@ class SystemCardCore extends Component {
...
               : []),
+            ...(hasWorkspace
+              ? [
+                  {
+                    title: 'Workspace',
+                    value:
+                      entity.groups?.length > 0 && entity.groups?.[0]?.name,
+                    size: 'md',
+                    customClass: 'sentry-mask data-hj-suppress',
+                    fullTarget: `/insights/inventory/workspaces/${entity.groups?.[0]?.id}`,
```
bastilian, line 172: We should not have this (`/insights/inventory/` specifically) hardcoded in here. The other paths for the links are constructed with help from the pathname, would this not work for this section?

src/components/GeneralInfo/LoadingCard/LoadingCard.js
xmicha82: target basically specifies only what modal should be opened based on URL
thus if i want to redirect to a completely different page, i use a different parameter

src/components/GeneralInfo/SystemCard/SystemCard.js
xmicha82: in this case most of the location would be discarded in order to arrive at the correct path, so i assumed this would be the better solution

bastilian: > in this case most of the location would be discarded in order to arrive at the correct path
I'm not sure I understand what you mean?

bastilian: Please look into how we can get rid of the `/insights/inventory/`. We are already within `/insights/inventory/` and should not need it. If the path changes in the future we have to remember to change it here as well, so we should try to not have it here in the first place.

johnsonm325: @xmicha82 we can use the `InsightsLink` component. See how other apps use it: https://github.com/search?q=org%3ARedHatInsights%20%3CInsightsLink&type=code

johnsonm325 (changes requested):
I have some ideas of how we can make this better. Let's discuss after standup on Monday and determine the best way to handle it. I'd like yours and others opinions.

src/components/GeneralInfo/LoadingCard/LoadingCard.js
johnsonm325: @xmicha82 Sebastian made a great suggestion, what if we pass in the link as `target` and then do a check in `LoadingCard` like so:

```to={target[0] === '/' ? target : `${pathname}/${target}`}```

File: src/components/GeneralInfo/LoadingCard/LoadingCard.js [hunk 3]
```diff
@@ -120,12 +130,13 @@ const LoadingCard = ({
...
                           )}
                           {!isLoading &&
-                            (onClick && value ? (
+                            ((onClick || target?.[0] === '/') && value ? (
                               <div>
                                 <Clickable
                                   onClick={onClick}
                                   value={value}
                                   target={target}
+                                  fullTarget={fullTarget}
```
johnsonm325, line 139: remove

src/components/GeneralInfo/LoadingCard/LoadingCard.js
```diff
@@ -88,6 +97,7 @@ const LoadingCard = ({
                       onClick,
                       value,
                       target,
+                      fullTarget,
```
johnsonm325, line 100: remove

johnsonm325 (changes requested):
This looks so much cleaner and straight-forward. Great job.

Just remove those 2 references to `fullTarget` and ping me in slack so I can approve it quickly.

johnsonm325 (approved):

PR #2356, opened 2025-02-07
fix: Fix JSDoc errors, add eslint rules

gkarat:
This fixes the errors that are triggered when running the JSDoc pipeline on master. Also it adds eslint rules that helps align the documentation comments.

### DIFF
--- a/src/components/InventoryDetail/AppInfo.js
+++ b/src/components/InventoryDetail/AppInfo.js
@@ -11,7 +11,11 @@ import {
 /**
  * Small component that just renders active detail with some specific class.
  * This component detail is accessed from redux if no component found `missing component` is displayed.
- * @param {*} props `componentsMapper` if you want to pass different components list.
+ *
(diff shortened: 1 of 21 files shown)


src/components/InventoryDetail/AppInfo.js
```diff
@@ -54,11 +58,7 @@ AppInfo.propTypes = {
   }),
 };
 
-/**
- * This component can be removed once all apps migrate to componentMapper and activeApp.
- * @param { componentMapper, activeApp } props.
- * @returns AppInfo component
- */
+// TODO: This component can be removed once all apps migrate to componentMapper and activeApp
```
bastilian, line 61: We could use the `@deprecated` tag here. Just to practice good practice. :) 

https://jsdoc.app/tags-deprecated

bastilian (approved):
LGTM! Thank you for setting this up @gkarat!

PR #2357, opened 2025-02-07
fix(RHINENG-15591): Fix OUIA ID passing for BaseDropdown component

gkarat:
This makes BaseDropdown to properly pass the OUIA ID prop. This was broken due to PF upgrades and end of direct `data-ouia-component-id` support.

Fixes https://issues.redhat.com/browse/RHINENG-15591.


leSamo (approved):
Nice, thank you for the test as well

PR #2361, opened 2025-02-13
chore(RHINENG-15918): Bump host-inventory-client to the latest version

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-15918.


gkarat:
Putting back to draft, I would like to also migrate the group/workspace-associated endpoints since they currently use instance directly, not through the APIFactory

gkarat:
I am still fighting with the ReactDebugCurrentFrame undefined error when clicking on the Workspaces filter in Inventory Table. But I believe the existing code changes are ready for review and I won't touch them anymore. @RedHatInsights/team-interact

johnsonm325 (approved):
Generally speaking, everything looks good. 47 files is a lot to review, but I see what's being done here. Once you rebase, you'll have to fix some differences with Samuel's workspace feature flag removal, but I don't think there will be anything significant to worry about.

Thanks for this!

src/api/api.js
```diff
@@ -425,36 +393,38 @@ export const getOperatingSystems = async (params = [], showCentosVersions) => {
...
+    return axiosInstance.get(`${EDGE_API_BASE}/devices/devicesview?limit=1`);
   } catch (err) {
     console.log(err);
   }
 };
 
 export const useGetImageData = () => {
   return (deviceIDs) => {
-    return instance.post(`${EDGE_API_BASE}/devices/devicesview`, deviceIDs);
+    return axiosInstance.post(
```
bastilian, line 417: We should at some point  make these use the proper axios instance.

.eslintrc.js
```diff
@@ -1,15 +1,17 @@
 module.exports = {
-  parser: '@babel/eslint-parser',
-  plugins: ['jsdoc'],
+  parser: '@typescript-eslint/parser',
+  plugins: ['jsdoc', '@typescript-eslint'],
```
bastilian, line 3: What is this for?

gkarat: This provides more lint rules for .ts modules

PR #2367, opened 2025-02-20
feat(RHINENG-15172): add OS mode to system detail

xmicha82:
# Before
## RHEL Package mode && RHEL Image mode 
System Detail

# After
## RHEL Package mode
RHEL Package mode
## RHEL Image mode
RHEL Image mode

### DIFF
--- a/src/components/InventoryDetail/DetailHeader.js
+++ b/src/components/InventoryDetail/DetailHeader.js
@@ -9,6 +9,7 @@ import { verifyCulledReporter } from '../../Utilities/sharedFunctions';
 import PageHeader from '@redhat-cloud-services/frontend-components/PageHeader';
 import classnames from 'classnames';
 import { REPORTER_PUPTOO } from '../../Utilities/constants';
+import OsModeLabel from './OsModeLabel';
 
 const HeaderInfo = ({
   entity,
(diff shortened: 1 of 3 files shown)


src/components/InventoryDetail/DetailHeader.js
```diff
@@ -34,6 +44,56 @@ const HeaderInfo = ({
   </Fragment>
 );
 
+const OsModeLabel = ({ isImageMode = false }) =>
```
bastilian, line 47: This might be better in a separate file.

src/components/InventoryDetail/OsModeLabel.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon, Label, Popover, PopoverPosition } from '@patternfly/react-core';
+import FontAwesomeImageIcon from '../FontAwesomeImageIcon';
+import { BundleIcon } from '@patternfly/react-icons';
+
+const OsModeLabel = ({ isImageMode = false }) =>
```
bastilian, line 7: This component has a few things that could be refactored. For example the `bodyContent`, header, icon and label text could be extracted into an `osModes` object, like for example `{ image: {icon: ..., headerContent: ...,bodyContent: ..., label: ...}, package: {...}`. 

This would allow to just pass in a `mode` prop with "image" or "package" and use it query the parts like `osModes[mode]` and allow avoid the ternary operator and just work with a single branch of component.

bastilian (approved):
Awesome! Nice work! Thank you @xmicha82!

PR #2369, opened 2025-02-25
fix(RHINENG-11372): Change Last upload to Last seen on System Details page

hugohezel:
Changes:

Before
image

After
Screenshot From 2025-02-25 10-24-14


bastilian (approved):
LGTM! Thank you for changing this @hugohezel!

PR #2376, opened 2025-03-04
feat(SystemStatus): add Last updated field to card

xmicha82:
Added Last updated field to System details -> System status card with data from `/host/:id` updated field
Last seen now displays most recent upload from reporters

### DIFF
--- a/src/components/GeneralInfo/GeneralInformation/__snapshots__/GeneralInformation.test.js.snap
+++ b/src/components/GeneralInfo/GeneralInformation/__snapshots__/GeneralInformation.test.js.snap
@@ -599,6 +599,20 @@ exports[`GeneralInformation should render correctly 1`] = `
                           Invalid date
                         </dd>
                         <dt
+                          aria-label="Last updated title"
+                          class=""
+                          data-ouia-component-id="Last updated title"
+                        >
+                          Last updated
+                        </dt>
+                        <dd
+                          aria-label="Last updated value"
+                          class=""
+                          data-ouia-component-id="Last updated value"
+                        >
+                          Not available
+                        </dd>
+                        <dt
                           aria-label="RHC title"
                           class=""
                           data-ouia-component-id="RHC title"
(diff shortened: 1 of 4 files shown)


src/components/GeneralInfo/GeneralInformation/__snapshots__/GeneralInformation.test.js.snap
```diff
@@ -598,6 +598,20 @@ exports[`GeneralInformation should render correctly 1`] = `
...
+                          data-ouia-component-id="Last updated title"
+                        >
+                          Last updated
+                        </dt>
+                        <dd
+                          aria-label="Last updated value"
+                          class=""
+                          data-ouia-component-id="Last updated value"
+                        >
+                          Invalid date
```
johnsonm325, line 613: These are a clue that we either need tests, and/or we have a bug.

I'm working under the assumption that there is never a system that does not have an `entity.updated` or `entity.per_reporter_staleness` value. But, I think we can safeguard against this by displaying something other than 'Invalid date'.

src/components/GeneralInfo/SystemStatusCard/SystemStatusCard.js
```diff
@@ -43,6 +44,25 @@ const SystemStatusCardCore = ({
...
+                  type="exact"
+                />
+              ),
+            },
+          ]
+        : []),
+      ...(hasLastUpdated
+        ? [
+            {
+              title: 'Last updated',
```
johnsonm325, line 65: Please change `entity.per_reporter_staleness &&` to `entity.per_reporter_staleness ?` and have the false option display "Not available". Do this for both "Last seen" and "Last updated".

Also, let's write a couple of tests in the `SystemStatusCard.tests.js` file.

- One that specifically tests the map and sort functionality
- One that tests that an invalid date properly displays something like, "Not available"

johnsonm325 (changes requested):
These should be straight-forward and I think relatively simple. But, if there is any confusion, please ping me. We can pair program this and get it done in under an hour I think.

johnsonm325 (approved):
beautiful

PR #2382, opened 2025-03-11
fix: Fix bugs with staleness and groups endpoints handlers

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-16620, https://issues.redhat.com/browse/RHINENG-16578.

This adds missing endpoints handlers and also makes it pass params correctly. This fixes the group creation and resetting staleness settings to default values.

## How to test

You must be able to create a group. You must be able to reset staleness settings to default values.


LightOfHeaven1994 (approved):
LGTM, both updates work as expected!

PR #2383, opened 2025-03-12
fix(InventoryTable): RHINENG-16584 - Add columnsProp to effect dependencies

bastilian:
We use `columnsProp`, so it needs to be a dependency. 

**How to test:**

1) Open the Inventory Systems page
2) Navigate to the Workspaces and to a workspace details page (**with systems**)
3) You should see the "Update method" column
4) Navigate to the Inventory Systems page
5) You should **not*& see the "Update method" column


LightOfHeaven1994 (approved):
LGTM, thank you!

PR #2384, opened 2025-03-12
fix(RHINENG-16066): Custom staleness menu displays undefined when selecting already selected value

hugohezel:
`onSelected()` method assigned selected item's `name` instead of the `value`, which caused that `getNameByValue(selected) `always returned `undefined`. However, the `undefined` text was visible only if user had selected item from menu, that had been already selected - that was the only case, when `useEffect()` reacting on `currentItem`'s change didn't override the `undefined` text.


bastilian (approved):
LGTM! Thank you for fixing this @hugohezel!

PR #2386, opened 2025-03-13
fix(GroupSystems): RHINENG-16584 - Copy array to not mutate columns

bastilian:
Fixes duplicate and unnecessary "update_method" columns in SytemsTable.

How to test:

1) Open the Inventory Systems page
2) Navigate to the Workspaces and to a workspace details page (with systems)
3) You should see the "Update method" column
4) Open the "add systems" modal and add systems to the workspace and save.
5) Navigate to the Inventory Systems page
6) You should *not& see any "Update method" columns


gkarat (approved):
LGTM, thanks @bastilian

PR #2392, opened 2025-03-18
fix(OSModeLabel): don't display on edge systems

xmicha82:
OsModeLabel should not be displayed on Edge systems


bastilian (approved):
LGTM! Thank you @xmicha82!

PR #2393, opened 2025-03-18
fix(TagsModal): RHINENG-12940 - Set activeSystemTag more explicitly

bastilian:
This fixes issues in applications like Compliance and Tasks, where the TagsModal misbehaves. 

The reason for this is that these applications do not use the inventory/insights ID of the systems as the `id` prop for their rows in the `InventoryTable`.

Changing the issue in the applications for them to use the inventory ID as `id` property might cause issues as these might rely on the `id` prop being their app specific systems ID. 

This PR adds properties ( id, insightsId, insights_id) to the lookup of the entity to set as activeSystemTag.

It also makes the setting of activeSystemTag more explicit to ensure the correct data is used in the modal even if the other data set exists.

**How to test:**

1) Run the Inventory with this PR
2) Open the Systems page
2.1) Verify the "tags links" in the table correctly open a TagModal showing the systems tags and no selection or apply button
2.2) Verify the Tags filters "More link" opens a TagModal correctly showing **all** tags available in the inventory, and it shows selection boxes as well as a apply button
2.2.1) Verify selected tags are applied as filters when clicking "Apply"
3) Open the details of a system
3.1) Verify the "tags link" opens a TagModal showing the systems tags and no selection or apply button.
4) Open the Compliance Systems page
4.1) Verify the things as with the Systems page in the Inventory and make sure the filter and tags link in each row open the correct TagModal

### DIFF
--- a/src/Utilities/TagsModal.js
+++ b/src/Utilities/TagsModal.js
@@ -31,11 +31,16 @@ const TagsModal = ({ filterTagsBy, onToggleModal, onApply, getTags }) => {
       entities?.tagModalLoaded || entityDetails?.tagModalLoaded
   );
 
-  const activeSystemTag = useSelector(({ entities, entityDetails }) =>
-    entities?.activeSystemTag?.created
-      ? entities.activeSystemTag
-      : entityDetails?.entity
-  );
+  const activeSystemTag = useSelector(({ entities, entityDetails }) => {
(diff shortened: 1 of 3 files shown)


LightOfHeaven1994 (approved):
Looks great, tested with couple apps + inventory. Works as expected!

PR #2394, opened 2025-03-19
fix(RHINENG-16138): Properly (un-)mount table in workspace modals

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-16138.

More instances of InventoryTable were rendered on Workspace details view causing race conditions in updating entities store. This fixes the bug where sorting or updating filters in modals can cause the table show incorrect data.

### How to test

Go to any workspace, either empty or having at least one system. Try to add new systems, both conventional or immutable, try to apply sorting, filters. Make sure the systems table on the workspace details shows only the systems that are included to this workspace. Make sure the systems shown in the add systems modal are all systems available in Inventory.


src/components/InventoryGroupDetail/GroupTabDetails.js
```diff
@@ -50,56 +51,37 @@ const GroupTabDetailsWrapper = ({
...
         <PageSection>
           {canViewHosts && hasEdgeImages ? (
             <Tabs
               className="pf-m-light pf-v5-c-table"
-              activeKey={activeTab && tab == 0 ? activeTab : tab}
-              onSelect={handleTabClick}
+              activeKey={activeSystemsTab}
+              onSelect={(_event, tabIndex) => setActiveSystemsTab(tabIndex)}
               aria-label="Hybrid inventory tabs"
+              unmountOnExit
```
gkarat, line 72: This makes the trick and fixes the bug. Otherwise, more InventoryTable's are mounted (even though not visible) and talk to the same redux store which caused conflicts.

bastilian: Nice catch! Maybe let's also get the components out of the `useMemo`s. They really don't need to be in there.

bastilian: Actually, I suspect that if they are out of the `useMemo` You'll also not need the `unmountOnExit`.

computercamplove:
@gkarat i tested your PR in ephemeral and sorting works.

i started to see another issue: data in "Update method" column in workspace details systems table are not displayed. When i open "Add systems" modal "Update method" column displays data, but after sorting I see "N/A"

computercamplove:
ignore it, there was some issue with uploading hosts - it just uploaded without update method

> i started to see another issue: data in "Update method" column in workspace details systems table are not displayed. When i open "Add systems" modal "Update method" column displays data, but after sorting I see "N/A".
> 
> In stage "Update method" column works as expected.

src/components/InventoryGroupDetail/GroupTabDetails.js
gkarat: I have removed useMemo's, though this unmountOnExit is better kept for now. Without it, I have noticed two instances of InventoryTable rendered for both conventional and immutable tabs. The table rows even more weird when I tried to remove this parameter: the conventional table had rows from immutable tab systems table which again has showed that the settings from `entities` are being used by both table instances. Let's try to keep at maximum one inventory table rendered on any view. We should later revisit the old PR where we tried to isolate stores.

PR #2395, opened 2025-03-19
fix(GroupFilter): RHINENG-16140 - Use deps and memoise properly

bastilian:
This solves the issue described in Jira and ensures the workspace filter is kept up to date with the values available/fetched.

**How to test**

1) Open the Systems table with an account that has more than 10-20 groups
2) Select the "Workspace" filter
3) Verify the filter is properly filled with values.

### DIFF
--- a/src/Utilities/hooks/useFetchBatched.js
+++ b/src/Utilities/hooks/useFetchBatched.js
@@ -1,43 +1,10 @@
+import { useCallback, useMemo } from 'react';
 import usePromiseQueue from './usePromiseQueue';
 
 const useFetchBatched = () => {
   const { isResolving: isLoading, resolve } = usePromiseQueue();
-
-  return {
-    isLoading,
-    fetchBatched: (fetchFunction, total, filter, batchSize = 50) => {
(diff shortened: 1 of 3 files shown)


src/Utilities/hooks/useFetchBatched.js
```diff
@@ -1,43 +1,10 @@
...
-      return results;
-    },
-    pageOffsetfetchBatched: (
-      fetchFunction,
-      total,
-      filter,
-      batchSize = 50,
-      pageOffset = 0
-    ) => {
+  const pageOffsetfetchBatched = useCallback(
```
gkarat, line 6: Since it's got only one `resolve` dependency in this useCallback, can't we just leave the function in the return `useMemo` object as we for fetchBatched and fetchBatchedInline? Does it necessarily need to be wrapped in separate useCallback?

bastilian: > Does it necessarily need to be wrapped in separate useCallback?

Yes, if it is not it will cause the function to always "change" and the `useEffect` is always executed.

AsToNlele (approved):
great job! 😄

PR #2397, opened 2025-03-24
fix(RHINENG-16137): Use host type filter 'nil' when adding hosts

gkarat:
This fixes https://issues.redhat.com/browse/RHINENG-16137. When adding new hosts to workspace, in the Conventional tab, only conventional systems must be listed.

### DIFF
--- a/src/components/InventoryGroupDetail/InventoryGroupDetail.js
+++ b/src/components/InventoryGroupDetail/InventoryGroupDetail.js
@@ -25,7 +25,7 @@ import {
 import useFeatureFlag from '../../Utilities/useFeatureFlag';
 import { hybridInventoryTabKeys } from '../../Utilities/constants';
 import useInsightsNavigate from '@redhat-cloud-services/frontend-components-utilities/useInsightsNavigate';
-import { getHostList } from '../../api/hostInventoryApi';
+import { getHostsCount } from './helpers/getHostsCount';
 
 const GroupDetailInfo = lazy(() => import('./GroupDetailInfo'));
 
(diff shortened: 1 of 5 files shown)


src/components/InventoryGroupDetail/InventoryGroupDetail.js
```diff
@@ -64,48 +60,52 @@ const InventoryGroupDetail = ({ groupId }) => {
...
   // TODO: append search parameter to identify the active tab
 
   const [hasEdgeImages, setHasEdgeImages] = useState(false);
-  const EdgeParityEnabled = useFeatureFlag('edgeParity.inventory-list');
+  const edgeParityEnabled = useFeatureFlag('edgeParity.inventory-list');
 
   useEffect(() => {
-    if (EdgeParityEnabled) {
+    // determines the active tab opened by default
+    const lookUpAvailableSystems = async () => {
```
bastilian, line 77: It would make sense to extract these API calls out of the effect and be a discrete separate function. It would further make sense to have two functions like `getConventionalHostList` and `getImmutableHostList` that perform Api calls with the desired filters set.

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.js
```diff
@@ -163,6 +160,19 @@ const AddSystemsToGroupModal = ({
...
     <InventoryTable
       columns={(columns) => prepareColumns(columns, false, true)}
+      getEntities={async (items, config, showTags, defaultGetEntities) =>
+        await defaultGetEntities(
+          items,
+          {
+            ...config,
+            filters: {
+              ...config.filters,
+              hostTypeFilter: 'nil', // request only conventional systems
```
bastilian, line 170: We could have a `defaultConventionalSystemsGetEntities` that sets the filter. That way it would already explain in the function name what it is/does and it wouldn't need a comment to further explain.

bastilian (approved):
Nice! Thank you for refactoring @gkarat!

There seems to be one small issue. The modal and systems tables seem to still work, but on occasion this error is raised: on the workspace details:

Screenshot 2025-03-25 at 12 33 26

PR #2398, opened 2025-03-28
ci: Add required sast tasks

ezr-ondrej:
This PR is being created to address the new Conforma (Enterprise Contract) requirements we have from April 1st.
Having these tasks included is necessary to keep building with Konflux after that date.

On top of that, I've found out that most of our frontends are cloning our repositories twice for no reason, so I'm commenting the useless cloning tasks :)


ezr-ondrej:
Already implemented ;)

PR #2407, opened 2025-04-01
fix(RHINENG-16137): Fix wrong empty state when no hosts in group

gkarat:
This is an extra PR to fix the group details page wrongly displaying empty state. The empty state was not showing "No systems added" when there were no hosts in such group. This adds "groupName" to one of the useEffect dependencies so that the request to hosts count is done correctly only when group name is available.

## How to test

Open an empty group, make sure the page shows "No groups added" message.

### DIFF
--- a/cypress/support/interceptors.js
+++ b/cypress/support/interceptors.js
@@ -229,6 +229,37 @@ export const hostsInterceptors = {
       }
     );
   },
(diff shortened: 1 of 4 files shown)


hugohezel (approved):

hugohezel:
LGTM, good work @gkarat !

PR #2414, opened 2025-04-08
fix(RHINENG-11371): Clear text input after reseting filters in InfoTable

hugohezel:
Table used in modals (Inventory > System's Details > CPU flags || SAP IDs) doesn't clear search input after removing filters. This PR should fix it.


gkarat (approved):
LGTM, thanks @hugohezel

PR #2415, opened 2025-04-09
fix(RHINENG-17344): Convert class component InfoTable into functional one

hugohezel:
This PR recreates the InfoTable component as a functional component.

### DIFF
--- a/src/components/GeneralInfo/InfoTable/InfoTable.js
+++ b/src/components/GeneralInfo/InfoTable/InfoTable.js
@@ -1,4 +1,4 @@
-import React, { Component, Fragment } from 'react';
+import React, { Fragment, useCallback, useMemo, useState } from 'react';
 import PropTypes from 'prop-types';
 import {
   Pagination,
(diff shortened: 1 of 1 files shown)


src/components/GeneralInfo/InfoTable/InfoTable.js
```diff
@@ -15,172 +15,173 @@ import {
...
+import { filterRows, generateFilters, prepareRows } from '../../../constants';
+
+const InfoTable = (props) => {
+  const {
+    cells = [],
+    rows = [],
+    expandable = false,
+    filters,
+    onSort = undefined,
+  } = props;
```
bastilian, line 27: This is not wrong, but usually the assignment can already be in the arguments assigning like:

```js
const InfoTable = ({
    cells = [],
    rows = [],
    expandable = false,
    filters,
    onSort = undefined,
  }) => {
```

Just having the `prop` argument and using it for assignments would only be useful if we were to further pass on the same exact props to a child component.

src/components/GeneralInfo/InfoTable/InfoTable.js
```diff
@@ -15,172 +15,173 @@ import {
...
+        )
+      : prepareRows(activeRows, pagination);
+  }, [activeRows, pagination, expandable, opened]);
 
-  onSort = (event, index, direction) => {
-    const { expandable } = this.props;
-    this.props.onSort(event, expandable ? index - 1 : index, direction);
-    this.setState({
+  // Handlers
+  const innerOnSort = (event, index, direction) => {
```
bastilian, line 66: We should wrap functions like these here in a `useCallback`.

src/components/GeneralInfo/InfoTable/InfoTable.js
```diff
@@ -15,78 +15,35 @@ import {
...
-  };
-
-  render() {
-    const { cells, rows, expandable, filters } = this.props;
-    const { sortBy, opened, pagination, activeFilters } = this.state;
-    const collapsibleProps = expandable ? { onCollapse: this.onCollapse } : {};
-    const activeRows = filterRows(rows, activeFilters);
-    const mappedRows = expandable
+  const mappedRows = useMemo(() => {
+    return expandable
```
gkarat, line 46:
```suggestion
  const mappedRows = useMemo(() => expandable
  ...
```

gkarat (approved):
LGTM, thanks for dealing with this component, it has been long lasting issue

gkarat:
I have restarted the enterprise-contract job again now

PR #2419, opened 2025-04-14
feat(RHINENG-17382): Convert class component GeneralInformation into functional one

hugohezel:
This PR converts the GeneralInformation component used to display various cards on System's Details page to the functioinal one.

To verify this change - go to System's Details page and verify that all cards (System properties, Infrasctructure, System status etc...) are displayed correctly and its functionalities remained the same (user can open modals, click on links etc...).

### DIFF
--- a/src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
+++ b/src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
@@ -1,7 +1,7 @@
-import React, { Component, Fragment } from 'react';
+import React, { Fragment, useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
-import { Grid, GridItem, Modal } from '@patternfly/react-core';
+import { Grid, GridItem, Modal, ModalVariant } from '@patternfly/react-core';
 import { SortByDirection } from '@patternfly/react-table';
 import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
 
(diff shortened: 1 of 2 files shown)


src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -23,219 +23,210 @@ import useInsightsNavigate from '@redhat-cloud-services/frontend-components-util
...
+  inventoryId,
+  loadSystemDetail,
+  systemProfilePrefetched = false,
+  showImageDetails = false,
+  isBootcHost = false,
+  showRuntimesProcesses = false,
+}) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [modalTitle, setModalTitle] = useState('');
+  const [modalVariant, setModalVariant] = useState('small');
```
gkarat, line 51: nit-pick:

```suggestion
  import { Grid, GridItem, Modal, ModalVariant } from '@patternfly/react-core';
  ...
  const [modalVariant, setModalVariant] = useState(ModalVariant.small);
```

so we take advantage of some constants available from PF/react

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -23,219 +23,210 @@ import useInsightsNavigate from '@redhat-cloud-services/frontend-components-util
...
+          '' + (secondRow[index].sortValue || secondRow[index])
+        ).toLocaleLowerCase();
+        return aSortBy < bSortBy ? -1 : 1;
+      });
+      setModalData((prevData) => ({
+        ...prevData,
+        rows: direction === SortByDirection.asc ? sorted : sorted.reverse(),
+      }));
+    },
+    [modalData.rows]
```
gkarat, line 84: nit-pick: onSort will be re-defined always since modalData.rows is an array, not a primitive, https://react.dev/reference/react/useCallback#parameters, `Object.is([], [])` always results in false.

As an alternative, you could use `useDeepCompareCallback` like we did in compliance: https://github.com/RedHatInsights/compliance-frontend/blob/8c237e4a38955db1d6dad407276eee77e230a354/src/Utilities/hooks/useQuery/useQuery.js#L51 or just leave the function without useCallback since it's directly reliant on the state variable modalData. This is not something critical, but good to consider in a long-term perspective and in terms of code quality.

gkarat (approved):
Works good, thanks. I am approving the PR, but you could consider some of the comments as well

gkarat:
LGTM, ready to merge

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -23,219 +23,208 @@ import useInsightsNavigate from '@redhat-cloud-services/frontend-components-util
...
+  }, [entity.id, inventoryId, loadSystemDetail, systemProfilePrefetched]);
+
+  const onSort = (_event, index, direction, customRows, secondaryIndex) => {
+    const sortAttributes = [
+      (row) => {
+        const value = row[index]?.sortValue || row[index];
+        return String(value).toLocaleLowerCase();
+      },
+    ];
+    if (secondaryIndex !== undefined) {
```
hugohezel, line 75: For better sort consistency, added option to sort by secondary column. Sorting by only one column caused different order of rows (from the other column's perspective) with each sort attempt.

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -23,219 +23,208 @@ import useInsightsNavigate from '@redhat-cloud-services/frontend-components-util
...
+      sortAttributes.push((row) => {
+        const value = row[secondaryIndex]?.sortValue || row[secondaryIndex];
+        return value.toLocaleLowerCase();
```
hugohezel, line 78: For better sort consistency, added option to sort by secondary column. Sorting by only one column caused different order of rows (from the other column's perspective) with each sort attempt.

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -23,219 +23,208 @@ import useInsightsNavigate from '@redhat-cloud-services/frontend-components-util
...
+      });
+    }
+    const toSort = [...(customRows || modalData.rows)];
+    const sorted = sortBy(toSort, sortAttributes);
```
hugohezel, line 82: Changed custom sort logic to the lodash.

src/components/GeneralInfo/GeneralInformation/GeneralInformation.js
```diff
@@ -23,219 +23,208 @@ import useInsightsNavigate from '@redhat-cloud-services/frontend-components-util
...
+  });
+
+  //Avoids duplicate profile fetch if consumer app already fetched, while staying backwards compatible
+  useEffect(() => {
+    if (!systemProfilePrefetched) {
+      loadSystemDetail?.(inventoryId || entity.id);
+    }
+  }, [entity.id, inventoryId, loadSystemDetail, systemProfilePrefetched]);
+
+  const onSort = (_event, index, direction, customRows, secondaryIndex) => {
```
hugohezel, line 68: This PR reacts also to unworking sorting reported in RHINENG-17344.

PR #2420, opened 2025-04-14
feat(RHINENG-17384): Convert class component TextInputModal into functional one

hugohezel:


src/components/GeneralInfo/TextInputModal/TextInputModal.test.js
```diff
@@ -1,48 +1,39 @@
...
-        value: undefined,
+  describe('State management with props', () => {
+    it('should reset value to undefined when modal is closed', () => {
+      const { rerender } = render(<TextInputModal isOpen={true} value="foo" />);
+      rerender(<TextInputModal isOpen={false} value="foo" />);
+
+      const input = screen.queryByRole('textbox', {
+        name: /input text/i,
       });
+      expect(input).not.toBeInTheDocument();
```
gkarat, line 16: This only asserts that the modal is not in the DOM, though the test description says that it tests if the value is reset to undefined when the modal is closed

gkarat (approved):
LGTM

PR #2429, opened 2025-04-25
feat(RHINENG-15545): Add tooltip to Last seen column

leSamo:
https://issues.redhat.com/browse/RHINENG-15545

image


bastilian:
> It would be clearer to extract the custom header rendering into a dedicated presentational component.

That would indeed make sense and keep it a bit cleaner.

leSamo:
@bastilian oki, I isolated it inside a separate component.

PR #2449, opened 2025-05-07
chore: Fix typo in Vulnerability intl provider

leSamo:
This change does not cause any issues, since Vulnerability accommodates both prop variants with and without a typo: https://github.com/RedHatInsights/vulnerability-ui/blob/a2eddb7ca2d811b32d718ee9fc1c0bd2229fd032/src/Components/SmartComponents/SystemCves/SystemCves.js#L287


LightOfHeaven1994 (approved):
lgtm! good catch

ezr-ondrej (approved):
🧡

PR #2451, opened 2025-05-09
fix: Fix tooltip next to "Last seen" header not showing

leSamo:
Fix regression after adding the Last check in feature flag.

## Summary by Sourcery

Fix the tooltip for the "Last seen" column header by introducing a custom column header component

Bug Fixes:
- Resolved tooltip issue for the 'Last seen' column header after adding the Last check in feature flag

Enhancements:
- Updated the 'Last seen' column configuration to use a custom header component


AsToNlele (approved):
image

PR #2453, opened 2025-05-12
chore(deps): Upgrade to ESLint version 9

gkarat:
No jira.

This upgrades ESLint to version 9 and subsequently replaces the no longer supported `.eslintrc.js` configuration file with `eslint.config.js`. The new version also deprecates the use of `.eslintignore`. New version of the FEC ESLint configuration brings a lot of error related to trailing commas or new line symbols. Given that there is no easy way to ignore this rule (prettier config is placed in the library, not in the application code), I have decided to automatically fix these errors and open a separate commit.

No other errors should be fixed with this PR (meaning, it's good to double-check with a review that we do not accidentally fix something more serious).


AsToNlele (approved):
Thanks for the upgrade! Went through the changes and it indeed changes only the trailing commas and removes some disable eslint lines.
Scripts for linting and fixing still work

PR #2460, opened 2025-05-14
fix(TagModal): RHINENG-11908 - Correctly create objects for selected items

bastilian:
This fixes the issue where the selection within the filter toolbar gets lost after applying tags from the TagModal. 

**How to test:**

1) Open the "Systems" page
2) Select a tag in the filter toolbar
3) Open the tag modal via the "Show more" link within the filter
4) Select a new tag in the modal and apply the tag(s)
5) Verify that the current selection is correct in the filter and chips displayed

## Summary by Sourcery

Fix lost filter toolbar selection after applying tags by correcting the structure of selection objects

Bug Fixes:
- Preserve current selection in the filter toolbar after applying tags

Enhancements:
- Generate selection entries with composite keys (`key=value`) and include `tagKey`, `tagValue`, `value`, and `id` fields

Chores:
- Remove stray debug `console.log` from `InventoryTable`

### DIFF
--- a/src/Utilities/constants.js
+++ b/src/Utilities/constants.js
@@ -165,14 +165,17 @@ export const arrayToSelection = (selected) =>
       ...acc,
       [namespace]: {
         ...acc[namespace?.title || namespace],
-        [key?.title || key]: {
+        [`${key?.title || key}=${value?.title || value}`]: {
           isSelected: true,
           group: {
             value: namespace?.title || namespace,
             label: namespace?.title || namespace,
           },
           item: {
-            value: key?.title || key,
+            tagKey: key?.title || key,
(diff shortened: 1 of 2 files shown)


LightOfHeaven1994 (commented):
Adding a comment here for visibility:

PR fixed the issue that has been found earlier. But I spotted another minor issue - reset filters works fine, but if I uncheck one of tags and open show more modal I still see selection

https://github.com/user-attachments/assets/b420bfc7-066f-4c44-a1a6-b6ab1bba51cf

LightOfHeaven1994 (approved):
LGTM, but that I found is super minor and hard to fix. So let's merge this as it is

PR #2477, opened 2025-05-22
feat (RHINENG-11475): inline edit for display name and ansible hostname

xmicha82:
Changed modal editing in system detail to inline editing for Ansible hostname and display name


src/components/GeneralInfo/SystemCard/NameInlineEdit.js
```diff
@@ -0,0 +1,109 @@
...
+              variant="plain"
+              type="button"
+              aria-label="cancel"
+              onClick={() => {
+                setEditingOpen(false);
+                setCurrentValue(textValue);
+              }}
+              style={{
+                display: 'inline',
+              }}
```
AsToNlele, line 84: .

PR #2479, opened 2025-05-26
fix(RHINENG-18216): show 'No workspace' option in Workspace filter

AsToNlele:
[RHINENG-18216](https://issues.redhat.com/browse/RHINENG-18216)

When Kessel FF ("hbi.kessel-migration") is disabled, the workspace filter in Systems page should have first option "No workspace".
When it's enabled, "No workspace" should be hidden in the workspace filter.

## Summary by Sourcery

Show the “No workspace” option in the Systems page workspace filter and update its label

Bug Fixes:
- Rename filter option label from “No group” to “No workspace”
- Enable display of the “No workspace” option via the showNoGroupOption prop in the workspace filter


Fewwy (approved):
LGTM

PR #2486, opened 2025-05-30
fix(RHINENG-18217): add Ungrouped option to group filter

FabriciaDinizRH:
## Summary by Sourcery

Bug Fixes:
- Include the missing 'type' parameter in the getGroupList API call to correctly filter group lists

### DIFF
--- a/src/components/filters/useGroupFilter.js
+++ b/src/components/filters/useGroupFilter.js
@@ -54,7 +54,7 @@ const useGroupFilter = (showNoGroupOption = false, isKesselEnabled = false) => {
     const fetchOptions = async () => {
       if (!hasAccess) return;
 
-      const search = isKesselEnabled ? { type: 'all' } : undefined;
+      const search = isKesselEnabled ? { groupType: 'all' } : {};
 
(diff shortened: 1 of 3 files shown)


src/components/filters/useGroupFilter.js
```diff
@@ -54,18 +54,15 @@ const useGroupFilter = (showNoGroupOption = false, isKesselEnabled = false) => {
...
-        ? await getGroups(search, { page: 1, per_page: 50 })
+        ? await getGroups(undefined, { page: 1, per_page: 50 })
         : { total: 0 };
 
       const groups =
         !ignore && firstRequest.total > 50
           ? await pageOffsetfetchBatched(
               getGroups,
               firstRequest.total - 50,
-              {},
```
ezr-ondrej, line 68: Ugh, this needs to be kept! :)

FabriciaDinizRH:
@johnsonm325 behavior should be fixed now 
image

FabriciaDinizRH:
@computercamplove I don't think so. It was 'Ungrouped hosts' before but now backend is sending it with capital H

PR #2493, opened 2025-06-06
ci(RHINENG-17620): Migrate to shared Konflux pipelines

ezr-ondrej:
It will simplify our pipelines, get us on the same pipeline as all the other images in Insights/ConsoleDot.

⚠️ drawback being it is one extra dependency.


ezr-ondrej:
Getting this in :)

PR #2509, opened 2025-06-19
chore: fix Konflux push pipeline

ezr-ondrej:
Fixing error introduced in https://github.com/RedHatInsights/insights-inventory-frontend/pull/2504

I've fixed a pull-request pipeline, but not push pipeline.


fstavela (approved):
Thanks!

PR #2515, opened 2025-06-24
fix(RHINENG-18805): Add to workspace button disabled

johnsonm325:
This PR fixes an issue where we were returning the incorrect boolean value to determine if the "Add to workspace" button was disabled.

To repro:
- Inventory -> Systems
- Select 1 or more system checkboxes with "Ungrouped hosts" value in the Workspace column
- Click action kebab in the toolbar

You should see that the "Add to workspaces" menu item is disabled. This PR fixes that.

## Summary by Sourcery

Bug Fixes:
- Correct the condition to properly disable the 'Add to workspace' button for ungrouped hosts

### DIFF
--- a/src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.js
+++ b/src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.js
@@ -180,7 +180,7 @@ const ConventionalSystemsTab = ({
 
       if (isKesselEnabled) {
         return selectedHosts.every(
-          ({ groups }) => groups[0].ungrouped !== true,
+          ({ groups }) => groups[0]?.ungrouped === true,
         );
       }
 


src/components/InventoryTabs/ConventionalSystems/ConventionalSystemsTab.js
```diff
@@ -180,7 +180,7 @@ const ConventionalSystemsTab = ({
 
       if (isKesselEnabled) {
         return selectedHosts.every(
-          ({ groups }) => groups[0].ungrouped !== true,
+          ({ groups }) => groups[0].ungrouped === true,
```
ezr-ondrej, line 183: I suppose it should be given, but should we check to make sure that the host is only in the ungrouped group?
```suggestion
          ({ groups }) => groups.length === 1 && groups[0].ungrouped === true,
```

johnsonm325: not a bad idea. I've changed it to `groups[0]?.ungrouped === true`. This will return false if `groups[0]` doesn't exist.

ezr-ondrej: that covers the case for lenght < 0 and avoids errors.
Case where we have 2 groups should not happen I assume? :)

PR #2544, opened 2025-07-16
fix(RHINENG-11337): use default warning patternfly Modal icon

raswonders:
Implements RHINENG-11337

This PR makes DeleteModal component use default patternfly icon instead of custom one

# How to reproduce
1. navigate to HCC for RHEL
2. go to Inventory
3. go to Systems
4. select a system
5. click on DELETE button

### DIFF
--- a/src/Utilities/DeleteModal.js
+++ b/src/Utilities/DeleteModal.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import {
   Button,
   ClipboardCopy,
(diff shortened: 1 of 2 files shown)


bastilian (approved):
LGTM! Thank you for updating this @raswonders!

bastilian:
@computercamplove Yeah, that does make sense. I looked at compliance as well and there we do indeed have a `?` for each of the "Delete.." titles in the modals.

PR #2554, opened 2025-07-23
feat(RHINENG-18834): Update System type filter

computercamplove:
PR to update System type filter and introduce new filter param `system_type`:

1. The new `system_type` filter aims to simplify the distinction between different system types as packaged-based and image-based (bootc and edge) for both API users and the frontend https://github.com/RedHatInsights/insights-host-inventory/pull/2758.
2. Update Bifrost page with new queries
3. Update "System type" filter with new filter param and updated image-based option to get bootc and edge hosts:
image

How to verify changes:
1. Enable `preview-mode` and open the Network tab
2. Visit Bifrost page:

- Get Bootc systems data with `/hosts?system_type=bootc&fields[system_profile]=bootc_status` instead of `/hosts?filter[system_profile][bootc_status][booted][image_digest][is]=not_nil&fields[system_profile]=bootc_status`
- Get Package-based systems total with ` /hosts?system_type=conventional&per_page=1` instead of `/hosts?filter[system_profile][bootc_status][booted][image_digest][is]=nil&filter[system_profile][host_type]=nil&per_page=1`
- Get Immutable systems total with ` /hosts?system_type=edge&per_page=1` instead of: `/hosts?filter[system_profile][host_type]=edge&per_page=1` (note: `[system_profile][host_type]` is still valid, i updated just for the consistency)

3. Filter inventory table with System type filter:
- Package-based systems: `&system_type=conventional`
- Image-based systems:` &system_type=bootc&system_type=edge`

## Summary by Sourcery

Introduce host type constants and new inventory query parameters, update BifrostPage to use the new constants

Enhancements:
- Define new host type constants and corresponding inventory query parameter constants
- Update BifrostPage to use the new inventory query constants for package-based, edge, and bootc systems

## Summary by Sourcery

Introduce a top-level system_type filter parameter and refactor UI, API, and constants to use it for distinguishing package-based and image-based (bootc and edge) systems

New Features:
- Add system_type filter param to host inventory API calls and BifrostPage queries

Enhancements:
- Implement helper utilities to expand and map system_type selections in the UI filter hook
- Update constants to replace legacy nested filters with system_type query parameters
- Refactor useSystemTypeFilter, BifrostPage, edge utilities, and hostInventoryApi to use new system_type-based constants

Build:
- Bump @redhat-cloud-services/host-inventory-client and add eslint-plugin-prettier dependency

Tests:
- Update BifrostPage test to verify calls against new system_type constants


package-lock.json
```diff
@@ -5710,9 +5710,9 @@
       }
     },
     "node_modules/@redhat-cloud-services/host-inventory-client": {
-      "version": "4.0.2",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/host-inventory-client/-/host-inventory-client-4.0.2.tgz",
-      "integrity": "sha512-+b+Rsss2XCo2DIV+I/WSQiMN4QOYNTuLjj7Iw1hZFSmonGXlrPHhNWSGrpIKiCvdJral4IZF0v5t2fUIUHLCIA==",
+      "version": "4.1.0",
```
computercamplove, line 5713: i am not sure about this - this one always appear when we update the package?

computercamplove: i just not sure if i was updating this package-lock, i missed it when i was rebased commits

computercamplove: @bastilian thank you!

src/api/api.js
```diff
@@ -151,7 +151,7 @@ export const calculateSystemProfile = (
     updateMethodFilter,
     hostTypeFilter,
     system_profile,
-    systemTypeFilter,
+    // systemTypeFilter,
```
johnsonm325, line 154: remove?

src/components/filters/helpers.js
```diff
@@ -182,3 +182,31 @@ export const appendGroupSelection = (selection = {}, groups) =>
       })
       .filter(([k, v]) => !!k && !!v),
   );
+// Utility to expand selected values into actual filterable values
+export const getExpandedSystemTypeValues = (
```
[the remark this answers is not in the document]
johnsonm325: @computercamplove You don't need to move this over to the `useSystemTypeFilter` if you don't want. But, I do think you could probably make a couple of the suggested changes in the functions Sourcery provides above to improve the code a bit.

johnsonm325 (changes requested):
All of this looks good. I found one breaking bug though. When you apply the system type filter and then refresh the page, you lose the filter on the table, but the filter remains in the url.

src/api/api.js
computercamplove: oops, going to remove it, thank you!

computercamplove:
@johnsonm325 chip removing after refresh is fixed!

src/Utilities/constants.js
```diff
@@ -127,11 +127,13 @@ export const updateMethodOptions = initUpdateMethodOptions;
...
     label: 'Package-based system',
-    value: 'nil',
+    value: 'conventional',
+    packageBasedValues: ['conventional'],
   },
   {
     label: 'Image-based system',
-    value: 'not_nil',
+    value: 'bootc',
+    imageBasedValues: ['bootc', 'edge'],
```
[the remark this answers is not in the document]
computercamplove: `imageBasedValues` is used to add to the the request, i added it because "Image-based option" apply 2 values in the same time

`value` key is check in `const setValue` - for example i had the issue when page is refreshed - the chip is removed from page even if URL is contains `system_type` values, so it check if at least "`bootc`" is present in there it will save the chip in UI.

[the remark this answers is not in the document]
fstavela: I think it's better the way it is now. If you want image-based, you can combine edge and bootc. But if you want just one of them, you can get it as well. So the more granular approach we have now gives us more options, but doesn't take any option away, so I think it's better.

PR #2569, opened 2025-07-29
feat(InventoryTable): Allow passing in axios instance for requests

bastilian:
This will allow IoP to pass in an axios instance that does not have any interceptors that call `chrome` functions.


src/components/InventoryTable/InventoryTable.js
```diff
@@ -85,6 +85,7 @@ const InventoryTable = forwardRef(
       showCentosVersions = false,
       enableExport,
       lastSeenOverride,
+      axios,
```
Fewwy, line 88: I am testing this in SSH env, should I pass the axios instance into the inventory table from the RHEL Advisor side?

adonispuente:
Can you add instructions for testing this?

PR #2571, opened 2025-07-30
feat(RHINENG-19382): update inventory table for the iop env

Fewwy:
1. Moved feature flag calls to the Routes file, that way we don't have issues with useFeatureFlag being called in the IOP env
2. The feature flags are via useContext and then props wherever they are used

3. I added a prop passed from the RHEL Advisor, "loadChromeless," and attempted to hide the useInventoryExport behind it. This hook is causing an issue because inside of it lives useExportApi, which uses the platform-provided "useAxiosWithPlatformInterceptors();" and contains Chrome.
I need advice on how to refactor that

These changes were tested in the SSH box environment and in the main inventory. If I can fix the issue with useInventoryExport, this PR should be noninvasive and not change the code too much.


src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -251,13 +250,16 @@ const EntityTableToolbar = ({
...
       !hideFilters.systemTypeFilter,
   };
-
-  const exportConfig = useInventoryExport({
-    filters: {
-      ...activeFilters,
-      ...customFilters,
-    },
-  });
+  // I don't know how to fix this yet It calls chrome from the useExportApi
```
bastilian, line 253: What and how is it calling chrome?

Fewwy: @bastilian the useAxiosWithPlatformInterceptors is called in this file src/components/InventoryTable/hooks/useInventoryExport/useExportApi.js

The path is
useInventoryExport in the EntityTableToolbar.js is calling useExportApi(); that calls useAxiosWithPlatformInterceptors();

src/components/InventoryTable/EntityTableToolbar.js
```diff
@@ -251,23 +253,20 @@ const EntityTableToolbar = ({
       !hideFilters.systemTypeFilter,
   };
   // I don't know how to fix this yet It calls chrome from the useExportApi
-  let exportConfig = null;
-  if (!props.loadChromelessInventory) {
-    exportConfig = useInventoryExport({
+  /* const exportConfig = useInventoryExport({
```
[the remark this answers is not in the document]
Fewwy: I will try it. Let me check

bastilian (approved):
LGTM! Nice work! Thank you @Fewwy!

PR #2573, opened 2025-07-31
fix(RHINENG-10851): subscription card not populating

johnsonm325:
When registering a new system with subscription-manager, the subscription card details are populated under the systemProfile details instead of the entityDetails. So this code is updated to read the subscription info from either location.

## Summary by Sourcery

Ensure SubscriptionCard can retrieve subscription data from either entity facts or system_profile system_purpose and update related selectors and tests accordingly.

Bug Fixes:
- Fix subscription card not populating when subscription data is under systemProfile.system_purpose instead of entityDetails

Enhancements:
- Extend subscriptionsSelector to fall back to system_purpose fields when facts are unavailable
- Update SubscriptionCard connector to pass both entity and systemProfile into subscriptionsSelector

Tests:
- Add unit tests for subscriptionsSelector to cover both facts-based and system_purpose-based data
- Update SubscriptionCard tests to verify rendering with system_purpose subscription data
- Rename and separate mock fixtures for facts and system_purpose subscription data

### DIFF
--- a/src/__mocks__/selectors.js
+++ b/src/__mocks__/selectors.js
@@ -68,10 +68,18 @@ export const collectInfoTest = {
   insights_egg_version: 'test-egg',
 };
 
-export const subscriptionsTest = {
+export const subscriptionsTestFacts = {
   facts: {
     SYSPURPOSE_USAGE: 'Development',
     SYSPURPOSE_SLA: 'Self-Support',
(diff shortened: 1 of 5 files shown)


computercamplove:
verified via `ephemeral`

system with system_purpose
image
system with facts
image

PR #2579, opened 2025-08-05
fix(RHINENG-16876): misalignment system table rows

johnsonm325:
Workspaces 'Add systems' modal's table rows not aligning properly while using Firefox browser.

To repro:
- Inventory -> Workspaces
- Click a workspace
- Click the 'Add systems' button to open the modal in the workspace details page

See that the content in the rows is not vertically aligned in the middle. This is because we had a vertical-align: -2 on a specific cell and it was somehow being inherited by all of the other cells.

This is now fixed with this PR.


computercamplove (approved):
LGTM! 

verified in Firefix
image

PR #2588, opened 2025-08-11
feat(fec): Expose SystemTable as federated module

bastilian:


fec.config.js
```diff
@@ -139,6 +139,10 @@ module.exports = {
         __dirname,
         '/src/modules/IOPInventoryTable.js',
       ),
+      './SystemsTable': resolve(
+        __dirname,
+        'src/routes/Systems/components/SystemsTable/SystemsTable.js',
```
ezr-ondrej, line 144: Should we have a wrapper in `modules`? I kinda like the pattern as it's easy to see what components are exposed, but not a deal breaker.

ezr-ondrej (commented):
Should we keep the pattern of exposed modules being wrapped and live under `/modules`?

bastilian:
@ezr-ondrej Added a shim to the modules folder for the `SystemsTable`

PR #2631, opened 2025-09-18
Add workspace test case for create,rename,delete

adarshdubey-star:
https://issues.redhat.com/browse/RHINENG-20815
https://issues.redhat.com/browse/RHINENG-20812
https://issues.redhat.com/browse/RHINENG-20811


adarshdubey-star:
I did not like it myself, just wanted to make sure it works with the pipeline.

PR #2636, opened 2025-09-19
fix(RHINENG-9836): Fix duplicate api call to fetch hosts data

jpramos123:
This PR addresses [RHINENG-9836](https://issues.redhat.com/browse/RHINENG-9836) 
It fixes a duplicate API call that was happening when rendering the InventoryTable.

It also has may resolve [RHINENG-9836](https://issues.redhat.com/browse/RHINENG-9836), that was reporting a delay during the inventory table load.

## Summary by Sourcery

Prevent duplicate API calls in InventoryTable by refactoring the refresh logic

Enhancements:
- Refactor onRefreshData into a useCallback to stabilize its dependencies
- Consolidate loadSystems dispatch into a single path within onRefreshData to avoid redundant requests
- Update useEffect dependencies to include onRefreshData, autoRefresh, and customFilters to prevent unnecessary fetches

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ docs/*
 
 # Auth files
 .auth/
+
+.claude/
+
--- a/src/components/InventoryTable/InventoryTable.js
+++ b/src/components/InventoryTable/InventoryTable.js
@@ -1,6 +1,7 @@
 import React, {
   Fragment,
   forwardRef,
+  useCallback,
   useContext,
   useEffect,
   useRef,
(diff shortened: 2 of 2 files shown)


ezr-ondrej (commented):
This is a great find! 🧡
I think this is actually something we could resolve by using react-query. It does all this by default (handling when to reload) and has it's own cache so elevates the state handling. I'd prefer using that.
We're already adding react-query in https://github.com/RedHatInsights/insights-inventory-frontend/pull/2598.

bastilian:
We tried something that included changes similar to this, and it required more work to make it not break other applications[0] or introduce other odd behaviours.

We should thoroughly test this in apps that use the federate `InventoryTable` module, specially interacting with filters, since interacting with them in apps brought a few issues to light.


[0] https://github.com/RedHatInsights/insights-inventory-frontend/pull/2129

PR #2646, opened 2025-09-24
fix(RHINENG-11434): Update add to workspace modal

johnsonm325:
The 'Add to workspace' modal contains some labeling issues (both conventional and immutable):

'Select a workspace' should be 'Select workspace'
'Create a new workspace' should be 'Create workspace'


bastilian (approved):
LGTM! Thank you @johnsonm325!

PR #2647, opened 2025-09-25
chore( RHINENG-20814): User cannot delete a workspace with systems

adarshdubey-star:
Jira - [RHINENG-20814](https://issues.redhat.com/browse/RHINENG-20814)

## Summary by Sourcery

Tests:
- Add test verifying that deleting a workspace with systems is prevented and shows an error message

### DIFF
--- a/_playwright-tests/test_workspace.test.ts
+++ b/_playwright-tests/test_workspace.test.ts
@@ -24,7 +24,6 @@ test('User can create, rename, and delete a workspace', async ({ page }) => {
 
   // Navigate to the Workspaces page
   await navigateToWorkspacesFunc(page);
-  await expect(page.getByRole('heading', { name: 'Workspaces' })).toBeVisible({ timeout: 100000 });
 
   // Generate a unique workspace name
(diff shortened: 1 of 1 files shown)


_playwright-tests/test_workspace.test.ts
```diff
@@ -86,3 +86,45 @@ test('User can create, rename, and delete a workspace', async ({ page }) => {
...
+         - assignee: addubey
+     */
+    
+      const WorkspaceName = "Workspace_with_systems" 
+      // Close any modals/popups before navigating
+      await closePopupsIfExist(page);
+    
+      // Navigate to the Workspaces page
+      await navigateToWorkspacesFunc(page);
+      await expect(page.getByRole('heading', { name: 'Workspaces' })).toBeVisible({ timeout: 100000 });
```
computercamplove, line 109: `navigateToWorkspacesFunc` already uses `{ timeout: 100000 }` for Workplaces page - i think we don't need it here - it is not necessary here in case it will timeout and fail already in the function. 

I would remove completely this line to apply DRY principle, but in this case to see what we expect in the test case is ok, just timeout not needed. Is there situation when it could timeout 2 times?

PR #2659, opened 2025-10-01
chore: clean warnings from stage error console

raswonders:
It's quite hard to see important stuff in error console when there are too many errors with stack traces. So cleaning some of these errors

- one related to wrong validator
- one complaining about defaultProps.

## Summary by Sourcery

Clean up PropType-related console errors by assigning default parameters in DedicatedAction and correcting the logger PropType in InventoryApp

Bug Fixes:
- Move DedicatedAction default values into function parameters and remove defaultProps to suppress related console warnings
- Change InventoryApp.logger PropType from object to func to eliminate mismatched type warnings

### DIFF
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -38,7 +38,7 @@ const InventoryApp = ({ logger }) => {
 };
 
 InventoryApp.propTypes = {
-  logger: PropTypes.object,
+  logger: PropTypes.func,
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Oh, yes please! It's so annoying to have so many, please do address as many as possible 🧡

PR #2664, opened 2025-10-02
chore(SystemCard): Remove defaultProps and clean-up test errors

ezr-ondrej:
Move defaultProps to js default values.
Remove unecessary set of path.
Clean-up some old props.
Add required parameter to be passed from tests.

### DIFF
--- a/src/components/GeneralInfo/SystemCard/SystemCard.js
+++ b/src/components/GeneralInfo/SystemCard/SystemCard.js
@@ -15,17 +15,17 @@ import { useAddNotification } from '@redhat-cloud-services/frontend-components-n
 const SystemCard = ({
   writePermissions,
   handleClick,
-  hasHostName,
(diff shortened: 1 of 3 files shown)


src/components/GeneralInfo/SystemCard/SystemCard.js
```diff
@@ -207,32 +207,6 @@ const SystemCard = ({
...
-
-        mount_point: PropTypes.string,
-        options: PropTypes.shape({}),
-        type: PropTypes.string,
-      }),
-    ),
-    sapIds: PropTypes.arrayOf(PropTypes.string),
-    systemPurpose: PropTypes.string,
-    cpuFlags: PropTypes.array,
-  }),
```
ezr-ondrej, line 235: These are now coming from store and not passed by props, so I'm cleaning it up.

PR #2670, opened 2025-10-06
refactor(SystemsTable): Improve integrate-ability and customising

bastilian:
This refactor better separates the concerns of the `SystemsTable` and `Systems` component and improve the customisation and integrate-ability in apps. 

* It improves the customisability of columns, filters and options.
* Columns and filter customisation functions are now called with objects, instead of arrays
* Exports/imports of columns & filters, where the `default` export is meant as the default if no options or customisation is provided and the "rest" of the exports is the whole set of available inventory systems table columns or filters
* Filter types are in their own file to allow following a similar pattern as columns & filters
* Options can now also be customised via a function similar to filters & columns
* Remaining props are passed on to allow use of all `TableToolsTable` features for consumers
* It moves table features only for the Inventory Systems page from the `SystemsTable` to the `Systems` component
* It refactors the `Systems` component to be a proper "consumer" of the `SystemsTable`

## Summary by Sourcery

Refactor the Systems listing page to leverage bastilian-tabletools for unified state management, customizable columns/filters/options, and centralized action handling via hooks and providers.

Enhancements:
- Wrap Systems component in TableStateProvider and replace custom table state logic with useItemsData, useFullTableState, and useStateCallbacks.
- Convert SystemsTable into a stateless wrapper around TableToolsTable, using resolveColumns, resolveFilters, and resolveOptions helpers for dynamic configuration.
- Extract default columns, filters, filter types, and options into dedicated helpers/constants, removing inline serialisers and legacy resolver functions.
- Centralize all modals (add/remove hosts, delete, edit display name) in the Systems component and delegate action logic to custom hooks (useTableActions, useDeleteSystems, useEditDisplayName, useToolbarActions).

Tests:
- Adapt SystemsTable unit tests to validate new helper-based defaults and function-based APIs for columns, filters, and options.

## Summary by Sourcery

Refactor SystemsTable and Systems components to decouple concerns, integrate bastilian-tabletools for centralized state and actions, and enable dynamic customization of columns, filters, and options via helper functions and hooks.

New Features:
- Enable function-based customization for table columns, filters, and options with default and named exports of available configurations.

Enhancements:
- Wrap Systems component in TableStateProvider and replace internal table state logic with bastilian-tabletools hooks.
- Convert SystemsTable into a stateless wrapper around TableToolsTable and move default columns, filters, filter types, and options into dedicated helper files.
- Centralize modal and action handling in the Systems component using custom hooks (useTableActions, useDeleteSystems, useEditDisplayName, useToolbarActions).

Tests:
- Adapt SystemsTable unit tests to verify helper-based defaults and function API for columns, filters, and options.

### DIFF
--- a/src/routes/Systems/Systems.js
+++ b/src/routes/Systems/Systems.js
@@ -1,51 +1,184 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { PageSection } from '@patternfly/react-core';
-import InventoryPageHeader from '../InventoryComponents/InventoryPageHeader';
-import SystemsTable from './components/SystemsTable';
 import {
-  filtersSerialiser,
-  sortSerialiser,
-  paginationSerialiser,
-} from './serialisers';
-import * as defaultColumns from './components/SystemsTable/columns';
-import * as defaultFilters from './components/SystemsTable/filters';
+  TableStateProvider,
+  useItemsData,
+  useFullTableState,
+  useStateCallbacks,
+} from 'bastilian-tabletools';
 
-const Systems = () => (
-  <>
-    <InventoryPageHeader />
-    <PageSection>
-      <SystemsTable
-        columns={[
-          defaultColumns.displayName,
-          defaultColumns.workspace,
-          defaultColumns.tags,
-          defaultColumns.operatingSystem,
-          defaultColumns.lastSeen,
-        ]}
-        filters={{
-          filterConfig: [
-            defaultFilters.displayName,
-            defaultFilters.statusFilter,
-            defaultFilters.dataCollector,
-            defaultFilters.rhcStatus,
-            defaultFilters.tags,
-            defaultFilters.systemType,
-            defaultFilters.operatingSystem,
-          ],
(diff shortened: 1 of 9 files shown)


src/routes/Systems/Systems.js
```diff
@@ -1,51 +1,184 @@
...
+            actions: toolbarActions,
+            actionResolver: tableActions,
+            onSelect: true,
+          })}
+          toolbarProps={{
+            exportConfig,
+          }}
+        />
+
+        {addHostGroupModalOpen && (
```
bastilian, line 136: We could refactor this in the future and use `Outlet`[0] here and have the modals be sub-routes of `/systems`.

[0] https://reactrouter.com/6.30.1/components/outlet

src/routes/Systems/components/SystemsTable/constants.js
```diff
@@ -1,6 +1,17 @@
+// TODO Move serialisers.js within SystemsTable directory
```
bastilian, line 1: We pass the default `items` function to fetch systems from the Inventory API in the `SystemsTable`, so this should also be the place where we pass the default serialisers for it.

I didn't move the file yet, but we should at some point.

PR #3032, opened 2026-04-02
feat(RHINENG-25526): disable edit name icon in system details

johnsonm325:
## Jira
[RHINENG-25526](https://redhat.atlassian.net/browse/RHINENG-25526)

## What
We need to properly disable the edit display name and ansible hostname button icons when kessel edit permissions are false.

Also added a no permissions state for System Details when the user doesn't have view permissions.

## How
This PR helps consolidate some of the logic around Kessel permissions checks in order to implement this change. It moves the logic out of the `useCanDeleteHostDetails` into new `useHostDetailKesselPermissions` to properly handle multiple permissions checks on System Details page.

Also consolidated the edit button into one button.

## Testing
Manually tested against 2 different accounts, one with edit permissions and one without.

[RHINENG-25526]: https://redhat.atlassian.net/browse/RHINENG-25526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Consolidate Kessel-based host permissions for system details and apply them consistently to view, edit, and delete actions.

New Features:
- Add a bulk Kessel permission hook for system details that checks view, update, and delete capabilities for a host.
- Introduce a no-permissions state on the system details page when a user cannot view a host under Kessel.

Bug Fixes:
- Ensure edit display name and Ansible hostname controls are properly disabled when users lack write permissions as determined by Kessel or RBAC.
- Correct the workspace move confirmation message to avoid showing an incorrect "from" phrase when a system has no current group.

Enhancements:
- Refactor the delete-permission hook on system details to delegate to the new consolidated Kessel permissions hook.
- Unify the system details edit control into a single button component that centralizes enable/disable and tooltip behavior based on permissions.
- Propagate write-permission state from inventory routing through inventory detail and application details to feature components.
- Improve handling of invalid system detail requests by showing a consistent error state for 400 responses.
- Minor formatting and type cleanups in shared UI components and filters.

Tests:
- Add unit tests for the new system details Kessel permissions hook and update delete-permission tests to use it.


src/Utilities/hooks/useCanDeleteHostDetails.ts
```diff
@@ -1,41 +1,13 @@
...
 /**
  * Single-resource Kessel delete permission check for the system details page.
- * Only runs when the Kessel migration feature flag is enabled; otherwise returns
- * undefined so the caller uses RBAC (useConditionalRBAC) for delete permission.
+ * Delegates to {@link useHostDetailsKesselPermissions} (same bulk request as view/update).
  *
  *  @param hostId - Current system/host ID (e.g. entity?.id). When empty and Kessel is on, returns no permission.
  *  @returns      canDelete: true when allowed, false when denied or loading; undefined when Kessel is off (use RBAC). isLoading only true when Kessel is on and check is in progress.
  */
 export const useCanDeleteHostDetails = (hostId: string | undefined) => {
```
raswonders: This hook seems quite shallow and its not being used anywhere but the tests, I'm thinking what is the purpose of it?

raswonders (changes requested):
I'd like to see useCanDeleteHostDetails refactored away or at least have explanation why we need it. 
Other than that I'm fine with stuff, but shared couple of possible improvements.

raswonders (approved):
lgtm

PR #3034, opened 2026-04-02
fix(RHINENG-22401): change page in Tags Modal on filter

johnsonm325:
## Jira
[RHINENG-22401](https://redhat.atlassian.net/browse/RHINENG-22401)

## What
Filtering by name on the System Tag modal is not behaving as expected. Things are fine when screen is on page 1, but filtering stops working on page 2.

When applying the Tags filter, it should put you back on page 1.

## Testing
- go to Inventory -> Systems.
- select Tags filter in the filter dropdown.
- scroll to the bottom of the the tags dropdown and select the link to view more tags. this should open the Tags Modal.
- change the pagination so you have at least 2 pages of tags.
- go to page 2.
- filter the table by a string that forces there to only be tags on page 1, but not page 2.

You'll notice that you get an empty state in error even there are tags in the table. This PR updates the page to page 1 whenever a filter is applied to the table.

[RHINENG-22401]: https://redhat.atlassian.net/browse/RHINENG-22401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Ensure tag filtering in the Tags modal correctly resets to the first page and reliably applies tag changes to the inventory table.

Bug Fixes:
- Reset the Tags modal pagination to page 1 whenever the in-modal tag search filter changes so filtered results are visible from the first page.
- Use an immediate data refresh instead of a shared debounced refresh when applying tag filters from the Tags modal to avoid pagination staying on a previous page.

Tests:
- Add a test verifying that changing the in-modal tag search requests tags starting from page 1 with the current page size.

### DIFF
--- a/src/Utilities/TagsModal.js
+++ b/src/Utilities/TagsModal.js
@@ -187,7 +187,11 @@ const TagsModal = ({
           filterValues: {
             value: filterBy,
             onChange: (_e, value) => {
-              debouncedFetch(pagination, value);
+              const nextPagination = {
+                ...(pagination || PAGINATION_DEFAULT),
+                page: 1,
+              };
+              debouncedFetch(nextPagination, value);
               setFilterBy(value);
             },
           },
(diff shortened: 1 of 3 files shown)


raswonders (approved):
/lgtm modal resets to first page as expected

PR #3042, opened 2026-04-08
feat(RHINENG-24137): SystemsView: add OperatingSystemsFilter

raswonders:
## Jira
Implements RHINENG-24137

## What
Adds Operating Systems filter to SystemsView. It supports the bulkSelection for group items so be sure to test that. I thinkg I got it right but considering current version in InventoryTable doesn't work quite well its hard to say. 

## How
We created brand new filter component which is based on custom MenuItem, the PF SelectOption and MenuItem components couldn't be used for items as they don't support indeterminate (dash) state of checkbox

## Testing
1. enable SystemsView by running localStorage.setItem("ui-inventory-views", "true")
2. go to Inventory / Systems
3. Select Operating Systems from filters dropdown
4. try stuff

!!!Beware sync of set filters to/from URL params currently doesn't work in SystemsView its a new bug which needs to be investigated

## Summary by Sourcery

Add an Operating Systems filter to SystemsView and wire it into the systems query via system profile filters.

New Features:
- Introduce an Operating Systems multiselect filter component with support for grouped major/minor versions and indeterminate group selection.
- Expose an operating_system field in SystemsView filters and map selected values into the host list system_profile.operating_system API filter.

Enhancements:
- Refine operating systems data fetching hook to expose fetch state needed by the new filter UI.
- Centralize shared SystemsView UI constants for filter dropdown sizing and loader identifiers.

Tests:
- Add unit tests for transforming selected OS tokens into the system profile operating_system filter structure used by the API.

## Summary by Sourcery

Add an operating system multiselect filter to SystemsView and wire its selections into the host list system profile query.

New Features:
- Introduce an OperatingSystemsFilter component that fetches OS version data and provides grouped, indeterminate-checkbox multiselect UI for major and minor versions.
- Expose an operating_system field in SystemsView filters and surface it through the filters toolbar.

Enhancements:
- Convert selected OS tokens into the system_profile.operating_system filter shape for the host list API and merge it with existing system profile filters.
- Refine the operating systems query hook to expose fetch state flags used by the new filter UI and centralize shared SystemsView UI constants.

Tests:
- Extend operating system select options tests to cover building the system profile filter from selected OS tokens.

### DIFF
--- a/src/components/SystemsView/filters/OperatingSystemsFilter.tsx
+++ b/src/components/SystemsView/filters/OperatingSystemsFilter.tsx
@@ -14,8 +14,8 @@ import xor from 'lodash/xor';
 import { useOperatingSystemsQuery } from '../hooks/useOperatingSystemsQuery';
 import {
   buildOperatingSystemSelectGroups,
+  buildOsFilterTokens,
   mapOperatingSystemApiResultsToVersionRows,
-  type OperatingSystemSelectGroup,
 } from '../utils/operatingSystemSelectOptions';
 import { FILTER_DROPDOWN_WIDTH } from '../constants';
 
@@ -24,9 +24,6 @@ interface OperatingSystemsFilterProps {
   value?: string[];
   onChange?: (event?: React.MouseEvent, values?: string[]) => void;
 }
-
-const groupTokens = (group: OperatingSystemSelectGroup): string[] =>
-  group.items.map((item) => `${group.value}:${item.value}`);
 
 /* PF MenuItem doesn't support indeterminate checkbox state,
  so we've re-created custom OsFilterOption which does */
@@ -94,11 +91,8 @@ export const OperatingSystemsFilter = ({
     </MenuToggle>
   );
 
-  const notifyChange = (
-    event: React.FormEvent<HTMLInputElement>,
-    next: string[],
-  ) => {
-    onChange?.(event as unknown as React.MouseEvent, next);
+  const notifyChange = (next: string[]) => {
+    onChange?.(undefined, next);
   };
 
   return (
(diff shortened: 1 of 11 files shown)


computercamplove (approved):
LGTM!

I verified locally:
- filtering by OS
- filtering by major OS
- unselect/ clear filter
- combined with other filters

The only small issue is filter is not visible in URL when applied, assume it is the issue you warned me about

src/components/SystemsView/filters/OperatingSystemsFilter.tsx
```diff
@@ -0,0 +1,184 @@
...
+            isChecked={isChecked}
+            onChange={onCheckboxChange}
+            aria-label={text}
+          />
+        </span>
+        <span className={css(menuStyles.menuItemText)}>{text}</span>
+      </span>
+    </label>
+  </li>
+);
```
johnsonm325, line 66: You can nest a checkbox inside of the MenuItem if you find that more useful to show an indeterminate checkbox state:
```
<Menu onSelect={onSelect} selected={selectedItems}>
      <MenuContent>
        <MenuList>
          <MenuItem itemId={0}>
            <Checkbox label='Checkbox 1' isChecked={null}/>
          </MenuItem>
        </MenuList>
      </MenuContent>
    </Menu>;
```

johnsonm325 (approved):
Feel free to use the `MenuItem` like so, if you want to.

src/components/SystemsView/filters/OperatingSystemsFilter.tsx
raswonders: Oh thanks. This is ai generated from big part, so my knowledge isn't as deep but your suggestion would require we give up on Select -> SelectList -> SelectOption markup right? I'm not sure if I'd be ok with that. 

All other filters use Select component so I'd like to keep it similar as much as possible.
Second thing I'm not sure how exactly we would implement that wiring for selection & indeterminate state without Select's API. 

Hence I'd keep this as is. We could experiment with this further when we decide create more minimalistic GroupComponent but for single filter it seems appropriate as is.

PR #3050, opened 2026-04-10
fix(RHINENG-25727): SystemsView: URL params for filters aren't same as legacy ones

raswonders:
## Jira
Fixes RHINENG-25727

## What
- makes URL param key=values for _OperatingSystem, Workspace and LastSeen_ filters consistent with legacy table

## Testing
1. Go to Inventory / Systems and make sure on Preview is on.
2. Set some of the values for above mentioned filters
3. Observe key=value pairs in URL are same in both SystemsView and InventoryTable

image

image

### DIFF
--- a/src/components/SystemsView/DataViewFiltersContext.tsx
+++ b/src/components/SystemsView/DataViewFiltersContext.tsx
@@ -1,6 +1,19 @@
-import React, { createContext, useContext, useMemo } from 'react';
+import React, {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { useDataViewFilters } from '@patternfly/react-data-view';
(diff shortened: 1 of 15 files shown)


computercamplove (approved):

src/components/SystemsView/DataViewFiltersContext.tsx
```diff
@@ -52,20 +69,55 @@ export const DataViewFiltersProvider = ({
...
+
+  const filters = useMemo(
+    () => ({
+      ...rawFilters,
+      last_seen: (rawFilters.last_seen ?? []).filter(isLastSeenKey),
+    }),
+    [rawFilters],
+  );
+
+  //TODO can we infer this from state during render?
```
computercamplove: @raswonders do you want to have jira card for it?

raswonders: good catch wanted to investigate this, just forgot

PR #3058, opened 2026-04-14
test(RHINENG-25741): enable tests on SystemsView table with Kessel changes

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-25741

## What
Added a CI workflow to run E2E tests. These tests specifically validate the integration between `SystemsView` and `Kessel` components within the Systems page to ensure data consistency and UI stability.

## Why
Since `SystemsView` and `Kessel` is planned to be enabled in Prod Preview while the Stable tier continues to use the `InventoryTable` component, these tests ensure that both table implementations remain functional and regression-free.

## How
To avoid flaky tests caused by account-level preview settings, I added a Playwright init script to toggle the view via `localStorage`. This ensures CI isolation and allows parallel testing without affecting the account's global state.

## Summary by Sourcery

Enable Playwright E2E tests to run reliably in both classic and preview inventory UIs and align table metadata for automated interactions.

New Features:
- Support running E2E inventory systems tests against the preview UI via the E2E_PREVIEW flag.

Bug Fixes:
- Stabilize preview mode toggling to reliably turn preview on or off before tests run.
- Fix flaky selectors in systems table actions and bulk select by using more robust labels and OUIA attributes.
- Relax sort-URL expectations to handle both legacy and new sort parameter formats.

Enhancements:
- Move the system-details edit/delete test into the dedicated System Details test suite.
- Add consistent data-label metadata for systems table headers and cells, including an explicit Actions column label.
- Set clearer labels and OUIA IDs for export and bulk delete toolbar actions.
- Rename the Systems View page title in the inventory views route to 'Systems' for consistency.

### DIFF
--- a/_playwright-tests/auth.setup.ts
+++ b/_playwright-tests/auth.setup.ts
@@ -1,12 +1,14 @@
 import { expect, test as setup } from '@playwright/test';
 import {
   ensureNotInPreview,
+  enableSystemsViewAndKessel,
   logInWithUser1,
   storeStorageStateAndToken,
   throwIfMissingEnvVariables,
   closePopupsIfExist,
   closeCookieBanner,
 } from './helpers/loginHelpers';
+import { isSystemsViewEnabled } from './helpers/constants';
 
 setup.describe('Setup', async () => {
   setup.describe.configure({ retries: 3 });
@@ -18,11 +20,14 @@ setup.describe('Setup', async () => {
   setup('Authenticate', async ({ page }) => {
     setup.setTimeout(120_000);
 
+    if (isSystemsViewEnabled) {
+      await enableSystemsViewAndKessel(page);
+    }
+
     await closePopupsIfExist(page);
     await logInWithUser1(page);
     await closeCookieBanner(page);
-    await storeStorageStateAndToken(page);
-
     await ensureNotInPreview(page);
+    await storeStorageStateAndToken(page);
   });
 });
--- a/_playwright-tests/auth.setup.ts
+++ b/_playwright-tests/auth.setup.ts
@@ -1,12 +1,14 @@
 import { expect, test as setup } from '@playwright/test';
 import {
   ensureNotInPreview,
+  enableSystemsViewAndKessel,
   logInWithUser1,
   storeStorageStateAndToken,
   throwIfMissingEnvVariables,
   closePopupsIfExist,
   closeCookieBanner,
 } from './helpers/loginHelpers';
+import { isSystemsViewEnabled } from './helpers/constants';
 
 setup.describe('Setup', async () => {
   setup.describe.configure({ retries: 3 });
@@ -17,6 +19,11 @@ setup.describe('Setup', async () => {
 
   setup('Authenticate', async ({ page }) => {
     setup.setTimeout(120_000);
+
+    if (isSystemsViewEnabled) {
+      await enableSystemsViewAndKessel(page);
+    }
+
     await closePopupsIfExist(page);
     await logInWithUser1(page);
     await closeCookieBanner(page);
--- a/_playwright-tests/auth.setup.ts
+++ b/_playwright-tests/auth.setup.ts
@@ -1,14 +1,12 @@
 import { expect, test as setup } from '@playwright/test';
 import {
   ensureNotInPreview,
-  enableSystemsViewAndKessel,
   logInWithUser1,
   storeStorageStateAndToken,
   throwIfMissingEnvVariables,
   closePopupsIfExist,
   closeCookieBanner,
 } from './helpers/loginHelpers';
-import { isSystemsViewEnabled } from './helpers/constants';
 
 setup.describe('Setup', async () => {
   setup.describe.configure({ retries: 3 });
@@ -19,11 +17,6 @@ setup.describe('Setup', async () => {
 
   setup('Authenticate', async ({ page }) => {
     setup.setTimeout(120_000);
-
-    if (isSystemsViewEnabled) {
-      await enableSystemsViewAndKessel(page);
-    }
-
     await closePopupsIfExist(page);
     await logInWithUser1(page);
     await closeCookieBanner(page);
--- a/_playwright-tests/helpers/filterHelpers.ts
+++ b/_playwright-tests/helpers/filterHelpers.ts
@@ -22,7 +22,12 @@ export const filterSystemsWithConditionalFilter = async (
 ) => {
   let optionCheckbox: Locator | undefined = undefined;
   // 1. SELECT FILER
-  await page.getByRole('button', { name: 'Conditional filter toggle' }).click();
+  await page
+    .locator(
+      '[data-ouia-component-id="DataViewFilters"] button.pf-v6-c-menu-toggle',
+    )
+    .or(page.getByRole('button', { name: 'Conditional filter toggle' }))
+    .click();
   // Wait for the filter menu to be visible before proceeding
   await page
     .getByRole('menuitem', { name: filterName })
@@ -66,7 +71,15 @@ export const filterSystemsWithConditionalFilter = async (
     await expect(optionCheckbox).toBeVisible({ timeout: 100000 });
     await optionCheckbox.click();
   } else if (filterName === 'Operating system') {
-    await page.getByRole('button', { name: 'Group filter' }).nth(1).click();
+    await page
+      .getByRole('button', { name: 'Group filter' })
+      .nth(1)
+      .or(
+        page.locator(
+          '[data-ouia-component-id="SystemsViewOperatingSystemsFilter"]',
+        ),
+      )
+      .click();
     await page.getByRole('checkbox', { name: option, exact: true }).check();
   } else if (filterName === 'Last seen') {
     const lastSeenToggle = page.locator('button', {
@@ -81,12 +94,9 @@ export const filterSystemsWithConditionalFilter = async (
 
   await page.mouse.click(0, 0); //to make sure menu is closed
   // wait for table to be filtered
-  await expect(
-    page.locator('[data-ouia-component-id="SkeletonTable"]'),
-  ).toBeVisible();
   await page
     .locator('[data-ouia-component-id="SkeletonTable"]')
-    .waitFor({ state: 'hidden' });
+    .waitFor({ state: 'detached', timeout: 10000 });
 };
 
 /**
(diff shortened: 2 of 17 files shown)


_playwright-tests/test_system.test.ts
```diff
@@ -293,8 +232,8 @@ test('User should be able to delete multiple systems from Systems page', async (
 
       // Expected URL sort parameter: ascending = display_name, descending = -display_name
       const expectedSortParam = {
-        ascending: 'sort=display_name',
-        descending: 'sort=-display_name',
+        ascending: /sort=display_name|order_how=asc/,
+        descending: /sort=-display_name|order_how=desc/,
```
computercamplove, line 252: @raswonders while updating tests i found this - when we sort by specific column - Inventory table uses `sort=` and SystemsView `order_how=` - do we keep it or we need update it?

src/components/SystemsView/hooks/useRows.tsx
```diff
@@ -23,15 +24,20 @@ export const useRows = ({
...
+          dataLabel: SYSTEMS_VIEW_COLUMN_DATA_LABELS[col.key] ?? col.key,
+        },
+      }));
 
     return {
       id: system.id,
       row: [
         ...selectableColumnCells,
         {
           cell: <SystemsViewRowActions system={system} />,
```
computercamplove, line 33: @raswonders is it ok if i added `data-label` to the rows ? with it easier to locate while testing,

_playwright-tests/test_system.test.ts
raswonders: I'll create JIRA and try to fix it in SystemsView today

computercamplove: in URL

src/components/SystemsView/hooks/useRows.tsx
raswonders: What type of access you're trying to replicate from InventoryTable?

1. Wouldn't cell-based ouia which are already generated be enough?

image

computercamplove: removing the data label

_playwright-tests/helpers/filterHelpers.ts
```diff
@@ -66,7 +71,16 @@ export const filterSystemsWithConditionalFilter = async (
     await expect(optionCheckbox).toBeVisible({ timeout: 100000 });
     await optionCheckbox.click();
   } else if (filterName === 'Operating system') {
-    await page.getByRole('button', { name: 'Group filter' }).nth(1).click();
+    // await page.getByRole('button', { name: 'Group filter' }).nth(1).click();
```
raswonders: needs cleanup

_playwright-tests/auth.setup.ts
```diff
@@ -18,11 +20,14 @@ setup.describe('Setup', async () => {
   setup('Authenticate', async ({ page }) => {
     setup.setTimeout(120_000);
 
+    if (process.env.SYSTEMS_VIEW === 'true') {
```
raswonders: process.env.SYSTEMS_VIEW === 'true' this is repeated on many places and probably could be refactored to boolean isSystemsViewEnabled

_playwright-tests/helpers/filterHelpers.ts
```diff
@@ -81,12 +95,18 @@ export const filterSystemsWithConditionalFilter = async (
...
-  await expect(
-    page.locator('[data-ouia-component-id="SkeletonTable"]'),
-  ).toBeVisible();
-  await page
-    .locator('[data-ouia-component-id="SkeletonTable"]')
-    .waitFor({ state: 'hidden' });
+  if (process.env.SYSTEMS_VIEW === 'true') {
+    await page
+      .locator('table[data-ouia-component-id="loading-state"]')
+      .waitFor({ state: 'detached', timeout: 10000 });
```
raswonders: it's not clear for me why we're waiting for something different than in InventoryTable case. I'd try to make it same and if it cannot be, I'd use comment to explain why.

_playwright-tests/test_system.test.ts
```diff
@@ -328,12 +286,15 @@ test('User should be able to delete multiple systems from Systems page', async (
...
       await expect(async () => {
         const url = page.url();
-        expect(url).toContain(expectedSortParam[order]);
+        expect(url).toMatch(expectedSortParam[order]);
       }).toPass({ timeout: 5000 });
     });
 
     await test.step('Verify systems are sorted alphabetically', async () => {
-      const nameCell = page.locator('td[data-label="Name"] a');
+      // const nameCell = page.locator('td[data-label="Name"] a');
```
raswonders: needs cleanup

_playwright-tests/test_systems_filter.test.ts
```diff
@@ -257,8 +264,14 @@ test.describe('Filtering Systems Tests', () => {
       }
 
       // Open conditional filter dropdown and select "Last seen"
+      // await page
```
raswonders: needs cleanup

raswonders (changes requested):
Thank you for your work. I've tested it locally with SYSTEMS_VIEW=true, it passes. Code looks good overall but I have these few pointers. 

- There are some commented code snippets which should be cleaned up
- We use `process.env.SYSTEMS_VIEW === 'true'` on many places so it make sense to refactor into boolean.
- In more than one test, where split logic happens, its not clear for me why we change the testing approach (don't mean locator), it would help understanding if its explained in a comment

_playwright-tests/helpers/filterHelpers.ts
computercamplove: @raswonders let me check if you rewrote the ouia for SkeletoTable component 
```
<SkeletonTable
                    ouiaId="loading-state"
                    isSelectable
                    rowsCount={pagination.perPage}
                    columns={tableHeaderNodes}
                  />
```

computercamplove: @raswonders https://www.patternfly.org/extensions/component-groups/skeleton-table/#skeletontable

PF uses already OUIA, i am going to remove yours in SystemsView, nice catch.

raswonders: heh I wasn't really pointing at locator here but happy, it solved something, I was more curious about that 'detached'

computercamplove: i decided use 'detached' to be completely sure loading state is removed from HTML

computercamplove: asked AI :)

The Golden Rule: Use detached specifically when you are testing the removal of an object (cleaning up the DOM). Use hidden when you just want to make sure the user can't see or click it anymore.

let's i put back `hidden`:)

computercamplove:
@raswonders could you please point me this cases? (removed from fixture)

> In more than one test, where split logic happens, its not clear for me why we change the testing approach (don't mean locator), it would help understanding if its explained in a comment

_playwright-tests/helpers/filterHelpers.ts
```diff
@@ -129,10 +139,19 @@ export const expectAllRowsHaveText = async (
...
  */
 export const searchByName = async (page: Page, name: string): Promise<void> => {
-  const searchInput = page.locator('input[placeholder="Filter by name"]');
-  await page.reload({ waitUntil: 'networkidle' });
-  await expect(searchInput).toBeVisible();
+  const searchInput = page
+    .locator('input[placeholder="Filter by name"]')
+    .or(page.getByRole('textbox', { name: /display name filter/i }));
+  if (process.env.SYSTEMS_VIEW !== 'true') {
+    await page.reload({ waitUntil: 'networkidle' });
```
raswonders: why do we await networkIdle on systemsView?

_playwright-tests/test_system.test.ts
```diff
@@ -34,12 +37,18 @@ test('User should be able to edit and delete a system from Systems page', async
       .getByLabel('Kebab toggle')
       .click();
 
-    await page
-      .getByRole('menuitem', { name: 'Edit display name' })
-      .nth(1)
-      .click();
-
+    if (isSystemsViewEnabled) {
+      const editButton = page.getByRole('menuitem', { name: /^\s*Edit\s*$/i });
```
raswonders: why do we check button is enabled in here but on not Inventory Table. In this case Locator is also weirdly similar but different one is regex and other text

_playwright-tests/test_system.test.ts
```diff
@@ -238,7 +185,14 @@ test('User should be able to delete multiple systems from Systems page', async (
      - importance: critical
    */
   const dialog = page.locator('[role="dialog"]');
-  const nameCell = page.locator('td[data-label="Name"]');
+  let nameCell: Locator;
+  if (isSystemsViewEnabled) {
```
raswonders: other places use or() can we do the same thing here?

_playwright-tests/test_systems_filter.test.ts
```diff
@@ -290,10 +300,12 @@ test.describe('Filtering Systems Tests', () => {
     });
 
     await test.step('Verify URL contains correct filter parameter', async () => {
-      await expect(async () => {
-        const url = page.url();
-        expect(url).toContain('last_seen=last24');
-      }).toPass({ timeout: 5000 });
+      if (process.env.SYSTEMS_VIEW !== 'true') {
```
raswonders: we should be using isSystemsViewEnabled boolean here as well

raswonders (changes requested):
did another pass as requested and pointed out problems, thanks!

_playwright-tests/helpers/filterHelpers.ts
raswonders: good I'm not agains 'detached', just curious if we could remove unnecessary branching

_playwright-tests/test_system.test.ts
computercamplove: @raswonders i check it because compared to **InventoryTable** it is sometimes disabled when test open too fast (i wasn't able to reproduce it manually. i will ping you in the thread.

I need to fix regex - it left from previews my implementation

_playwright-tests/helpers/filterHelpers.ts
computercamplove: @raswonders we don't do await on  SYSTEMS_VIEW , here !==, but i need to update it anyway to` !isSystemsViewEnabled`

it is for `InventoryTable` only , it was known issue while table refresh, it was here before

computercamplove: i still wait here, hmm let me check if for SYSTEMS_VIEW i can just check for SkeletonTable not visible 

`await page.reload({ waitUntil: 'load' });`

the reason the issue when you fill into input while table is loading

raswonders: I'm trying to investigate why or if we do need branching... why not check same thing for both tables. Branching = bad

_playwright-tests/test_system.test.ts
raswonders: if its necessary to await button to be enabled on SystemsView let's await the same thing in InventoryTable and lose the branching

_playwright-tests/helpers/filterHelpers.ts
computercamplove: i left how it was without if condition

raswonders (approved):
I think it looks good now, thanks for the updates!

PR #3067, opened 2026-04-16
fix: Fix playwright setup steps

computercamplove:
## Jira
Fix playwright setup steps

## What
use cache step
update playwright dependencies step in CI

## Why
Ubuntu incident https://status.canonical.com/
Take a lot time to fetch ubuntu packages

## Summary by Sourcery

Update Playwright CI workflows to use dependency caching and a more targeted Playwright installation command.

CI:
- Add node_modules and dist caching to Playwright-related GitHub Actions workflows to speed up CI runs.
- Change Playwright installation in CI to install only the Chromium browser using the shell-only option across relevant workflows.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -74,7 +74,7 @@ jobs:
         run: npm ci
 
       - name: Install Playwright
-        run: npx playwright install chromium --only-shell
+        run: npx playwright install chromium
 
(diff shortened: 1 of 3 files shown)


raswonders:
only-shell isn't fully capable browser why do we install this instead of normal chromium?
npx playwright install chromium  wouldn't work?

raswonders (approved):

PR #3070, opened 2026-04-17
fix(RHINENG-25801): legacy sort URL param not applied in SystemsView

raswonders:
## Jira
RHINENG-25801

## What
When we reuse url link from legacy table the sorting wouldn't get applied in SystemsView. This PR helps to parse sorging from legacy url and apply it in new table 

## Why
We need this so legacy bookmarks won't break for customer

## Testing
Try to load up legacy InventoryTable & set sorting to some column. Then reload same URL with SystemsView enabled. Verify same sorting state is reflected in the new table

## Summary by Sourcery

Ensure SystemsView correctly interprets and applies legacy sort URL parameters while aligning sort query param names with the new view.

Bug Fixes:
- Apply legacy single-parameter sort URLs (with optional leading dash for descending) to SystemsView so existing bookmarks preserve sorting.

Enhancements:
- Introduce a reusable utility to normalize legacy sort URL parameters into separate sort and direction parameters, configurable by query param name.
- Standardize SystemsView sort-related query parameter names via shared constants and simplify sort parsing logic.
- Add unit tests covering legacy sort URL normalization across various parameter combinations.

Tests:
- Add dedicated tests for the legacy sort search parameter normalization helper to verify behavior for ascending, descending, mixed, and custom parameter cases.

## Summary by Sourcery

Preserve and apply sorting state from legacy inventory table URLs to the SystemsView data table so old bookmarks continue to work.

New Features:
- Add normalization of legacy single-parameter sort URLs into separate sort and direction parameters for SystemsView and other consumers of useDataViewSort.

Bug Fixes:
- Ensure SystemsView correctly interprets and applies legacy sort URLs that use a leading dash to indicate descending order.

Enhancements:
- Introduce shared sort URL parameter constants for SystemsView and use a memoized, normalized view of search parameters when configuring table sorting.
- Refine global search parameter parsing logic to more clearly distinguish leading-dash descending sort syntax from hyphens inside field keys.

Tests:
- Add unit tests covering legacy sort URL normalization, including descending, ascending, mixed-hyphen keys, two-parameter sort URLs, URLs without sort, and custom parameter names.

### DIFF
--- a/_playwright-tests/test_system.test.ts
+++ b/_playwright-tests/test_system.test.ts
@@ -248,8 +248,8 @@ test('User should be able to delete multiple systems from Systems page', async (
 
       // Expected URL sort parameter: ascending = display_name, descending = -display_name
       const expectedSortParam = {
-        ascending: /sort=display_name|order_how=asc/,
-        descending: /sort=-display_name|order_how=desc/,
+        ascending: /sort=display_name|sort_dir=asc/,
+        descending: /sort=-display_name|sort_dir=desc/,
       };
 
       // Keep clicking until we reach the desired sort direction (max 3 clicks)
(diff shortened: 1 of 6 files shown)


johnsonm325 (approved):
lgtm

PR #3073, opened 2026-04-20
fix(RHINENG-25775): SystemsView perPage url param is not compatible with legacy table

raswonders:
## Jira
Fixes RHINENG-25775

## What
Changes URL param from perPage to per_page in SystemsView so it's consistent with legacy table.

## Why
Needed for backward support of url links.

## Testing
Verify SystemsView renders per_page.

## Summary by Sourcery

Bug Fixes:
- Use the per_page query parameter instead of perPage in SystemsView pagination to ensure legacy URLs continue to work.


johnsonm325 (approved):
lgtm

PR #3074, opened 2026-04-20
fix(RHINENG-14708): fix mount ApplicationTab for connection-gated tabs

Siasurai:
## Jira
https://redhat.atlassian.net/browse/RHINENG-14708

## What
The tab wrapper always loads when you open a tab. It gets a flag when the host is not connected to Insights.
`ApplicationTab` shows things in this order: permissions → not connected → real tab content.
The active tab also matches the `?appName=` value in the URL when the query changes.

## Why
For application tabs (for example Vulnerability), we check permission first.
If the user cannot read that tab, they see `Access denied`, not `Not connected`, even when the system is disconnected.

## How
Before: If the host was not connected to Insights, we showed Not connected and we did not load the tab component.
After: We always load the tab component (`ApplicationTab`) when you open that tab.

## Testing

1. Log in as a user without Vulnerability read permission.
2. Open Host details → Vulnerability tab on a disconnected host.
3. You should see `Access denied`, not `Not connected`.

before:
https://github.com/user-attachments/assets/e366e36f-d384-4acd-bbf0-366ce279f4cf

after fix
https://github.com/user-attachments/assets/f1fcebdc-5f24-43f2-b768-6b771bd29a60

## Summary by Sourcery

Ensure application detail tabs always mount their content and gate visibility by permissions before host connectivity status.

Bug Fixes:
- Show access denied for disconnected hosts when the user lacks required permissions instead of only showing a not connected state.

Enhancements:
- Pass host connectivity status into application tabs so they can render a not connected state after permission checks.
- Update application tab selection effect to respond to URL query changes for the active app.

### DIFF
--- a/src/ApplicationTab.js
+++ b/src/ApplicationTab.js
@@ -2,6 +2,7 @@ import { useConditionalRBAC } from './Utilities/hooks/useConditionalRBAC';
 import PropTypes from 'prop-types';
 import React from 'react';
 import AccessDenied from './Utilities/AccessDenied';
+import { NotConnected } from '@redhat-cloud-services/frontend-components/NotConnected';
 import {
   AdvisorTab,
   ComplianceTab,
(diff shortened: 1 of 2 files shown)


johnsonm325 (approved):
lgtm. great job!

PR #3075, opened 2026-04-20
fix(RHINENG-25907): Add useSearchParamsWithFragment hook to unify query and hash params

computercamplove:
## Jira
[RHINENG-25907](https://redhat.atlassian.net/browse/RHINENG-25907)

## What
 Add `useSearchParamsWithFragment` hook to unify query and hash params

## Why
Global ta filter is not working on `SystemsView` - this change is part of decommissioning of GlobalTag filter

## How
This hook merges standard query strings and hash-based parameters (used in Global tag filter) into a single URLSearchParams object.

## Testing
1. in stable navigate to https://console.redhat.com/ansible/inventory
2. change to preview version - any global tag filter will map to new Workload filter, tags filter stays without normalization 
3. Global tag filter should not  cause any errors when switch between stable and preview UI versions

[RHINENG-25907]: https://redhat.atlassian.net/browse/RHINENG-25907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/SystemsView/SystemsView.tsx
+++ b/src/components/SystemsView/SystemsView.tsx
@@ -37,6 +37,7 @@ import {
   useDataViewFiltersContext,
 } from './DataViewFiltersContext';
 import { useDebouncedValue } from '../../Utilities/hooks/useDebouncedValue';
+import { useSearchParamsWithFragment } from './hooks/useSearchParamsWithFragment';
 import { useResetPage } from './hooks/useResetPage';
 import { INITIAL_PAGE, NO_HEADER } from '../InventoryViews/constants';
 import { PER_PAGE } from '../../constants';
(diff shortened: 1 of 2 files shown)


raswonders (changes requested):
We can implement this purely with URL parsing in SystemsView.tsx rather than coupling it with filters and its context. This solution would be streamlined and part of a normal render, no useEffects needed.

In this solution we would fill state [searchParams, setSearchParams] with our new custom hook useSearchParamsWithFragment. 
Inside of this hook, we could parse searchParams as we originally did, together with parsing key=value pairs from Fragment portion of URL. (The part used by global filter) In the end we would merge those together, deduplicate them , and return it.

Once done our filtering logic won't need to change because all params would be consolidated into searchParams which we already use. 
  
We should also apply this parsing unconditionally.That means not only when ansible path is visited, as all paths can benefit from this and shouldn't be harmed

computercamplove:
cc @raswonders i applied your suggestion:

1. just a note about known issue in Global tag - when we select in global tag filter multiple workloads - the URL only displays the last applied workload
2. otherwise any combination with tags+workload in global tag should work when we switch between preview mode and stable mode

raswonders (approved):
looks great, thank you!

PR #3086, opened 2026-04-22
feat(RHINENG-25960): Add kessel permissions to workspace details

johnsonm325:
## Jira
[RHINENG-25960](https://redhat.atlassian.net/browse/RHINENG-25960)

## What
The Inventory table on the Workspace details page needs kessel permissions implemented. 

## How
- Gate “Remove from workspace” on row action
- Gate “Remove from workspace” in toolbar action
- Gate “Add systems” in toolbar action
- Show no access page when a user doesn’t have view permissions on the workspace

[RHINENG-25960]: https://redhat.atlassian.net/browse/RHINENG-25960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Integrate Kessel-based workspace permission checks into the workspace details page and related actions while keeping RBAC as a fallback.

New Features:
- Add a Kessel workspace permissions hook to derive view/edit capabilities for a single workspace on the details page.
- Gate workspace details header actions, systems table actions, and empty-state actions on combined RBAC and Kessel edit permissions, with appropriate loading and no-access messaging.
- Show a dedicated access-denied or loading state on the workspace details page when Kessel view/edit checks fail or are still in progress.

Enhancements:
- Refine workspace group header behavior to disable and tooltip header actions when edit access is not granted or permissions are loading.
- Propagate workspace permission context through the workspace systems wrapper so child components consistently respect Kessel edit gating.
- Introduce shared helpers for computing workspace edit-action access and Playwright helpers for waiting on an editable workspace header.
- Add a mock Kessel checkselfbulk route and feature-flag aware tests to stabilize Kessel-dependent unit and end-to-end tests.

Tests:
- Extend unit tests for the group detail header to cover Kessel feature-flag scenarios, permission-denied states, and ungrouped workspace behavior.
- Add unit tests for the Kessel workspace permissions hook and update Playwright tests to use new workspace helpers and mocked Kessel access responses.

### DIFF
--- a/_playwright-tests/test_workspace.test.ts
+++ b/_playwright-tests/test_workspace.test.ts
@@ -6,11 +6,14 @@ import {
 import { searchByName, waitForTableKebabReady } from './helpers/filterHelpers';
 import {
   installKesselCheckSelfBulkAllowAll,
+  installKesselCheckSelfBulkDenyEdit,
+  installKesselCheckSelfBulkDenyView,
   uninstallKesselCheckSelfBulkMock,
 } from './helpers/kesselAccessRouteMock';
 import {
   generateUniqueWorkspaceName,
   createNewWorkspace,
+  getWorkspaceIdFromWorkspacesListLink,
   isWorkspaceResponse,
   waitForWorkspaceDetailPageReady,
   workspaceHeaderActionsToggle,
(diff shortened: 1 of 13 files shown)


_playwright-tests/test_workspace.test.ts
raswonders, line 1: I'm not sure about value of these tests. With kessel call being stubbed, it looks like we kind of addressing unit/integration concerns rather then e2e. Could we do without these? And if needed test rather on component level?

src/Utilities/hooks/useWorkspaceDetailEditActionsAccess.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Merges RBAC workspace write with Kessel `edit` for workspace-details actions
+ * (Add systems, Remove from workspace) when {@link Params.workspaceKesselGateActive} is true.
+ */
+export const useWorkspaceDetailEditActionsAccess = ({
+  workspaceKesselGateActive,
+  workspaceKesselCanEdit,
+  workspaceKesselPermissionsLoading,
+  rbacCanModify,
+}: Params) => {
+  const useKesselEditGate =
```
raswonders: this boolean clashes with use* naming convention for hooks, I'd use more boolean appropriate name like isKesselEditEnabled hasKesselEdit or similar....

raswonders (approved):
Hey thanks for this change, i tried it and UI seems to work as prescribed. Only thing I'd rethink whether we really want those PW tests. Also left small refactor nit-pick for use* names :)

PR #3087, opened 2026-04-22
fix(RHINENG-25664): Move workspaces in Workspace details

johnsonm325:
## Jira
[RHINENG-25664](https://redhat.atlassian.net/browse/RHINENG-25664)

## What
A user should be able to move a system from another workspace into the current workspace in workspace details via the Add system modal.

Currently, a user with view permissions can select and attempt to add, but the api will block them. This is a known interaction and will be fixed in RHINENG-24481.

## How
To test:
- go to workspace details
- click Add System button in toolbar
- test in modal that you can select a system in another workspace and the alert doesn't appear, and the Add system button in the modal isn't disabled.

[RHINENG-25664]: https://redhat.atlassian.net/browse/RHINENG-25664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Allow adding systems that already belong to another workspace from the workspace details "Add systems" modal.

New Features:
- Permit adding systems that already belong to another workspace via the Add systems modal in workspace details.

Enhancements:
- Simplify Add systems modal footer by removing the inline warning message and its dependency on system group membership.

Tests:
- Update Cypress test expectation to verify that systems already in another workspace no longer trigger a warning or disable the Add systems button.

### DIFF
--- a/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
+++ b/src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
@@ -77,8 +77,8 @@ describe('AddSystemsToGroupModal', () => {
 
     cy.wait('@getHosts');
     cy.get('h1').contains('Add systems');
-    cy.get('button').contains('Add systems');
-    cy.get('button').contains('Cancel');
+    cy.contains('button', 'Add systems');
+    cy.contains('button', 'Cancel');
   });
 
   it('renders the inventory table', () => {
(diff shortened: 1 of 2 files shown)


computercamplove:
@johnsonm325  Verified - works as expected.

Have questions 
1. it doesn't check for feature flag Kessel flag - is this change expected to be available for user before Move button is introduced in Kessel V2?
2. Is modal name and modal's button name should still be "Add" instead of Move?

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
```diff
@@ -108,7 +108,7 @@ describe('AddSystemsToGroupModal', () => {
       .should('deep.equal', ['consectetur']);
   });
 
-  it.skip('cannot add systems that are already in group', () => {
+  it.skip('can add systems that already belong to another workspace', () => {
```
computercamplove: why we skip this test?

computercamplove (approved):

src/components/InventoryGroups/Modals/AddSystemsToGroupModal.cy.js
johnsonm325: that's a great question. haha. looks like Claude updated a test but left it skipped.

PR #3100, opened 2026-04-27
fix(RHINENG-26093): SystemsView actions have incomplete access control

raswonders:
## Jira
RHINENG-26093

## What
When kessel UI is disabled, actions in SystemsView have incomplete access control. They are enabled at times when they should be disabled. For instance, `Add to workspace` is allowed on rows which clearly has workspace already, which is wrong.  

## Why
SystemsView must be able to work also without Kessel enabled. Hence this UI needs to have proper access control.

## Testing
1. Turn on SystemsView
2. observe BulkActions and RowActions correctly handle situations when system(s) have or doesn't have workspace assigned.

## Summary by Sourcery

Tighten workspace-related access control in SystemsView actions to respect existing workspace assignments.

Bug Fixes:
- Disable bulk and row 'Add to workspace' actions when systems already belong to a workspace.
- Disable bulk and row 'Remove from workspace' actions when systems do not belong to any workspace.

### DIFF
--- a/src/components/SystemsView/SystemsViewBulkActions.tsx
+++ b/src/components/SystemsView/SystemsViewBulkActions.tsx
@@ -14,6 +14,7 @@ import {
   GENERAL_GROUPS_WRITE_PERMISSION,
   GENERAL_HOSTS_WRITE_PERMISSIONS,
 } from '../../constants';
+import { hasWorkspace } from './utils/systemHelpers';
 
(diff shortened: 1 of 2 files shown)


raswonders:
Back to draft as it's not clear how we deal with actions in RBACv1

johnsonm325 (approved):
lgtm

PR #3101, opened 2026-04-28
chore(RHINENG-21928): Use group_id instead of group_name

johnsonm325:
## Jira
[RHINENG-21928](https://redhat.atlassian.net/browse/RHINENG-21928)

## What
When you click on a group in the Inventory -> Workspaces tab in the UI, it will show you the list of hosts belonging to that group. UI uses "GET /hosts?group_name=<...>" endpoint to get this list of hosts. In the past, the group names were unique, so this worked fine. However, with Kessel multiple groups can have the same name if they have a different parent. In this situation, "GET /hosts?group_name=<...>" can return hosts from a different group than intended and the list of hosts in the UI will be wrong.

UI should start using the new "group_id" filter on the "GET /hosts" endpoint to get the correct list of hosts for a specific group.

group_name is also used when filtering the table on a workspace. This PR updates that as well.

## Testing
These should be tested on both Inventory Table and SystemsView tables.

- Create multiple groups with the same name (through Kessel - they need to have a different parent)
- Assign some hosts to both of these groups
- Go to Inventory -> Workspaces, open one of these two groups, and check that only the hosts that belong to the selected group are shown

To test workspace filters:
- Go to Inventory table
- Apply a workspace filter (note that the group_id is include in the url and the network request
- Apply Ungrouped Hosts filter as well
- Refresh the page and make sure they are applied

[RHINENG-21928]: https://redhat.atlassian.net/browse/RHINENG-21928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Switch host and systems workspace filtering from group names to stable group IDs and introduce shared workspace utilities for IDs, labels, and ungrouped handling.

Bug Fixes:
- Ensure hosts and systems views return the correct hosts when multiple workspaces share the same name by querying and filtering via group_id instead of group_name.
- Fix ungrouped hosts filtering so it consistently uses the Kessel ungrouped workspace ID and supports empty-string fallbacks without breaking chips or bulk selection.

Enhancements:
- Add shared hooks and queries for workspace/group lookups, including infinite queries, ungrouped workspace ID resolution, and chip display-name resolution with loading states.
- Update workspace filter UIs to show counts, handle ungrouped workspaces explicitly, and use group IDs throughout while keeping chips and deselection behavior consistent.
- Refactor API client wrappers and filter serialization to pass structured parameter objects to the updated host-inventory client and align URL/query parameter names with backend expectations.
- Adjust inventory and systems filter chip rendering to support non-string chip labels by keying chips off a separate value field.

Build:
- Bump @redhat-cloud-services/host-inventory-client to version 5.0.0 to support group_id-based filtering.

Tests:
- Update and extend unit, Cypress, and Playwright tests to cover group_id-based workspace filtering, ungrouped workspace behavior, and the new workspace query hooks.


raswonders (changes requested):
So I've tried SystemsView, InventoryTable and Workspace groups none of them seem to filter selected workspace properly now. It shows unfiltered lists of systems for each of them

johnsonm325:
Try again. I remember this happening to me when I was creating the PR. I had to `rm -rf node_modules` and `npm i`. After that, everything worked fine.

Here's what I'm seeing locally right now on Inventory Table and Systems View:

image

src/routes/Systems/components/SystemsTable/components/filters/WorkspaceFilter.js
raswonders, line 1: this is SystemsTable POC file that will be removed

src/routes/Systems/components/SystemsTable/components/filters/WorkspaceFilter.test.js
raswonders, line 1: this is SystemsTable POC file that will be removed

src/routes/Systems/components/SystemsTable/filterTypes.js
raswonders, line 1: this is SystemsTable POC file that will be removed

src/routes/Systems/components/SystemsTable/filters.js
raswonders, line 1: this is SystemsTable POC file that will be removed

src/components/SystemsView/filters/WorkspaceFilter.tsx
```diff
@@ -22,23 +23,27 @@ import {
...
+import { DataViewFiltersContext } from '../DataViewFiltersContext';
 
 interface WorkspaceFilterProps {
   placeholder?: string;
   value?: string[];
   onChange?: (event?: React.MouseEvent, values?: string[]) => void;
 }
 
-export const UNGROUPED_ID = 'Ungrouped hosts';
+export { UNGROUPED_HOSTS_LABEL };
```
raswonders: we should probably export this just from constants

src/components/SystemsView/filters/SystemsViewFilters.tsx
```diff
@@ -38,6 +46,8 @@ export const isToolbarLabel = (
 
 export const SystemsViewFilters = () => {
   const { filters, onSetFilters } = useDataViewFiltersContext();
+  const { names, isFetching, ids, pendingLabelFetchIds } =
```
raswonders, line 49: The hook here is refetching names based on ids... I don't think we need that. Our WorkspaceFilter already has structure with both name and id in it.

src/components/SystemsView/filters/SystemsViewFilters.tsx
```diff
@@ -128,25 +138,34 @@ export const SystemsViewFilters = () => {
...
           ouiaId="SystemsViewWorkspaceFilter"
           filterComponent={WorkspaceFilter}
           createLabel={(value, title) =>
-            value?.map((item: string) => {
-              return {
-                key: title,
-                node: item,
-              };
-            }) ?? []
+            value?.map((item: string) => ({
```
raswonders, line 147: Why do we need special Chip component here? This together with refetching group names via useWorkspaceDisplayNames seems like regression.

raswonders (changes requested):
The UI works as it suppose to. Some code in this PR updates SystemsTAble PoC I think it should be removed. My main concern is regarding chip generation for WorkspaceFilter. We seems to regress and now are refetching names which are already present in WorkspaceFilter state. I'd suggest to completely revert this part to how it was.

src/components/SystemsView/filters/SystemsViewFilters.tsx
johnsonm325: So, the reason we have that extra name fetch is because we don't fetch the names of the workspaces until we open the workspace filter. So, if a user loads the page with a workspace filter in the url, the chip will show the workspace id: 
image

src/components/SystemsView/filters/SystemsViewFilters.tsx
johnsonm325: The custom chip component was there to handle the loading state while the chip name was being fetched, among a couple of other interactions.

raswonders (approved):
lgtm, thanks for change and due diligence!

PR #3102, opened 2026-04-29
feat(AsyncInventory): conditionaly render RBACProvider

raswonders:
Renders RBACProvider conditionally and so prevents potential unhandled exceptions inside of it when Kessel is enabled.

## Summary by Sourcery

Enhancements:
- Integrate Kessel migration feature flag into AsyncInventory to control whether RBACProvider is used.


raswonders:
closing, after further discussion RBACv1 calls doesn't seem to cause any crashes in apps.

PR #3117, opened 2026-05-04
fix(RHINENG-26210): Add tooltip to inline name edit sys details

johnsonm325:
## Jira
[RHINENG-26210](https://redhat.atlassian.net/browse/RHINENG-26210)

## What
In RBAC v2, when a user doesn’t have permissions to edit a system, the inline edit icon on the System details page does not show a tooltip telling the user why the button is disabled.

image

[RHINENG-26210]: https://redhat.atlassian.net/browse/RHINENG-26210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add configurable permission-denied tooltips for the system name inline edit button and align behavior with Kessel migration state.

New Features:
- Allow EditButton to accept a custom no-access tooltip message via a new prop.
- Introduce a Kessel-specific tooltip message used when migration is enabled and write permissions are false.

Enhancements:
- Use aria-disabled instead of fully disabling the edit button to improve accessibility semantics for the disabled state.

Tests:
- Extend EditButton no-access tests to cover Kessel migration behavior and the new tooltip messaging variants.

### DIFF
--- a/src/components/GeneralInfo/EditButton/EditButton.js
+++ b/src/components/GeneralInfo/EditButton/EditButton.js
@@ -7,7 +7,9 @@ import { PencilAltIcon } from '@patternfly/react-icons';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 
 import { useConditionalRBAC } from '../../../Utilities/hooks/useConditionalRBAC';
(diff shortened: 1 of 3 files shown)


raswonders (approved):
lgtm!

PR #3118, opened 2026-05-04
fix(RHINENG-26211): Add tooltip to action dropdown workspace details

johnsonm325:
## Jira
[RHINENG-26211](https://redhat.atlassian.net/browse/RHINENG-26211)

## What
If a user doesn’t have RBAC v2 permissions to edit a workspace, the Action dropdown in the top right of the page should show a tooltip when it’s disabled.

image

[RHINENG-26211]: https://redhat.atlassian.net/browse/RHINENG-26211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/InventoryGroupDetail/GroupDetailHeader.js
+++ b/src/components/InventoryGroupDetail/GroupDetailHeader.js
@@ -6,6 +6,7 @@ import {
   FlexItem,
   MenuToggle,
   Skeleton,
+  Tooltip,
 } from '@patternfly/react-core';
 import {
   PageHeader,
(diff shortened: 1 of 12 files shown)


src/components/InventoryGroupDetail/GroupDetailHeader.js
```diff
@@ -123,21 +127,54 @@ const GroupDetailHeader = ({
...
+              if (!canModifyWorkspaceForActions) {
+                return (
+                  <Tooltip content={noAccessEditTooltip}>
+                    <span
+                      data-testid="group-detail-header-actions-tooltip-trigger"
+                      className="pf-v6-u-display-inline-block ins-c-group-detail-header__actions-tooltip-anchor"
+                    >
+                      {menuToggle}
+                    </span>
+                  </Tooltip>
```
raswonders: This seems like one of those things which we could reuse in multiple places, I believe I just merged one PR with similar task for edit button.

Wouldn't it be better to create standalone wrapper for this?
If permissions would be missng it could render tooltip around children and pass isDisabled={true}

raswonders: We would also not impact internals of this component which became quite more complex

raswonders (commented):
I'm not approving yet, as this seems like repeating pattern which might be worthy of time investment put into creating of standalone component. Let me know what you think.

src/components/NoAccessTooltipWrap.tsx
raswonders, line 1: Thanks for introducing the component, 

Do you think we can have it in TS though? 
WHYs
- We don't have to convert it later.
- It's gonna be more reusable
- It's a leaf so should be easily convertible

raswonders (approved):
Thanks for introducing the wrapper component. Looks much better now. If you made it TS it would be perfect. /lgtm

PR #3123, opened 2026-05-06
feat(SystemsView): revamp of columns

raswonders:
## Jira
first part of RHINENG-26074

## What
This PR does few things:

#### Simplifies useColumns by consolidating column structures into one
- This was achieved by patching Patternfly ColumnManagementModal which can now work with our RenderableColumn type by default

#### Major refactor of Inventory Column Cells
- we use our own copies of column cells, this help us remove SystemsTable later on
- every Cell componets was cleaned & transformed to Typescript
- every Cell component was covered by unit tests
- fixes Bug in Name column where we didn't suggest upgrade from CentOS

#### Creates wiring for adding App based columns
- established pattern and folder structure allow us adding column simultaneously

## Why
This is to prepare SystemsView column for major extension. It should be posibble introduce columns in paralel without unnecessary merge conflicts. It was also needed because since POC we didn't really address tech debt on table cells.

## Testing
1. localStorage.setItem("ui.inventory-views","true") in console (in order to see column management modal)
2. Visit SystemsView in Inventory, try to paginate and observe whether old table cells behave the same. 
3. Open column management modal and make sure it works

## Summary by Sourcery

Revamp SystemsView table column handling to use a unified, renderable column definition model and prepare for per-app column extensions.

New Features:
- Introduce a local, customizable ColumnManagementModal compatible with SystemsView renderable columns.
- Define inventory-specific column definitions for SystemsView as a composable list to enable future app-based column sets.

Enhancements:
- Refactor SystemsView to consume a single RenderableColumn model for both rendering and management instead of separate base and derived column structures.
- Extract inventory table cell renderers (display name, workspace, OS, last seen, tags) into dedicated TypeScript components under the SystemsView columns namespace.
- Adjust SystemsView hooks and utilities to use the new column definition registry and updated type contracts.

Tests:
- Add unit tests for DisplayName, Workspace, OperatingSystem, LastSeen, and Tags cell components in SystemsView.

### DIFF
--- a/src/components/SystemsView/columns/inventory/cells/DisplayName.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/DisplayName.tsx
@@ -1,0 +1,78 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { ConversionPopover } from '../../../../InventoryTable/ConversionPopover/ConversionPopover';
+import { Flex, FlexItem, Icon, Popover } from '@patternfly/react-core';
+import { BundleIcon } from '@patternfly/react-icons';
+import FontAwesomeImageIcon from '../../../../FontAwesomeImageIcon';
+import type { System } from '../../../hooks/useSystemsQuery';
+
+const isImageBasedSystem = (system: System) =>
+  system?.system_profile?.bootc_status?.booted?.image_digest ||
+  system?.system_profile?.host_type === 'edge';
+
+const isCentosLinuxSystem = (system: System) =>
+  system?.system_profile?.operating_system?.name === 'CentOS Linux';
+
+interface DisplayNameProps {
+  system: System;
+}
+
+const DisplayName = ({ system }: DisplayNameProps) => {
+  return (
+    <div className="ins-composed-col sentry-mask data-hj-suppress">
+      <div key="data">
+        <Flex gap={{ default: 'gapSm' }}>
+          <FlexItem>
+            {isImageBasedSystem(system) ? (
+              <Popover
+                triggerAction="hover"
+                headerContent="Image-based system"
+                bodyContent={
+                  <div>
+                    Image mode for Red Hat Enterprise Linux is a
+                    container-native approach that uses the same bits but
+                    delivers them as a container image. Updates are immutable
+                    and the experience is very close to running a containerized
+                    application.
+                  </div>
+                }
+              >
+                <Icon aria-label="Image mode icon">
+                  <FontAwesomeImageIcon
+                    fill="var(--pf-t--global--icon--color--subtle)"
+                    margin="0px"
+                  />
+                </Icon>
+              </Popover>
+            ) : (
+              <Popover
+                triggerAction="hover"
+                headerContent="Package-based system"
+                bodyContent={
+                  <div>
+                    Package mode is a familiar RHEL experience across any
+                    footprint where the OS is assembled and updated from rpm
+                    packages. This is traditionally how RHEL is deployed and
+                    will remain the preferred method for many.
+                  </div>
+                }
+              >
+                <Icon aria-label="Package mode icon">
+                  <BundleIcon color="var(--pf-t--global--icon--color--subtle)" />
+                </Icon>
+              </Popover>
+            )}
+          </FlexItem>
+          <FlexItem>
+            <Link to={system.id || ''}>{system.display_name}</Link>
+          </FlexItem>
+          <FlexItem>
+            {isCentosLinuxSystem(system) && <ConversionPopover />}
+          </FlexItem>
+        </Flex>
+      </div>
+    </div>
+  );
+};
+
+export default DisplayName;
--- a/src/components/SystemsView/columns/inventory/cells/LastSeen.test.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/LastSeen.test.tsx
@@ -1,0 +1,75 @@
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import LastSeen from './LastSeen';
+import { TestWrapper } from '../../../../../Utilities/TestingUtilities';
+import type { System } from '../../../hooks/useSystemsQuery';
+
+const LONG_AGO_LAST_SEEN = '2020-01-01T00:00:00.000Z';
+
+const SYSTEM_ID = 'test-system-id';
+
+const systemWithLongAgoLastCheckIn = {
+  id: SYSTEM_ID,
+  last_check_in: LONG_AGO_LAST_SEEN,
+} as unknown as System;
+
+const systemWithUndefinedLastCheckIn = {
+  id: SYSTEM_ID,
+  last_check_in: undefined,
+} as unknown as System;
+
+const systemWithNullLastCheckIn = {
+  id: SYSTEM_ID,
+  last_check_in: null,
+} as unknown as System;
+
+const systemWithEmptyPerReporterStaleness = {
+  id: SYSTEM_ID,
+  last_check_in: LONG_AGO_LAST_SEEN,
+  per_reporter_staleness: {},
+} as unknown as System;
+
+describe('LastSeen cell', () => {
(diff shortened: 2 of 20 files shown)


src/components/SystemsView/columns/inventory/cells/LastSeen.tsx
```diff
@@ -0,0 +1,59 @@
...
+const DEFAULT_CULLING_DATE: CullingDate = new Date(0);
+
+interface LastSeenProps {
+  system: System;
+}
+
+const LastSeen = ({ system }: LastSeenProps) => {
+  const updated = system.last_check_in;
+  const culled = system.culled_timestamp;
+  const staleWarn = system.stale_warning_timestamp;
```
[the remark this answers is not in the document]
raswonders: That's CullingInformation fallback, we just reuse it

src/components/SystemsView/columns/inventory/cells/LastSeen.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    render(
+      <TestWrapper>
+        <LastSeen system={systemWithUndefinedLastCheckIn} />
+      </TestWrapper>,
+    );
+
+    expect(screen.getAllByText('Invalid date').length).toBeGreaterThan(0);
+  });
+
+  it('should show Invalid date when last_check_in is null', () => {
```
computercamplove, line 54: i wonder if these Invalid date cases even possible - does backend maybe has fallback in case it is null/invalid?

computercamplove: ignore , not critical

src/components/SystemsView/columns/inventory/cells/OperatingSystem.test.tsx
```diff
@@ -0,0 +1,53 @@
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import OperatingSystem from './OperatingSystem';
+import type { System } from '../../../hooks/useSystemsQuery';
+
+const SYSTEM_ID = 'test-system-id';
+
+const systemWithRhelOs = {
```
computercamplove, line 9: case for Centos Linux can be added to unit test, but again - not critical

src/components/SystemsView/columns/inventory/cells/Workspace.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { System } from '../../../hooks/useSystemsQuery';
+
+interface WorkspaceProps {
+  system: System;
+}
+
+const Workspace = ({ system }: WorkspaceProps) => {
+  const [firstGroup] = system.groups ?? [];
+
+  if (firstGroup === undefined) {
```
computercamplove, line 12: i wonder if we even need this fallback? now production should not have hosts without workspace. When new host is uploaded automatically added to "Ungouped Hosts" workspace

src/components/SystemsView/columns/inventory/cells/Workspace.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+} as unknown as System;
+
+describe('Workspace cell', () => {
+  it('should show the first workspace name when present', () => {
+    render(<Workspace system={systemWithNamedWorkspace} />);
+
+    expect(screen.getByText('Production')).toBeInTheDocument();
+  });
+
+  it('should show ungrouped label when system has no groups', () => {
```
computercamplove: since system always in the workspace i don't think we need to test "Ungrouped hosts" cases in the columns

computercamplove (approved):
Great work! LGTM!

i left comments, but they are not blockers or critical changes.

src/components/SystemsView/columns/inventory/cells/Workspace.tsx
computercamplove: if this case happened for some reason that hosts doesn't have "groups" - i think it is not correct to display "Ungrouped hosts" since it is name of the workspace - before it was just a label to identify that host has no group/workspace, but now ungrouped:true means host in the workspace with the name "Ungrouped Hosts"

this is invalid case when groups: [] doesn't exists - should we display different message for it? like not available. but it s very edge case

raswonders: I agree we should distinguish between not having workspace and having default, so when Backend shifts we could detect this and display in UI... Not Available is fine with me

src/components/SystemsView/columns/inventory/cells/DisplayName.tsx
```diff
@@ -0,0 +1,78 @@
...
+            ) : (
+              <Popover
+                triggerAction="hover"
+                headerContent="Package-based system"
+                bodyContent={
+                  <div>
+                    Package mode is a familiar RHEL experience across any
+                    footprint where the OS is assembled and updated from rpm
+                    packages. This is traditionally how RHEL is deployed and
+                    will remain the preferred method for many.
```
johnsonm325, line 56: Do you think these are worth defining as constants?

src/components/SystemsView/columns/inventory/cells/DisplayName.tsx
```diff
@@ -0,0 +1,78 @@
...
+                  </div>
+                }
+              >
+                <Icon aria-label="Image mode icon">
+                  <FontAwesomeImageIcon
+                    fill="var(--pf-t--global--icon--color--subtle)"
+                    margin="0px"
+                  />
+                </Icon>
+              </Popover>
```
johnsonm325, line 46: Up to you, but we could define this as a component with the same file since it's repeated below.

src/components/SystemsView/columns/inventory/cells/LastSeen.tsx
```diff
@@ -0,0 +1,59 @@
...
+    updated === undefined || updated === null ? '' : updated;
+
+  return (
+    <CullingInformation
+      className=""
+      content=""
+      culled={culled ?? DEFAULT_CULLING_DATE}
+      staleWarning={staleWarn ?? DEFAULT_CULLING_DATE}
+      stale={stale ?? DEFAULT_CULLING_DATE}
+      currDate={DEFAULT_CULLING_DATE}
```
johnsonm325, line 34: This seems sus. How is `currDate` used? It's either a param we need to throw out, or we need to provide it the current date.

johnsonm325: I see the sourcery bot comment.

src/components/SystemsView/columns/inventory/cells/Workspace.tsx
raswonders: As discussed we will show N/A in case endpoint broke.

src/components/SystemsView/columns/inventory/cells/Workspace.test.tsx
raswonders: I removed these

src/components/SystemsView/columns/inventory/cells/OperatingSystem.test.tsx
raswonders: added

src/components/SystemsView/columns/inventory/cells/DisplayName.tsx
raswonders: If we don't count SystemsTable which will be deleted in following weeks, we don't have another instance where we render the popover with this text. So I guess its fine to keep it inside this component as its unique to it. 

I'm putting to constants file, things which are currently reused or have strong potential to be reused in future. The component by itself is small and quite readable so I don't see big benefits in cleaning the markup further.

raswonders: Ah also the fact we render Popover twice in same component is fine I believe. Popover is meant to be reusable. If we kept spawning these icons elsewhere I'd agree that abstracting it would be great idea.

johnsonm325 (approved):

PR #3173, opened 2026-05-13
feat(RHINENG-26251): Move system in workspace details

johnsonm325:
## Jira
[RHINENG-26251](https://redhat.atlassian.net/browse/RHINENG-26251)

## What
This change is currently only being implemented when kessel is enabled for an organization.

We currently use “Remove from workspace” for a system in the Workspace details page. This supports the old workflow where a system has to be removed from a workspace to Ungrouped Hosts before it could be moved to another workspace. Now that we allow moving directly from one workspace to another, it makes more sense to replace it with “Move system”.

## Testing
- Go to workspaces
- Click on a workspace to see its details
- Click on the row kebab for a system

[RHINENG-26251]: https://redhat.atlassian.net/browse/RHINENG-26251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adarshdubey-star:
We have some conflicts

src/components/InventoryTable/moveSystemRowAction.tsx
```diff
@@ -0,0 +1,170 @@
+import React from 'react';
```
bastilian: This file should be split up into helpers, types and the component should be in here on its own.

src/components/InventoryTable/moveSystemRowAction.tsx
```diff
@@ -0,0 +1,170 @@
...
+  workspaceActionsAllowed: boolean;
+  ungrouped: boolean;
+  selectedCount: number;
+  selectedHosts: MoveSystemActionsColumnRow[];
+};
+
+/**
+ * Disabled state for workspace-detail bulk remove (legacy) or bulk move (Kessel).
+ * Encapsulates the Kessel vs non-Kessel rules used on group detail.
+ *  @param root0
```
bastilian: We should either add docs and properly describe the input paraemters, or not put this in here at all.

src/components/InventoryTable/moveSystemRowAction.tsx
```diff
@@ -0,0 +1,170 @@
...
+export type MoveSystemActionDropdownItemProps = Omit<
+  MenuItemProps,
+  'children'
+> &
+  ActionRBACProps;
+
+/**
+ * Row kebab item for moving a host to another workspace (InventoryTable / ActionDropdownItem).
+ * Only render when the parent has already gated on `useKesselMigrationFeatureFlag()`.
+ *  @param root0
```
bastilian: Same as above. Also, the props are already described in the type definition(s).

config/cypress.webpack.config.js
```diff
@@ -19,6 +19,15 @@ const { config: webpackConfig, plugins } = config({
   }),
 });
 
+webpackConfig.resolve.alias = {
+  ...webpackConfig.resolve.alias,
+  '@scalprum/react-core': resolve(__dirname, './cypressScalprumMock.tsx'),
+  '@redhat-cloud-services/frontend-components/AsyncComponent': resolve(
```
bastilian: This is not a proper way to deal with this. The `MockAsyncComponent` is also very specific for the selector, but with this it'll pop up in any place, and other cypress tests where `AsyncComponent` is used.

johnsonm325:
@bastilian I made your suggested changes. Good looking out. Let me know what you think.

config/setupTests.js
```diff
@@ -72,27 +72,38 @@ jest.mock('@scalprum/react-core', () => ({
...
 
-// Mock AsyncComponent so components that load federated modules render a test-friendly selector
+// Mock AsyncComponent so federated modules render fallback/error stubs in tests.
+// WorkspaceSelector gets an interactive stub; other modules render fallback only.
 jest.mock('@redhat-cloud-services/frontend-components/AsyncComponent', () => {
   const React = require('react');
   return {
     __esModule: true,
     default: function MockAsyncComponent(props) {
+      if (props.module === './modules/WorkspaceSelector') {
```
bastilian: Why not mock the Async component only in those tests that actually use it?

config/cypress.webpack.config.js
```diff
@@ -19,6 +19,19 @@ const { config: webpackConfig, plugins } = config({
...
+    __dirname,
+    './cypressAsyncComponentMock.tsx',
+  ),
+  [resolve(
```
bastilian: I think this is the only mock you actually need. You even state in the `WorkspaceSelectorField` docs that it is to not have all `AsyncComponents` mocked.

raswonders:
I've verified UI works as prescribed. Will green light when mocking issues are addressed.

raswonders (approved):
lgtm

src/components/GroupSystems/GroupSystems.js
```diff
@@ -132,6 +135,37 @@ const GroupSystems = ({
     groupId,
   );
 
+  const reloadDataAfterWorkspaceMove = useCallback(() => {
```
raswonders, line 148: we need to clear selection similarly like we do in ConventionalSystemsTab with `dispatch(selectEntity(-1, false))`

raswonders (changes requested):
I've run cursor's bugbot and it found a bug, TL;DR: we forgot to sync REDUX state after we move systems. so we get stale state selection which could cause unwanted visual artifacts like this 

image

src/components/GroupSystems/GroupSystems.js
```diff
@@ -132,6 +135,37 @@ const GroupSystems = ({
...
+    isKesselEnabled,
+    ungrouped,
+    canModifyWorkspaceForActions,
+    selectedCount,
+    selectedSystemsList,
+    removeLabel,
+    setCurrentSystem,
+    setMoveSystemsToWorkspaceModalOpen,
+    setRemoveHostsFromGroupModalOpen,
+  });
```
bastilian: At least these should be memoised. They are probably not causing any issues now, because other components defend against it, but normally things like this can cause unstable variables and unnecessary rerenders.

raswonders (approved):
looks good, thanks for fixing bug and improving it.

PR #3190, opened 2026-05-19
fix(RBAC): avoid calling v1 when kessel enabled

LightOfHeaven1994:
## What
Wrapping components into RBAC provider only when Kessel is disabled + also fixing the issue when Table was crashing with
```
installHook.js:1 TypeError: Cannot read properties of undefined (reading 'rows')
    at InventoryTable.js:101:28
```
Problem was that other apps are registering entities in onLoad from AsyncInventory, for example, `onLoad?.({ mergeWithEntities, ... })`  in compliance. That used to run in useEffect, which runs after the first render.
With RBACProvider children are hidden behind a spinner until RBAC finishes, so onLoad usually runs before InventoryTable mounts. So if we remove RBAC provider - inventory table mounts immediately and with not registered entities yet page crashes.

## Testing

1. Have Kessel feature flag enabled
2. Run inventory locally and open any app with systems page
3. See that table loads without any issues

## Summary by Sourcery

Gate RBAC provider usage on the Kessel migration feature flag and ensure the inventory store is initialized before rendering consumers to prevent crashes.

New Features:
- Add conditional rendering of the RBAC provider based on the Kessel migration feature flag in AsyncInventory.
- Introduce a loading spinner fallback while the inventory store initializes before rendering child content.

Bug Fixes:
- Prevent InventoryTable and other consumers from crashing due to uninitialized entities by delaying rendering until the inventory store is ready.

### DIFF
--- a/src/modules/AsyncInventory.js
+++ b/src/modules/AsyncInventory.js
@@ -1,8 +1,11 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Provider } from 'react-redux';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { Bullseye } from '@patternfly/react-core/dist/dynamic/layouts/Bullseye';
+import { Spinner } from '@patternfly/react-core/dist/dynamic/components/Spinner';
 import { RBACProvider } from '@redhat-cloud-services/frontend-components/RBACProvider';
+import { useKesselMigrationFeatureFlag } from '../Utilities/hooks/useKesselMigrationFeatureFlag';
 
 import * as storeMod from '../store/redux';
 import * as utils from '../Utilities/index';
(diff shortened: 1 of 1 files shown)


johnsonm325:
@LightOfHeaven1994 thoughts on this suggestion from Sourcery? I think it might be a good one unless you are aware of some interaction that I am not.

> The isStoreReady flag is set immediately after calling onLoad, even though onLoad can perform async registration work; if that’s the case, you may want to await or otherwise signal completion from onLoad before switching from the loading fallback to the table.

LightOfHeaven1994:
@johnsonm325 

> @LightOfHeaven1994 thoughts on this suggestion from Sourcery? I think it might be a good one unless you are aware of some interaction that I am not.
> 
> > The isStoreReady flag is set immediately after calling onLoad, even though onLoad can perform async registration work; if that’s the case, you may want to await or otherwise signal completion from onLoad before switching from the loading fallback to the table.

Sourcery is wrong this time. onLoad is synchronious. Example, 
```
export const defaultOnLoad =
  (columns) =>
  ({ INVENTORY_ACTION_TYPES, mergeWithEntities }) =>
    getRegistry().register({
      ...mergeWithEntities(entitiesReducer(INVENTORY_ACTION_TYPES, columns)),
    });
```
So we are preparing store that prepares structure like `rows: [], ...` and fetching entities happens later when `getEntities` is run.

johnsonm325 (approved):

PR #3194, opened 2026-05-19
fix(RHINENG-25942): RBAC v2 check staleness against Root workspace

johnsonm325:
## Jira
[RHINENG-25942](https://redhat.atlassian.net/browse/RHINENG-25942)

## What
The hosts and staleness permissions got tied to the Default workspace after the RBAC v2 migration. This isn’t correct, because a user should have a global access to all hosts if they want to see/edit the staleness configuration, but the Default workspace doesn’t ensure a global access. We should be checking for permissions on the root workspace, but we can’t do that until RBAC mirates those existing permissions from default workspace to the root workspace.

[RHINENG-25942]: https://redhat.atlassian.net/browse/RHINENG-25942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Update host staleness RBAC access checks and related tests to use the Root workspace instead of the Default workspace for permissions.

Bug Fixes:
- Ensure staleness view and edit permissions are validated against the Root workspace rather than the Default workspace when Kessel is enabled.

Tests:
- Adjust useHostStalenessKesselAccess tests to mock Root workspace fetching and expectations instead of Default workspace behavior.

### DIFF
--- a/_playwright-tests/helpers/kesselAccessRouteMock.ts
+++ b/_playwright-tests/helpers/kesselAccessRouteMock.ts
@@ -75,6 +75,29 @@ const WORKSPACE_RESOURCE_TYPE = 'workspace';
 const WORKSPACE_RELATION_VIEW = 'view';
 const WORKSPACE_RELATION_EDIT = 'edit';
 
+/** Root-workspace staleness page relations (see `useHostStalenessKesselAccess`). */
+const STALENESS_WORKSPACE_RELATION_VIEW = 'staleness_staleness_view';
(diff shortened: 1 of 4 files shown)


fstavela (approved):
Looks good to me, but I haven't done almost anything in frontend ever, so take it with caution :D

PR #3204, opened 2026-05-25
feat(RHINENG-26470): add Advisor columns

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-26470

## What
add Advisor columns
image

## Why
Inventory views phase 1

## How

- reads `system.app_data.advisor.recommendations` from the `InventoryViewHost` data (same for all Advisor columns)
- column is hidden by default
- advisor columns are registered via `allColumnDefinitions.ts`

## Testing

1. enable localStorage.setItem("ui.inventory-views", "true")
2. open Manage columns modal to select "Recommendation" column - apply (same for all Advisor columns)
3. check System table now displays "Recommendation" column

image
image

## Summary by Sourcery

New Features:
- Introduce an Advisor Recommendations column for systems, populated from inventory view host advisor data.


computercamplove:
i assume RBAC is not yet supported

src/components/SystemsView/columns/advisor/cells/Important.tsx
raswonders, line 1: are these cell components really necessary? They're really really shallow abstractions and not containing any markup or styling for the output they render. 

We are just parsing data from system and render it as plain string this could hopefuly be done in renderCell functions directly and we can be 10-12 files lighter as we don't need to test it either I feel.

src/components/SystemsView/columns/advisor/columnDefinitions.test.tsx
raswonders, line 1: Happy to talk this over but this possibly also doesn't deserve its own unit tests. We just checking whether static values are there.

raswonders (changes requested):
I've tested UI, columns are displayed with correct titles, only thing I'm not sure about in UI is Incidnets I couldn't find it in Advisor's Systems. 

Regarding cell components I'd create them only if column requires unique functionality or style. Parsing data i'd keep out of them otherwise they become un-sharable between columns with different data sources.

src/components/SystemsView/columns/advisor/columnDefinitions.test.tsx
computercamplove: @raswonders i am fine to combine all and parametrize tests, i was just following the jira description "For seamless integration follow code & type structure outlined by /src/components/SystemsView/columns/inventory" - i took it too far :grin:

computercamplove:
> I've tested UI, columns are displayed with correct titles, only thing I'm not sure about in UI is Incidents I couldn't find it in Advisor's Systems.

yes, they don't have - Advisor display the label against system if it  has recommendations with incident
if system has 6 incidents it means system has 6 recommendation with incident (here 6 recommendations with incident out of 18 recommendations)
image

image

i assume we don't want this label against display name column since this column is not "advisor data" - i can ask Susan

src/components/SystemsView/columns/advisor/columnDefinitions.test.tsx
raswonders: I don't think we need to test renderCell() methods.

computercamplove:
@raswonders i made a singe test file for all columns, please let me know if i need to remove something

raswonders (approved):
looks great, thanks!

PR #3207, opened 2026-05-26
ci: Roll back node version to 22.x

adarshdubey-star:
CI:
- Update all GitHub Actions workflows to run on Node.js 22.x rather than Node.js 24.x for linting, building, and Playwright test jobs.


computercamplove (approved):
thank you for the fix!

PR #3223, opened 2026-05-27
fix(RHINENG-21928): Partial revert: Restore group_name for InventoryTable

johnsonm325:
## Jira
[RHINENG-21928](https://redhat.atlassian.net/browse/RHINENG-21928)

## What
This is a partial revert of commit d3f0813b1428059675ac2ce8d1461385e23c2820 to restore backward compatibility for apps importing InventoryTable as a federated module.

Changes:
- InventoryTable components now use group_name parameter (not group_id)
- SystemsView components continue using group_id (unchanged)
- Reverted HOST_GROUP_CHIP constant back to 'group_name'
- Reverted API calls in api.js to use groupName parameter
- Reverted useGroupFilter hook to use selectedGroupNames
- Reverted related test files

## Why
The group_id change broke other apps that import InventoryTable as a federated module because they cannot handle the group_id parameter yet. This change is currently in production and causing issues.

SystemsView can safely use group_id because it uses its own separate API call path (useSystemsQuery) and is not exported as a federated module.

## How
- Go to Inventory Table
- Select Workspaces from the filter dropdown
- Choose a workspace to filter on

What you should see with the InventoryTable is the group_name filter passing the group name as it used to (url and api calls). When you use a workspace filter in the new SystemsView table you should see the group_id passed in the url and api.

[RHINENG-21928]: https://redhat.atlassian.net/browse/RHINENG-21928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Restore InventoryTable workspace filtering to use group_name instead of group_id while keeping SystemsView on group_id, re-establishing backward-compatible URL/query behavior.

Bug Fixes:
- Fix InventoryTable host group filtering to pass group_name (including empty for ungrouped) to the hosts API to match existing consumers’ expectations.

Enhancements:
- Simplify group filter state to track selected group names rather than IDs and remove indirection for resolving chip display names.
- Align workspace filter constants so InventoryTable uses HOST_GROUP_CHIP=group_name while SystemsView uses its own group_id parameter name.
- Update group filter components and tests to operate on group names, including ungrouped hosts handling and chip rendering.

Tests:
- Adjust unit and Cypress tests for SearchableGroupFilter, useGroupFilter, and InventoryTable to assert group_name-based filtering and selection behavior.

### DIFF
--- a/src/Utilities/constants.js
+++ b/src/Utilities/constants.js
@@ -12,8 +12,7 @@ export const UPDATE_METHOD_KEY = 'system_update_method';
 export const SYSTEM_TYPE_KEY = 'system_type';
 export const WORKLOAD_FILTER_KEY = 'workloads';
 export const LAST_SEEN_CHIP = 'last_seen';
-/** URL query and chip `type` for workspace filter; must match `GET /hosts?group_id=`. */
-export const HOST_GROUP_CHIP = 'group_id';
+export const HOST_GROUP_CHIP = 'group_name'; // use the same naming as for the back end parameter
 //REPORTERS
 export const REPORTER_PUPTOO = 'puptoo';
 export const REPORTER_RHSM_CONDUIT = 'rhsm-conduit';
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -321,18 +321,18 @@ export async function getEntities(
         : Array.isArray(rawHostGroupFilter)
           ? rawHostGroupFilter
           : [rawHostGroupFilter];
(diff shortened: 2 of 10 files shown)


raswonders (approved):
lgtm, code looks alright, UI seems to work also as prescribed gj

PR #3228, opened 2026-05-29
fix(RHINENG-26704): prevent stale workspaces showing during search debounce

johnsonm325:
## Jira
[RHINENG-26704](https://redhat.atlassian.net/browse/RHINENG-26704)

## What
Fixes race condition where unfiltered workspaces briefly appear in the dropdown while searching.

## Why
Bug RHINENG-26704: When users search in the workspace filter, stale unfiltered workspaces flash in the dropdown during the debounce period and while the search query is executing. This creates confusion as users see workspaces that don't match their search term.

## How
- Added `isSearchPending` state to detect the gap between user typing and debounced API call
- Added `isSearchLoading` state that combines debounce detection with query execution state
- Modified `groups` memo to return empty array during loading to prevent stale data rendering
- Added `isLoading` check in SearchableGroupFilter for defense-in-depth
- Comprehensive test coverage validates stale workspaces don't appear during search

## Testing
**Automated:**
- Added test case: "does not show stale unfiltered workspaces while remote search is debouncing"
- Pre-commit hooks passed

**Manual steps to test:**
1. Navigate to Inventory → Systems page
2. Open the Workspace filter dropdown
3. Verify initial list of workspaces loads
4. Type slowly into the search field (e.g., "hello")
5. Watch the dropdown during typing - should NOT see unfiltered workspaces flash
6. Verify only matching workspaces appear after search completes
7. Clear search and repeat with different terms to confirm consistent behavior

**Expected:** No stale/unfiltered workspaces visible during search debounce or query execution **Bug behavior (before fix):** Unfiltered workspaces briefly flash in dropdown during typing.

[RHINENG-26704]: https://redhat.atlassian.net/browse/RHINENG-26704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Prevent stale/unfiltered workspaces from appearing in the workspace filter dropdown during remote search debounce and loading states.

Bug Fixes:
- Ensure group results are cleared while a remote search term is debouncing or loading so stale workspaces do not flash in the dropdown.

Enhancements:
- Propagate a combined loading state from the group filter hook to the SearchableGroupFilter component to better control when options are rendered.

Tests:
- Add a regression test verifying that stale unfiltered workspaces are not shown while a remote search is debouncing and the remote query is loading.


raswonders (commented):
Doesn't seem to fix it. I can still reproduce the bug by searching for "d" erasing searching again "d". 

image

johnsonm325:
@raswonders That's being fixed here: https://github.com/RedHatInsights/insights-inventory-frontend/pull/3234

src/components/filters/useGroupFilter.js
```diff
@@ -141,7 +150,11 @@ const useGroupsQueryWithFilter = ({
     remoteSearchEnabled,
   ]);
 
-  // Collect data from all pages and filter groups based on the search term if remote search is disabled
+  const isSearchDebouncing =
+    remoteSearchEnabled && searchTerm !== debouncedTerm;
+  const isRemoteSearching =
+    remoteSearchEnabled && Boolean(debouncedTerm) && isFetching && isLoading;
```
bastilian, line 172: You can just use `isFetching`. `isLoading` is only `true whenever the first fetch for a query is in-flight`.

johnsonm325: Why we want `isLoading` check there:

If you remove `isLoading` check and type a search term in the dropdown, like "a", and then click the "View more" button, The results that populated in the workspace dropdown from searching "a" will temporarily disappear while the new results are loaded.

raswonders (approved):
lgtm, it solves the issue with flashing unrelated workspaces.

PR #3234, opened 2026-06-01
fix(RHINENG-26730): clear stale workspace options during InventoryTable filter search

johnsonm325:
Jira:
[RHINENG-26730](https://redhat.atlassian.net/browse/RHINENG-26730)

What:
Fixed workspace filter dropdown showing stale/duplicate "duplicate workspace name" entries while typing. Removed xor import and cleaned up ID-based selection logic in SearchableGroupFilter.

Why:
Users saw workspaces in dropdown that didn't match their current search term. The UI was rendering cached options from previous queries during the debounce window, causing duplicate entries and confusion when a workspace was named "duplicate workspace name".

How:
- Use workspace ID (itemId) consistently for unique selection
- Remove unused xor import from SearchableGroupFilter
- Clean up formatting in selection logic and chip building
- Maintain ID-based tracking to prevent name-based duplication

Testing:
Manual:
1. Open Inventory → Systems table
2. Open Workspace filter dropdown
3. Type 'd' → backspace → type 'd' again
4. Verify no duplicate "duplicate workspace name" entries appear
5. Type a term that doesn't match (e.g., "hello")
6. Verify dropdown only shows API results for current search term

[RHINENG-26730]: https://redhat.atlassian.net/browse/RHINENG-26730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Switch workspace/group filtering to be driven by stable workspace IDs instead of names to eliminate stale or duplicate options in the Inventory Systems table filter.

Bug Fixes:
- Prevent duplicate or stale workspace entries from appearing in the workspace filter dropdown when searching and debouncing results.

Enhancements:
- Use ID-based selection and chip values throughout group/workspace filtering to ensure unique, stable selection state, including support for ungrouped hosts.
- Update API host group filtering to send groupId parameters derived from workspace objects while preserving backward compatibility with existing string filters.
- Tighten prop types for group filters to reflect the new workspace object shape and remove an unused lodash xor import.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -327,7 +327,9 @@ export async function getEntities(
     const groupIds = hostGroupFilterArr.map((item) =>
       typeof item === 'object' && item !== null ? item.id : item,
     );
-    const nonEmptyGroupIds = groupIds.filter((id) => id !== '');
+    const nonEmptyGroupIds = groupIds.filter(
+      (id) => typeof id === 'string' && id !== '',
+    );
     const filterByUngroupedHosts = groupIds.includes('');
 
     /** Use groupId parameter (supports unique IDs, fixes duplicate name bug) */
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -322,25 +322,27 @@ export async function getEntities(
           ? rawHostGroupFilter
           : [rawHostGroupFilter];
 
-    // Extract IDs from workspace objects (new format: [{ id, name }])
-    // Fall back to string format for backwards compatibility
-    const groupIds = hostGroupFilterArr.map((item) =>
-      typeof item === 'object' && item !== null ? item.id : item,
+    // Selection uses workspace IDs; host list API filters by group_name.
+    const groupNames = hostGroupFilterArr.map((item) => {
+      if (typeof item === 'object' && item !== null) {
+        return item.ungrouped ? '' : item.name;
+      }
+      return item;
+    });
+    const nonEmptyGroupNames = groupNames.filter(
+      (name) => typeof name === 'string' && name !== '',
     );
-    const nonEmptyGroupIds = groupIds.filter(
-      (id) => typeof id === 'string' && id !== '',
-    );
-    const filterByUngroupedHosts = groupIds.includes('');
+    const filterByUngroupedHosts = groupNames.includes('');
 
-    /** Use groupId parameter (supports unique IDs, fixes duplicate name bug) */
-    const groupIdParam =
-      nonEmptyGroupIds.length || filterByUngroupedHosts
-        ? [...nonEmptyGroupIds, ...(filterByUngroupedHosts ? [''] : [])]
+    /** Ungrouped hosts: use group_name (empty value) per API. */
+    const groupNameParam =
+      nonEmptyGroupNames.length || filterByUngroupedHosts
+        ? [...nonEmptyGroupNames, ...(filterByUngroupedHosts ? [''] : [])]
         : [];
 
     return apiGetHostList({
       hostnameOrId: filters.hostnameOrId,
-      ...(groupIdParam.length ? { groupId: groupIdParam } : {}),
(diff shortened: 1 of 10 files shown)


src/components/filters/useGroupFilter.js
```diff
@@ -19,18 +19,27 @@ export const groupFilterReducer = (_state, { type, payload }) => ({
...
       : {
-          name: group,
-          value: group,
-        },
-  );
+          name, // Display name in chip
+          value: id, // Store ID as value
+        };
+  });
   return chips?.length > 0
```
bastilian, line 44: This can for sure be refactored and made more sensible.

src/components/filters/SearchableGroupFilter.js
```diff
@@ -163,7 +162,31 @@ const SearchableGroupFilter = ({
       return;
     }
 
-    setSelectedGroupNames(xor(selectedGroupNames, [itemId]));
+    // Handle ungrouped hosts (empty string ID)
+    if (itemId === '') {
+      const isSelected = selectedGroupNames.some((g) => g.id === '');
```
bastilian: Why are there empty strings as id's in the first place?

src/api/api.js
```diff
@@ -321,18 +321,26 @@ export async function getEntities(
...
-    const groupNameParam =
-      nonEmptyGroupNames.length || filterByUngroupedHosts
-        ? [...nonEmptyGroupNames, ...(filterByUngroupedHosts ? [''] : [])]
+    // Extract IDs from workspace objects (new format: [{ id, name }])
+    // Fall back to string format for backwards compatibility
+    const groupIds = hostGroupFilterArr.map((item) =>
+      typeof item === 'object' && item !== null ? item.id : item,
+    );
+    const nonEmptyGroupIds = groupIds.filter(
+      (id) => typeof id === 'string' && id !== '',
```
bastilian: Again with the empty id...

bastilian:
Can you explain how all these changes contribute to fixing the issue? and what the root cause was of the workspaces being stale? Because looking through the code my gut feeling tells me that about 2/3 of the code aren't needed.

johnsonm325:
> Can you explain how all these changes contribute to fixing the issue? and what the root cause was of the workspaces being stale? Because looking through the code my gut feeling tells me that about 2/3 of the code aren't needed.

I can go through section by section and explain if you want. But, the high-level view is that previously, we filtered system table data by workspace name. Now, with RBAC v2, we can have 2 workspaces with the same name. So the backend added a filter for `group_id`. But since other apps' backends weren't prepared to filter by id, we needed to continue to filter the inventory table by group name. At the same time, we're filtering by `group_id` in the new Data View table.

Before we were keeping up with selected workspace names via an array of names, but the new design includes id, name, and host count:

```
  // BEFORE: Selection tracked by NAME (string)
  selectedGroupNames = ['duplicate workspace name', 'prod-workspace']

  // Problem: If two workspaces both have name "duplicate workspace name":
  groups = [
    { id: '123', name: 'duplicate workspace name', host_count: 5 },
    { id: '456', name: 'duplicate workspace name', host_count: 3 },
    { id: '789', name: 'prod-workspace', host_count: 10 }
  ]
```

So we had the issue where we would select the duplicate workspace name from the list and it would select both duplicates because the name was the same. This PR selects the workspace by id from the list but filters by name in Inventory Table and by id in Systems View (this of course isn't perfect because it still filters by both workspace names even though 1 is selected, but it's the best trade-off for UX).

It's hard to have the perfect solution at the moment, but this is a good solution until we migrate to the new table.

raswonders (approved):
still reviewing

raswonders (approved):
lgtm, I see no more extra duplicate names in dropdown.

PR #3239, opened 2026-06-05
chore: colocate inventory views column min-widths on definitions

johnsonm325:
Move minWidth from the central columnMinWidths map onto each column definition so layout stays with column metadata. Optional minWidth keeps new columns compiling; the sticky Name column falls back to a default width.

## Summary by Sourcery

Co-locate Systems View inventory column min-width configuration with the column definitions and wire it through to header/body rendering and sticky name behavior.

Enhancements:
- Add an optional minWidth layout property to Systems View column definitions and use it to derive per-column min-width styles instead of a central mapping.
- Update sticky Name column helpers to accept the resolved min-width so sticky layout is driven by column metadata, with a default fallback for Name when omitted.

Tests:
- Add a test ensuring all inventory columns define a valid rem-based minWidth for the inventory views scroll layout.


raswonders (approved):
lgtm, works and code looks fine. Thanks for adding this

PR #3246, opened 2026-06-10
fix(InventoryViews): Update Malware columns

computercamplove:
## What
Update Malware columns - display only Total malware matches, similar to Malware page

## Why
UX review

## Summary by Sourcery

Enhancements:
- Remove the Last malware matches column from the systems inventory malware table so only total malware matches are displayed for each system.


raswonders (approved):
lgtm

PR #3247, opened 2026-06-10
fix(InventoryViews): Do not display Remediations column in Phase 1

computercamplove:
## What
Do not display Remediations column in Phase 1

## Why
Requirements needs to be revisit with the Remediations team in Q3

## Summary by Sourcery

Bug Fixes:
- Remove the Remediations column from the systems view to align the UI with current phase requirements.


raswonders (approved):
lgtm

PR #3249, opened 2026-06-10
fix(RHINENG-26715): disable manage columns while table loads

johnsonm325:
## Jira
[RHINENG-26715](https://redhat.atlassian.net/browse/RHINENG-26715)

## What
Steps to Reproduce

- visit systems page with enabled inventory views
- while table is loading open Manage columns modal
- observe the system table after load is finished

We now disable the manage columns button until the activeState === 'active'.

[RHINENG-26715]: https://redhat.atlassian.net/browse/RHINENG-26715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Gate the systems table “Manage columns” bulk action on the table’s active state to prevent interaction while data is loading or in error/empty states.

Bug Fixes:
- Disable the inventory views “Manage columns” action when the systems table is not in the active state to avoid opening column management during loading or error conditions.

Tests:
- Add unit tests covering visibility, enabled/disabled behavior, and modal opening for the “Manage columns” action across feature-flag and active state combinations.

### DIFF
--- a/src/components/SystemsView/SystemsViewBulkActions.test.tsx
+++ b/src/components/SystemsView/SystemsViewBulkActions.test.tsx
@@ -1,0 +1,194 @@
+import '@testing-library/jest-dom';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { expect, jest } from '@jest/globals';
+import { SystemsViewBulkActions } from './SystemsViewBulkActions';
+import { SystemActionModalsContext } from './SystemActionModalsContext';
+import { ColumnManagementModalProvider } from './ColumnManagementModalContext';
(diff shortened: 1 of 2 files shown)


raswonders:
@johnsonm325 were you able to reproduce the bug?

I'm curious because I cannot, even with network throttling. I can open management modal just fine while table is loading. And this is expected. 

I think users should be able to change column setting while data are being loaded.

johnsonm325:
@raswonders I also had a hard time reproducing. I thought I had reproduced it before I "fixed" it, but apparently not. @computercamplove Can you still reproduce it?

PR #3250, opened 2026-06-10
fix(RHINENG-26892): Inventory Views: refresh table after row actions

johnsonm325:
## Jira
[RHINENG-26892](https://redhat.atlassian.net/browse/RHINENG-26892)

## What
In Inventory Views table, when you edit a system's display name, workspace, or delete a system, the table doesn't automatically reload after the action. This PR addresses this bug.

Changes
Shared invalidation helper — invalidateSystemsViewQueries.ts invalidates both table query keys so Inventory Views and legacy Systems stay in sync.

Exported query key constants

`SYSTEMS_QUERY_KEY` in `useSystemsQuery.ts`
`INVENTORY_VIEWS_QUERY_KEY` in `useInventoryViewsQuery.ts` Wired through all mutation/refetch paths

`usePatchSystemsMutation.ts` — display name edit (fixes the reported bug) `useDeleteSystemsMutation.ts` — delete success/error `SystemActionModalsContext.tsx` — workspace add/remove/move reloadData Tests

`invalidateSystemsViewQueries.test.ts` — asserts both query keys are invalidated `usePatchSystemsMutation.test.ts` — asserts the mutation calls the shared helper on success

[RHINENG-26892]: https://redhat.atlassian.net/browse/RHINENG-26892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Ensure Inventory Views and Systems tables refresh consistently after row actions by centralizing query invalidation for related data.

Bug Fixes:
- Fix Inventory Views table not reloading after editing a system display name, changing workspace assignments, or deleting systems.

Enhancements:
- Introduce shared query invalidation helper that invalidates both systems and inventory views queries and reuse it across mutations and modal actions.
- Export and reuse stable query key constants for systems and inventory views queries.

Tests:
- Add tests to verify systems patch mutation and shared invalidation helper correctly invalidate the relevant queries.


raswonders (approved):
I love the change, not sure about unit tests thouhg. There is just one test case with tons of setup. I'd maybe not test it in unit tests due to heavy mocking required. Seems like integration/e2e tests would be more apropriate place. And we could do them outside of this pr.

PR #3252, opened 2026-06-12
test: Fixed Playwright test tags for proper workflow filtering

computercamplove:
## What
Changed from listing specific test files with `--grep-invert` to using positive tag matching (`--grep @inventory_views|@systems-table` and `--grep @systems-table`). This ensures each workflow runs only its intended tests:

-   `playwright-inventory-views.yml` - runs tests tagged with @inventory_views or @systems-table
-   `playwright-systems-view.yml` - runs tests tagged with @systems-table

  Also moved tags from individual tests to `test.describe()` blocks for cleaner organization.

## Summary by Sourcery

Align Playwright workflows with tag-based test selection for inventory and systems views.

Tests:
- Tag systems-related Playwright suites with @systems-table for targeted execution.
- Update Playwright CI workflows to run tests via positive tag matching instead of file lists and negative greps.

### DIFF
--- a/.github/workflows/playwright-inventory-views.yml
+++ b/.github/workflows/playwright-inventory-views.yml
@@ -94,7 +94,7 @@ jobs:
           npx wait-on http://localhost:8003/apps/inventory/
 
       - name: Run front-end Playwright tests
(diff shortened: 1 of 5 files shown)


johnsonm325 (approved):
looks great

PR #3261, opened 2026-06-16
fix(RHINENG-27054): useGroupFilter is returning object instead of array

raswonders:
RHINENG-27054

## What
Solves serialization problem for groupFilter.
Issue was introduced by latest workspace PR #3234 which introduced change to useGroupFilter api. It started returning Object instead of string[].

This PR maps over this internal selection and returns API compatible array.

## Testing

go to Inventory/Systems table and filter by workspace 
go to Malware/Systems and filter by workspace
go to Compliance/Systems and filter by workspace
go to Content/Systems and filter by workspace
go to Advisor and filter by workspace

## Summary by Sourcery

Normalize host group filter values to use group_name strings rather than workspace objects across filtering, chips, and URL parameters.

Bug Fixes:
- Fix useGroupFilter to return API-compatible string values for host group filters instead of workspace objects, including ungrouped hosts.

Enhancements:
- Introduce a helper to convert mixed workspace group selections (objects and strings) into group_name values for API and URL usage.
- Align toolbar chip values and workspace filter deletion logic to operate consistently on group_name strings rather than internal ids.

Tests:
- Extend unit tests for useGroupFilter, filter calculation, and onDeleteFilter to cover string-based group_name handling and workspace selection deletion.

### DIFF
--- a/src/components/InventoryTable/helpers.test.js
+++ b/src/components/InventoryTable/helpers.test.js
@@ -151,16 +151,13 @@ describe('onDeleteFilter', () => {
     expect(data).toMatchObject(filter);
   });
 
-  it('should delete workspace filter selected by id', () => {
-    const workspaceFilter = [
-      { id: 'ws-1', name: 'Workspace One' },
-      { id: 'ws-2', name: 'Workspace Two' },
-    ];
+  it('should delete workspace filter selected by name', () => {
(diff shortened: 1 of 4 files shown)


Fewwy (approved):
LGTM tested it in advisor and compliance and it works now

PR #3269, opened 2026-06-17
feat(RHINENG-11654): OS filter should be typeahead

johnsonm325:
## Jira
[RHINENG-11654](https://redhat.atlassian.net/browse/RHINENG-11654)

## What
'Filter by operating system' is quite cumbersome to use when trying to include multiple OS versions. List is ordered by version, causing RHEL and CentOS to be present in the list.

Using a 'Multiple typeahead with checkboxes' component would allow user to type the OS they are interested in (e.g. RHEL) and filter out other entries (e.g. CentOS). Similarily, they would be able to search for particular releases (e.g. 8.) and filter all others.

## Testing

Steps to Reproduce

- Navigate to Inventory
- Add a filter for Operating system

Actual Behavior
The list of OSes can be very long and hard to use.

Expected Behavior
User should be able to filter the list using a 'Multiple typeahead with checkboxes' component.

When a filter is applied to the OS list and only some of the children of a parent are displayed, doing a selection on the parent will select all of the unfiltered children, but none of the children that are filtered out. It will result in a checkbox on the parent.

Then, if you remove the filter and show all of the children of the same parent, the parent will have a '-' indicating a partial selection. This is the intended interaction.

image

image

[RHINENG-11654]: https://redhat.atlassian.net/browse/RHINENG-11654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Convert the operating system filter in SystemsView to a searchable typeahead experience and align workspace filter debouncing with the shared hook.

New Features:
- Enable typeahead-driven search for the operating system multi-select filter with debounced input and inline text entry.

Enhancements:
- Refine operating system filter behavior to keep the menu open during interaction, clear search when closed, and show a message when no OS matches the query.
- Reuse the shared debounced value hook for the workspace filter instead of maintaining a custom debouncing implementation.

Tests:
- Extend OperatingSystemsFilter tests to cover search filtering behavior, no-results messaging, and clearing the search input when the menu is closed.


johnsonm325:
@sourcery-ai review

computercamplove:
@johnsonm325 it is only for `SystemsView` implemented - it is not expected for `InventoryTable`?

johnsonm325:
@computercamplove I'm not sure it's worth developing this feature for Inventory table as we plan to sunset it by the end of the year.

computercamplove (approved):

raswonders (approved):
lgtm, thanks for the addition

PR #3273, opened 2026-06-18
fix(CI): add codecov action

LightOfHeaven1994:
## What
ci:verify script was taking care of uploading coverage before, but it got broken because of outdated codecov cli. To avoid such issues in the future, we better use upload codecov action which is supported

## Summary by Sourcery

CI:
- Add a Codecov upload step to the checks workflow to publish multiple coverage-final.json reports and fail the job if upload errors occur.


raswonders (approved):
lgtm

PR #3280, opened 2026-06-22
test(RHINENG-25381): add tests for utils from useSystemQuery and useInventoryViewsQuery

raswonders:
RHINENG-25381

## What
Refactors and tests parts of useSystemsQuery and useInventoryViewsQuery.

## Why
In order to increase test coverage of logic related to Queries.

## How
Considering Queries themselves are just thin wrappers they are not good target for tests themselves. But they did contain lots of serialization logic for Axios calls. This logic is refactored to seperate files and tested. 
 
## Testing
Make sure UI for Inventory / Systems works correctly. Especially filtering.
Make sure to test with both Preview ON and OFF as each uses different Query.

IMPORTANT: on SystemsView table (Preview ON), workspace filter doesn't work currently. It doesn't because of lagging implementation of group_id for /hosts-view endpoint. This feature will be enabled in subsequent PR.

## Summary by Sourcery

Extract query parameter construction for systems and inventory views into shared utilities and add targeted test coverage for these helpers.

Enhancements:
- Refactor host list and host-views query parameter assembly into reusable utility functions, including shared handling of system profile filters, system types, and Axios options.

Tests:
- Add comprehensive unit tests for host list and host-views parameter builders, system profile filter mapping, and system type mapping utilities.

### DIFF
--- a/src/components/SystemsView/filters/SystemsViewFilters.tsx
+++ b/src/components/SystemsView/filters/SystemsViewFilters.tsx
@@ -26,9 +26,7 @@ import {
   HostGroupChipNode,
   useWorkspaceDisplayNames,
 } from '../hooks/useWorkspaceDisplayNames';
-import { ApiHostViewsGetHostViewsStalenessEnum } from '@redhat-cloud-services/host-inventory-client/ApiHostViewsGetHostViews';
 
-type Test = ApiHostViewsGetHostViewsStalenessEnum[];
 export interface InventoryFilters {
   hostname_or_id: string;
   status: ApiHostGetHostListStalenessEnum[];
--- a/src/components/SystemsView/filters/SystemsViewFilters.tsx
+++ b/src/components/SystemsView/filters/SystemsViewFilters.tsx
@@ -26,7 +26,9 @@ import {
   HostGroupChipNode,
   useWorkspaceDisplayNames,
 } from '../hooks/useWorkspaceDisplayNames';
+import { ApiHostViewsGetHostViewsStalenessEnum } from '@redhat-cloud-services/host-inventory-client/ApiHostViewsGetHostViews';
 
+type Test = ApiHostViewsGetHostViewsStalenessEnum[];
 export interface InventoryFilters {
   hostname_or_id: string;
   status: ApiHostGetHostListStalenessEnum[];
--- a/src/components/SystemsView/utils/buildHostListParams.ts
+++ b/src/components/SystemsView/utils/buildHostListParams.ts
@@ -30,12 +30,12 @@ export interface BuildHostListParamsInput {
 const buildGroupIdParam = (
   groupIds: string[] | undefined,
 ): Pick<ApiHostGetHostListParams, 'groupId'> | Record<string, never> => {
-  const groupIdParam = [
-    ...(groupIds ?? []).filter((id) => id),
-    ...((groupIds ?? []).includes('') ? [''] : []),
-  ];
+  if (!groupIds?.length) return {};
 
-  return groupIdParam.length ? { groupId: groupIdParam } : {};
+  const hasEmpty = groupIds.includes('');
+  const nonEmpty = groupIds.filter((id) => id !== '');
+
+  return { groupId: hasEmpty ? [...nonEmpty, ''] : nonEmpty };
 };
 
 export const buildHostListParams = ({
--- a/src/components/SystemsView/utils/buildHostListParams.ts
+++ b/src/components/SystemsView/utils/buildHostListParams.ts
@@ -1,0 +1,77 @@
+import {
+  ApiHostGetHostListSystemTypeEnum,
+  type ApiHostGetHostListParams,
+  ApiHostGetHostListOrderByEnum,
+} from '@redhat-cloud-services/host-inventory-client/ApiHostGetHostList';
+import type { InventoryFilters } from '../filters/SystemsViewFilters';
+import type { LastSeenCustomRange } from '../DataViewFiltersContext';
+import type { SortDirection } from '../SystemsView';
+import { buildSystemProfileFilters } from './buildSystemProfileFilters';
+import { buildHostQueryOptions } from './buildHostListOptions';
+import { lastSeenKeysToApiParams } from './lastSeenKeysToApiParams';
+import { buildSystemType } from './buildSystemType';
+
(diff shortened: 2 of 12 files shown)


src/components/SystemsView/utils/buildHostViewsParams.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export interface BuildHostViewsParamsInput {
+  page: number;
+  perPage: number;
+  filters: InventoryFilters;
+  lastSeenCustomRange: LastSeenCustomRange;
+  sortBy: ApiHostViewsGetHostViewsOrderByEnum | undefined;
+  direction: SortDirection | undefined;
+}
```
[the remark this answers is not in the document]
raswonders: It fragile but not broken. I'm scheduling some work outside of this pr for improvements.

src/components/SystemsView/filters/SystemsViewFilters.tsx
```diff
@@ -26,7 +26,9 @@ import {
   HostGroupChipNode,
   useWorkspaceDisplayNames,
 } from '../hooks/useWorkspaceDisplayNames';
+import { ApiHostViewsGetHostViewsStalenessEnum } from '@redhat-cloud-services/host-inventory-client/ApiHostViewsGetHostViews';
 
+type Test = ApiHostViewsGetHostViewsStalenessEnum[];
```
johnsonm325: Is this used? Should it be removed?

src/components/SystemsView/utils/buildHostListParams.ts
```diff
@@ -0,0 +1,77 @@
...
+] as const;
+
+export interface BuildHostListParamsInput {
+  page: number;
+  perPage: number;
+  filters: InventoryFilters;
+  lastSeenCustomRange: LastSeenCustomRange;
+  sortBy?: ApiHostGetHostListOrderByEnum;
+  direction?: SortDirection;
+}
```
johnsonm325, line 28: Is it worth extracting this into a types file?

src/components/SystemsView/utils/buildHostListParams.ts
```diff
@@ -0,0 +1,77 @@
...
+const buildGroupIdParam = (
+  groupIds: string[] | undefined,
+): Pick<ApiHostGetHostListParams, 'groupId'> | Record<string, never> => {
+  const groupIdParam = [
+    ...(groupIds ?? []).filter((id) => id),
+    ...((groupIds ?? []).includes('') ? [''] : []),
+  ];
+
+  return groupIdParam.length ? { groupId: groupIdParam } : {};
+};
```
johnsonm325, line 39: Thoughts on this simplification?

```
  const buildGroupIdParam = (
    groupIds: string[] | undefined,
  ): Pick<ApiHostGetHostListParams, 'groupId'> | Record<string, never> => {
    if (!groupIds?.length) return {};

    const hasEmpty = groupIds.includes('');
    const nonEmpty = groupIds.filter(id => id !== '');

    return { groupId: hasEmpty ? [...nonEmpty, ''] : nonEmpty };
  };
```

src/components/SystemsView/utils/buildHostViewsParams.ts
johnsonm325, line 40: Same here, would it be worth extracting into the same types file as the other one?

src/components/SystemsView/utils/buildSystemProfileFilters.ts
```diff
@@ -0,0 +1,48 @@
...
+} from './operatingSystemSelectOptions';
+import {
+  buildWorkloadsFilter,
+  type WorkloadsPresenceFilter,
+} from './workloadsFilter';
+
+export type SystemProfileFilterInput = Pick<
+  InventoryFilters,
+  'rhcStatus' | 'operating_system' | 'workloads'
+>;
```
johnsonm325, line 14: extract to types file?

src/components/SystemsView/utils/buildSystemProfileFilters.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/** Nested `filter.system_profile` fragment for host list and host-view APIs. */
+export type SystemProfileFilter = {
+  rhc_client_id?: string[];
+  operating_system?: OperatingSystemProfileFilter;
+  workloads?: WorkloadsPresenceFilter;
+};
```
johnsonm325, line 21: extract to types file?

johnsonm325 (approved):
Looks good. Only a few questions and one possible simplification of a function, but nothing that would hold up an approval.

src/components/SystemsView/filters/SystemsViewFilters.tsx
raswonders: good spot this was just for me, I'll remove it

src/components/SystemsView/utils/buildHostListParams.ts
raswonders: That is not ideal. This interface BuildHostListParamsInput is unique to builder buildHostListParams. And as such it should stick right next to it in same file 

We could make it so we share part (everything but sortBy) with other builder buildHostListParams. BUT this type would be more deep&complex => harder to understand and track. (multiple files)

There is a quite famous "rule of collocation" which I try to adhere to. 
"Keep types as close as possible to the code that consumes them."

src/components/SystemsView/utils/buildHostListParams.ts
raswonders: looks equivalent and more clear I'll use it

src/components/SystemsView/utils/buildSystemProfileFilters.ts
raswonders: It should stay both types describe what buildSystemProfileFilters takes and returns. Rule of collocation still applies. There is no reason why it should be farther from the definition it wouldn't give us much.

src/components/SystemsView/utils/buildSystemProfileFilters.ts
raswonders: It should stay both types describe what buildSystemProfileFilters takes and returns. Rule of collocation still applies. There is no reason why it should be farther from the definition it wouldn't give us much.

PR #3286, opened 2026-06-23
feat(RHINENG-22169): enhance column management modal

raswonders:
RHINENG-22169

## What
Introduces several improvements to ColumnManagementModal

- updates description
- enables dragAndDrop
- shrinks spacing between draggable icon and checkbox
- moves action buttons to Modal footer so they're not scrolled with list content
- changes listItem layout and renders both title & appName for each column
- adds detection of column changes before it allows to save
- resetToDefault button resets also order

And fixes
- updates each column def with appName
- fixes initial columns filtering bug, which caused SystemsView not render any columns

Thing which isn't being delivered in this PR:

list item for "Name" even though is not selectable is still draggable. It's interplay of two different systems and each system is handled by different Patternfly component. Figuring it out without making mess of it, seems little too much for scope of this change I create a card where this gets fixed.

## Why
This is part of Phase I of InventoryViews.

## How
Because patternfly orig. component doesn't allow given changes, we created temporary  local copy of ListManager. I Will follow up and push changes to Patternfly so we can remove it.

## Testing
1. Navigate Inventory / Systems with preview ON
2. Make sure you have enabled also inventoryViews by running localStorage.setItem("ui.inventory-views", "true") in your console
3. Observe Modal's behavior

image

## Summary by Sourcery

Enhance the column management modal for inventory views with drag-and-drop reordering, footer-based actions, and app-specific column metadata.

New Features:
- Support drag-and-drop reordering in the column management modal via a locally vendored ListManager component.
- Display the owning app name alongside each column title in the column management modal.
- Disable the save action until column visibility or order has changed.

Enhancements:
- Refine the column management modal layout to use PatternFly ModalHeader/Body/Footer, keeping primary actions fixed while the list scrolls.
- Unify column definitions across inventory consumer apps with a stricter typed appName field and propagate appName to all relevant column configs.


raswonders:
@sourcery-ai review

computercamplove:
image
@raswonders should "Reset to default" reset positions of columns too, not only checkboxes?

computercamplove:
image
@raswonders  shouldn't any columns be before "Name" column? or it is expected?

raswonders:
> image @raswonders shouldn't any columns be before "Name" column? or it is expected?

as description explained this is expected and will be tackled in future PR. Column is still sticky, but it probably shouldn't move. The fix isn't easy nor fast so I decided to create separate PR for it.

src/components/SystemsView/hooks/useColumns.tsx
```diff
@@ -39,16 +39,7 @@ export const useColumns = ({
...
-      .map((col) => ({ ...col }))
-      .filter((col) => {
-        if (isInventoryViewsEnabled) {
-          return true;
-        }
-
-        const isConsumerAppColumn = 'appName' in col && col.appName;
-        return !isConsumerAppColumn;
-      }),
+    initialColumns.map((col) => ({ ...col })),
```
bastilian: ... isn't that a lil' bit redundant. 😉

src/components/SystemsView/ColumnManagementModalContext.tsx
```diff
@@ -48,12 +48,13 @@ export const ColumnManagementModalProvider = ({
       {isOpen && (
         <ColumnManagementModal
           appliedColumns={columns}
+          defaultColumns={defaultColumns.map((column) => ({ ...column }))}
```
bastilian: 😉

src/components/ListManager/ListManager.tsx
```diff
@@ -0,0 +1,312 @@
...
+    column: ListManagerItem & { id: string },
+    index: number,
+  ) => (
+    <>
+      <DataListControl>
+        {renderColumnCheck(column, index, true)}
+      </DataListControl>
+      {renderColumnCells(column, index)}
+    </>
+  );
```
bastilian, line 223: Doing things like this can (very) negatively affect React rendering. These should be proper components outside the function body, preferably in separate component files and called as JSX within the return/render.

src/components/SystemsView/ColumnManagementModalContext.tsx
raswonders: yes good spot! I've avoided need for copy by casting api to readonly.

src/components/SystemsView/hooks/useColumns.tsx
raswonders: yes, fixed thanks

src/components/ListManager/ListManager.tsx
raswonders: Would you care going into details where and how big of a slowdown in rendering is happening? I have some ideas but they all seem negligible especially at the size of our list (est. 28-50 items max) 

This component is also a temporary copy of Patternfly's original. We need it until we merge functional improvements to Patternfly. I tried to stick to patterns which it used before so I don't unnecessarily diverge.

src/components/ColumnManagementModal/ColumnManagementModal.tsx
```diff
@@ -88,22 +102,50 @@ export function ColumnManagementModal<
...
+      return true;
+    }
+
+    return applied.some(
+      (column, index) =>
+        column.key !== current[index].key ||
+        column.isShown !== current[index].isShown,
+    );
+  }, [appliedColumns, currentColumns]);
+
```
johnsonm325, line 119: Love this implementation!

src/components/ColumnManagementModal/ColumnManagementModal.tsx
```diff
@@ -88,22 +102,50 @@ export function ColumnManagementModal<
...
+      const column = currentByKey.get(defaultColumn.key);
+      if (column) {
+        resetColumns.push({
+          ...column,
+          isShown: defaultColumn.isShownByDefault ?? false,
+        });
+      }
+    }
+
+    setCurrentColumns(resetColumns);
```
johnsonm325: Maybe I'm missing something, so please explain if there's something going on here that I don't understand, but it seems to me that when a user clicks Reset to default we could simply set the selection equal to default columns which we have hard-coded. Claude suggests:

```
  const resetToDefault = () => {
    const source = defaultColumns ?? appliedColumns;
    setCurrentColumns(source.map(col => ({
      ...col,
      isShown: col.isShownByDefault ?? false
    })));
  };
```

This is the interaction as I understand it:

```
  ┌──────────────────┬─────────────────────────────────────────────────────────────────┐
  │      Button      │                            Behavior                             │
  ├──────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Cancel           │ Discard currentColumns, close modal → reverts to appliedColumns │
  ├──────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Reset to default │ Set currentColumns = defaultColumns (factory reset)             │
  ├──────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Save             │ Apply currentColumns → becomes new appliedColumns               │
  └──────────────────┴─────────────────────────────────────────────────────────────────┘
```

Let me know what you think.

johnsonm325 (approved):
I'm approving because I don't want to hold you up tomorrow (Thursday) morning with getting this merged, but I'd really like to see that reset to default logic simplified if it all works the way I suspect it does.

src/components/ListManager/ListManager.tsx
johnsonm325: I'm fine with this as long as we make a card to update the PF component so we can get rid of this implementation.

src/components/ColumnManagementModal/ColumnManagementModal.tsx
```diff
@@ -161,38 +203,70 @@ export function ColumnManagementModal<
     onClose({} as KeyboardEvent);
   };
 
+  const handleSaveClick = () => {
+    handleSave(listManagerItems);
+  };
+
   return (
     <Modal
```
johnsonm325, line 199: We should also set the `maxHeight` to 60% or something. I think with the narrowness of this modal and the fact that it will always span the entire height of the page because of the number of columns, we can make this look a little better.

Image

sbuenafe-rh: Setting the modal max height to 60% sounds good. I don't currently see any issues with making that call.

src/components/ListManager/ListManager.tsx
bastilian: > Would you care going into details where and how big of a slowdown in rendering is happening?

This is was just a general note to avoid things like that. It is a bad practice and there are numerous blog posts that describe the issue in more detail. For example: https://www.xjavascript.com/blog/it-s-okay-to-declare-other-components-inside-a-component/#performance-issues-component-identity-and-unnecessary-re-renders

It's these little things that make our apps feel sluggish overall.

bastilian: > This component is also a temporary copy of Patternfly's original. We need it until we merge functional improvements to Patternfly.

My personal opinion... I would not put in temporary code, because we are waiting for Patternfly to merge something. Help get the changes you need merged in Patternfly first and then finish work in the app. "temporary" is the most permanent, and creating a card for it is just theatre. :D

raswonders: I agree ideally we would do upstream first, but it's hard to guarantee timely delivery of anything this way. I find the keeping temporary patched versions good enough so far. We need to streamline to tracking and cleanup though, thats the hardest part.

src/components/ColumnManagementModal/ColumnManagementModal.tsx
raswonders: much better and works gonna use it

src/components/ColumnManagementModal/ColumnManagementModal.tsx
raswonders: I'm happy to do so and I just tried few times but it isn't straight-forward. I probably need some extra time to figure this out. I cannot easily style it because everything in /src is prepended with .inventory class. Thing is thought we portal Modal outside of this container. Not sure what the good solution to this is.

PR #3287, opened 2026-06-24
test: Increase workers in playwright CI

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-28057

## What
Increase Playwright workers from 1 to 4 in CI, reducing test execution time by ~50%.

 Verified across 3 CI workflows (only PW test run time, not CI setup time):
  - E2E: Full Test Suite - 38 tests (~10.5m > ~5.1m)
  - E2E: InventoryViews- 30 tests (~7.3m > ~3.6m)
  - E2E: SystemsView - 16 tests (~5.0m > ~3.6m)

## Why
Reduce PR check test time

## How

- ` playwright.config.js`: workers: isCI ? 4 : 1
-  `test_staleness.test.ts`: Serial mode (modifies global account settings)
-  `test_system.test.ts` : Serial mode (/hosts CRUD)
- All other tests run in parallel via existing `fullyParallel: true` config

Risk Mitigation:
  - Auth setup remains serialized (via dependencies in config)
  - Global setup runs once before workers start
  - Only 2 test suites require serial execution
  - All mutation tests create isolated resources
  -  Tests run on same `instance-size:large` (not 4 instances)
  - Shorter runtime = lower AWS costs

## Testing
All 3 CI should pass


adarshdubey-star (approved):
Let's try this : )

raswonders:
Cannot make tests pass locally

`npx playwright test --workers=4 --grep-invert "@integration|@inventory-views"` is this correct command to test this with?

raswonders (approved):
lgtm

computercamplove:
> Cannot make tests pass locally
> 
> `npx playwright test --workers=4 --grep-invert "@integration|@inventory-views"` is this correct command to test this with?

yes, if you see failed tests - it is due flakiness (unless you made the change that it could break it). The difference in local testing is that here's no reties like we have in our CIs when test failed.

PR #3289, opened 2026-06-25
fix(e2e): drop JS sort verification from sorting tests

adarshdubey-star:
Jira
[RHINENG-28080](https://redhat.atlassian.net/browse/RHINENG-28080)

What
fix(e2e): drop JS sort verification from sorting tests

[RHINENG-28080]: https://redhat.atlassian.net/browse/RHINENG-28080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_playwright-tests/test_system.test.ts
+++ b/_playwright-tests/test_system.test.ts
@@ -291,27 +291,14 @@ test.describe('System Sorting', { tag: ['@systems-table'] }, () => {
         }).toPass({ timeout: 5000 });
       });
 
-      await test.step('Verify systems are sorted alphabetically', async () => {
+      await test.step('Verify systems are displayed', async () => {
         const nameCell = page
           .locator('[data-ouia-component-id="systems-view-table-td-0-0"]')
           .or(page.locator('td[data-label="Name"]'));
         await expect(nameCell.first()).toBeVisible();
 
         const displayedNames = await nameCell.allTextContents();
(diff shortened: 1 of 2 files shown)


raswonders (commented):
I still think we shouldn't be performing JS sorting in tests. 

What I think good asserts here are: 
1. assert that when I sent specific sorting params, server responds with OK and response containing systems. 
2. assert that systems from response I just got are being displayed in table.

_playwright-tests/test_system.test.ts
```diff
@@ -300,8 +303,10 @@ test.describe('System Sorting', { tag: ['@systems-table'] }, () => {
         );
 
         const sortFunctions = {
-          ascending: (_a: string, _b: string) => _a.localeCompare(_b),
-          descending: (_a: string, _b: string) => _b.localeCompare(_a),
+          ascending: (_a: string, _b: string) =>
```
raswonders: We seem to use same thing on three different places, I'd refactor it to helpers.

raswonders (approved):
This seems to fix sorting issue as prescribed. I'm gonna ok this but If this will fail on collation issue again we should consider removing sorting from these tests. 

There is also opportunity to refactor/reuse sorting function defs. But its not a blocker.

computercamplove (approved):

PR #3291, opened 2026-06-26
test(RHINENG-26765): Test sticky columns, apply all columns

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-26765

## What
Test sticky columns, apply all columns

## Why
Requirement: Checkboxes, Name and Actions columns are sticky colums - stay in place while scrolling the table left-right

## How
New Helper Functions:
  - `isTableHorizontallyScrollable() `- Checks if table can scroll horizontally
  - `scrollTableToPosition(page, position)` - Scrolls table to specific position (0-1 range)
  - `isVisibleInViewport(element)` - Verifies element is actually visible in viewport (not just in DOM)

## Testing
  Test Coverage:
  - Enables all columns to create horizontal scroll
  - Verifies checkbox, Name, and Actions columns remain visible in viewport while scrolling
  - Tests scrolling to middle (50%) and maximum right (100%) positions

## Summary by Sourcery

Add end-to-end coverage for sticky inventory table columns and related helper utilities for horizontal scrolling behavior.

Enhancements:
- Introduce helper utilities to detect horizontal scrollability, scroll the inventory table to a relative position, and validate left/right sticky column properties and viewport visibility.
- Add JSDoc-style parameter annotations to various UI helper components and utilities for clearer typing and documentation.

Tests:
- Add a Playwright test that enables all inventory columns to force horizontal scrolling and verifies that checkbox, Name, and Actions columns remain visible while scrolling to middle and far-right positions.

### DIFF
--- a/_playwright-tests/helpers/columnHelpers.ts
+++ b/_playwright-tests/helpers/columnHelpers.ts
@@ -50,6 +50,15 @@ export const vulnerabilityColumns = [
   'CVEs with known exploits',
 ];
 
+export const allColumns = [
+  ...advisorColumns,
+  ...complianceColumns,
+  ...patchColumns,
+  ...malwareColumns,
+  ...inventoryColumns,
+  ...vulnerabilityColumns,
+];
+
 /**
  * Opens the 'Manage columns' modal from the systems view toolbar.
  * Wraps the action in toPass to handle loading skeletons and dropdown rendering.
--- a/_playwright-tests/helpers/columnHelpers.ts
+++ b/_playwright-tests/helpers/columnHelpers.ts
@@ -112,31 +112,6 @@ export async function isTableHorizontallyScrollable(
 }
 
 /**
- * Checks if an element is actually visible in the viewport (not just in the DOM).
- *  @param   {Locator}          element - The element locator to check.
- *  @returns {Promise<boolean>}         - True if the element is visible in the viewport.
- */
-export async function isVisibleInViewport(element: Locator): Promise<boolean> {
-  return await element.evaluate((el) => {
-    const rect = el.getBoundingClientRect();
-    const scrollContainer = document.querySelector(
-      '.ins-c-systems-view-table-scroll',
-    );
-    if (!scrollContainer) return false;
-
-    const containerRect = scrollContainer.getBoundingClientRect();
-
-    // Check if element is within the visible viewport of the scroll container
-    return (
-      rect.top >= containerRect.top &&
-      rect.left >= containerRect.left &&
-      rect.bottom <= containerRect.bottom &&
-      rect.right <= containerRect.right
-    );
-  });
-}
-
-/**
  * Scrolls the table horizontally to a specific position.
  *  @param {Page}   page     - The Playwright page instance.
  *  @param {number} position - Position to scroll to (0 = left, 0.5 = middle, 1 = right).
--- a/_playwright-tests/helpers/columnHelpers.ts
+++ b/_playwright-tests/helpers/columnHelpers.ts
@@ -86,6 +86,64 @@ export async function openManageColumnsModal(page: Page, timeout = 45000) {
 }
 
 /**
+ * Checks if the systems table is horizontally scrollable.
+ *  @param   {Page}             page - The Playwright page instance.
+ *  @returns {Promise<boolean>}      - True if the table is scrollable, false otherwise.
+ */
+export async function isTableHorizontallyScrollable(
+  page: Page,
+): Promise<boolean> {
+  return await page.evaluate(() => {
+    const scrollContainer = document.querySelector(
+      '.ins-c-systems-view-table-scroll',
+    );
+    if (!scrollContainer) return false;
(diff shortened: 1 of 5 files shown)


_playwright-tests/test_inventory_views.test.ts
```diff
@@ -239,4 +242,100 @@ test.describe('Inventory Views application columns', () => {
...
+          ...advisorColumns,
+          ...complianceColumns,
+          ...patchColumns,
+          ...malwareColumns,
+          ...inventoryColumns,
+          ...vulnerabilityColumns,
+        ];
+
+        for (const columnName of allColumns) {
+          await dialog.getByLabel(columnName, { exact: true }).check();
```
raswonders: there is a bulkSelect button which we can click instead of checking all them individually

_playwright-tests/test_inventory_views.test.ts
```diff
@@ -239,4 +242,100 @@ test.describe('Inventory Views application columns', () => {
...
+        expect(await isVisibleInViewport(checkboxHeader)).toBe(true);
+
+        await expect(nameHeader).toBeVisible();
+        expect(await isVisibleInViewport(nameHeader)).toBe(true);
+
+        await expect(actionsHeader).toBeVisible();
+        expect(await isVisibleInViewport(actionsHeader)).toBe(true);
+      });
+
+      await test.step('Scroll to maximum right and verify sticky columns still visible', async () => {
```
raswonders, line 296: This use case or the one when we scroll into the middle seems little extra. Is there a good reason we need both? I'd make sense to make this test smaller.

_playwright-tests/helpers/columnHelpers.ts
```diff
@@ -85,6 +85,64 @@ export async function openManageColumnsModal(page: Page, timeout = 45000) {
...
+    return scrollContainer.scrollWidth > scrollContainer.clientWidth;
+  });
+}
+
+/**
+ * Checks if an element is actually visible in the viewport (not just in the DOM).
+ *  @param   {Locator}          element - The element locator to check.
+ *  @returns {Promise<boolean>}         - True if the element is visible in the viewport.
+ */
+export async function isVisibleInViewport(element: Locator): Promise<boolean> {
```
raswonders: We could do without this function and use standard matcher.

Playwright has matcher called toBeInViewport it checks whether element is in viewport. It doesn't check whether its within our scrollable container but that shouldn't matter. Once column is detectable in viewport it means it was scrolled into view.

_playwright-tests/helpers/columnHelpers.ts
```diff
@@ -85,6 +85,64 @@ export async function openManageColumnsModal(page: Page, timeout = 45000) {
   }).toPass({ timeout });
 }
 
+/**
+ * Checks if the systems table is horizontally scrollable.
+ *  @param   {Page}             page - The Playwright page instance.
+ *  @returns {Promise<boolean>}      - True if the table is scrollable, false otherwise.
+ */
+export async function isTableHorizontallyScrollable(
```
raswonders, line 102: I like this helper.

raswonders (commented):
Test is passing. I think for single test though its little too long. Steps help readability little bit but it doesn't change that for Playwright engine its just one long test.

I left few pointers, addressing those should significantly shorten the test.

_playwright-tests/test_inventory_views.test.ts
```diff
@@ -239,4 +242,100 @@ test.describe('Inventory Views application columns', () => {
...
+        const isScrollable = await isTableHorizontallyScrollable(page);
+        expect(isScrollable).toBe(true);
+      });
+
+      await test.step('Scroll right and verify sticky columns remain visible while non-sticky scrolls out', async () => {
+        // Scroll to the middle
+        await scrollTableToPosition(page, 0.5);
+
+        // Non-sticky column should have scrolled out of viewport (proving scroll happened)
+        await expect(nonStickyColumn).toBeVisible(); // Exists in DOM
```
raswonders: is asserting toBeVisible before we assert element to be in viewport necessary? Why not asserting just toBeInViewport?

_playwright-tests/test_inventory_views.test.ts
computercamplove: @raswonders you are right, i can use it here

_playwright-tests/helpers/columnHelpers.ts
computercamplove: interesting, i need to check it out, thank you!

_playwright-tests/test_inventory_views.test.ts
computercamplove: hmm, i need to check toBeInViewport timeout it enough, but good point!

computercamplove:
@raswonders 

I've applied suggestion:
- Use bulk select to apply all columns
- use build-in `toBeInViewport()`;
- remove extra verefications

raswonders (approved):
looks good now thank you!

PR #3293, opened 2026-06-29
  feat(RHINENG-28009): add localStorage draft persistence for Inventory table state

johnsonm325:
## Jira
  [RHINENG-28009](https://redhat.atlassian.net/browse/RHINENG-28009)

  ## What
  Implements temporary draft persistence for SystemsView table configuration using localStorage. Users' table preferences (columns, filters, sorting,
  pagination) now survive page reloads.

  ## Why
  Enables users to maintain their preferred Inventory table configuration across sessions. This is a temporary localStorage solution until RHIN-2536
  (database-backed Inventory Views) is implemented.

  ## How
  - Created 4 new hooks: useInventoryOrgId, useBootstrapInventoryTableDraft, useHydrateDraftField, usePersistInventoryTableDraft
  - Added localStorage utilities for CRUD operations and URL param merging
  - Modified SystemsView to orchestrate bootstrap lifecycle with ready gate
  - Modified DataViewFiltersContext and useColumns to use shared hydration hook
  - Storage keyed per-org: `insights.inventory.systems-view.draft.{orgId}`
  - Added 5-second timeout fallback for production safety

  Lifecycle: Mount → fetch orgId → bootstrap (merge draft + URL) → hydrate fields → persist changes

  ## Testing
  - Verified column visibility/order persists across page reload
  - Verified filters and sorting persist across page reload
  - Verified URL params override draft (deep links work correctly)
  - Verified clear filters also clears draft from localStorage
  - Verified timeout fallback works when bootstrap hangs
  - Build passes with no TypeScript errors

[RHINENG-28009]: https://redhat.atlassian.net/browse/RHINENG-28009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add per-organization localStorage-backed draft persistence for SystemsView table state and gate rendering on a bootstrap-ready lifecycle.

New Features:
- Persist SystemsView columns, filters, sorting, and pagination across sessions using org-scoped localStorage drafts.
- Introduce hooks to bootstrap, hydrate, and persist inventory table draft state based on URL params and readiness.
- Support hydrating specific table fields (e.g., custom last seen range, columns) from stored drafts while respecting deep-link URLs.

Bug Fixes:
- Ensure SystemsView degrades gracefully with a timeout fallback if draft bootstrap hangs and avoids rendering before org context is available.
- Make clearing filters also clear persisted draft filter state while preserving column preferences.

Enhancements:
- Refactor filter defaults into a shared inventoryFilterDefaults module and reuse them across draft storage utilities.
- Add robust utilities for reading, writing, merging, and applying inventory table draft state, including org ID extraction from chrome identity.

Tests:
- Add unit tests for inventory draft storage utilities, including URL merging, column merging, filter clearing, and org scoping.
- Add tests for the draft bootstrap hook to verify URL hydration behavior and readiness signalling.

### DIFF
--- a/_playwright-tests/test_system.test.ts
+++ b/_playwright-tests/test_system.test.ts
@@ -7,6 +7,7 @@ import { isSystemsViewEnabled } from './helpers/constants';
 
 test.describe('System CRUD', { tag: ['@systems-table'] }, () => {
   test.describe.configure({ mode: 'serial' });
+
   test('User should be able to edit and delete a system from Systems page', async ({
     page,
   }) => {
--- a/src/components/SystemsView/DataViewFiltersContext.tsx
+++ b/src/components/SystemsView/DataViewFiltersContext.tsx
@@ -12,24 +12,16 @@ import { normalizeLastSeenFilterValue } from './constants';
 import { useConditionalRBAC } from '../../Utilities/hooks/useConditionalRBAC';
 import { GENERAL_GROUPS_READ_PERMISSION } from '../../constants';
 import { useUngroupedWorkspaceId } from '../../hooks/useUngroupedWorkspaceId';
+import { clearInventoryTableDraftFilters } from './utils/inventoryTableDraftStorage';
+import { INITIAL_INVENTORY_FILTERS } from './inventoryFilterDefaults';
+import { useHydrateDraftField } from './hooks/useHydrateDraftField';
 
 export type LastSeenCustomRange = {
   start?: string;
   end?: string;
 } | null;
 
-export const INITIAL_INVENTORY_FILTERS: InventoryFilters = {
-  hostname_or_id: '',
-  status: [],
-  source: [],
-  rhcStatus: [],
-  system_type: [],
-  group_id: [],
-  last_seen: '',
-  tags: [],
-  operating_system: [],
-  workloads: [],
-};
+export { INITIAL_INVENTORY_FILTERS };
 
 export interface DataViewFiltersContextValue {
   filters: InventoryFilters;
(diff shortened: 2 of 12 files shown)


raswonders (commented):
@johnsonm325 this might be too extensive of a change for what is needed.
It also feels very complex for temporary change and scopes inside things which we don't really need at the very moment. The filters/sorting state already persists in URL and I believe thats what we want.

What was missing was capability of keeping columns setting across reload/refresh. Using localStorage as naive but sufficient approach should be no more than 100 lines in single file and should be relatively easy to remove. 
I quickly oneshot something through AI, please have a look to understand what I mean. https://github.com/RedHatInsights/insights-inventory-frontend/pull/3294

This should also be just SystemsView change. We should be avoiding anything which couples InventoryViews and SystemsView any further until they're refactored apart. Otherwise it will just get harder and more bug prone to do it.

I'll  do this refactoring asap after we do the SPUR, as it seems like we really need it.

I'd love to hear what you think and what next steps you would like to take.

johnsonm325:
@raswonders Yes, I tested it out and completely agree with you, it's much more simple and does the same thing. Once you read this message, please close this card and take yours off draft so we can get it merged.

At one point Claude told me that we only needed to implement for columns but then appeared to implement for everything. I just got so caught up in multiple projects that I never circled back around to verify it. Thanks for taking a look.

raswonders:
As suggested I'm closing this one and feature will be delivered here: https://github.com/RedHatInsights/insights-inventory-frontend/pull/3294

PR #3294, opened 2026-06-30
feat(RHINENG-28009): persist column settings

raswonders:
RHINENG-28009

## What
It Persists column settings temporarily inside localStorage. It saves them under acc&user specific key so they don't collide and allow multiple users. 

NOTE: This is temporary solution and will be superseeded by views management.

## Testing
1. Go to InventoryViews
2. change column settings and save
3. Refresh or navigate away from the page and come back
4. Make sure settings persisted
5. Try log in as different user, and make sure its settings are separate from first one

## Summary by Sourcery

Persist Systems View column configuration per user using localStorage and keep the column management UI in sync without overwriting unsaved edits.

New Features:
- Store Systems View column visibility and ordering in localStorage under an account- and user-specific key so column preferences persist across sessions.

Enhancements:
- Introduce a reusable hook to load, merge, and persist column preferences with sensible fallbacks when user identity or storage is unavailable.
- Update the column management modal to only resync from applied columns when there are no unsaved changes, preventing user edits from being overwritten.

### DIFF
--- a/src/components/ColumnManagementModal/ColumnManagementModal.tsx
+++ b/src/components/ColumnManagementModal/ColumnManagementModal.tsx
@@ -92,16 +92,6 @@ export function ColumnManagementModal<
     })),
   );
 
-  // Sync with appliedColumns when they change
-  useEffect(() => {
-    setCurrentColumns(
-      appliedColumns.map((column) => ({
-        ...column,
-        isShown: column.isShown ?? column.isShownByDefault,
-      })),
-    );
-  }, [appliedColumns]);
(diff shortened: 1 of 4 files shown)


computercamplove:
@raswonders everything works as expected. Before we merge, could we add some test coverage for the
  `localStorage` persistence logic? Even just a simple unit test that verifies:
  - column settings are saved with the correct key (acc+user specific)
  - settings are properly retrieved on load

computercamplove (approved):

PR #3305, opened 2026-07-02
fix: hide Global tag filter when SystemsView is enabled

computercamplove:
## What
hide Global tag filter when SystemsView is enabled

## Why
`SystemsView` table doesn't support Global tag filter

## How
Check for feature flag

## Testing
Stage: flag is enabled
- visit https://stage.foo.redhat.com:1337/
- navigate Systems page
- check for global tag filter
- shouldn't be visible since stage has enabled feature flag for systems view

## Summary by Sourcery

Bug Fixes:
- Ensure the global tag filter is hidden when SystemsView is enabled and restored when the view or feature flag is disabled.

### DIFF
--- a/src/routes/InventoryViews.tsx
+++ b/src/routes/InventoryViews.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext, useLayoutEffect, useState } from 'react';
 import { Flex, FlexItem, PageSection } from '@patternfly/react-core';
 import SystemsView from '../components/SystemsView';
 import {
@@ -10,6 +10,7 @@ import { OutageAlert } from '../components/OutageAlert';
 import SystemsViewToggle from '../components/SystemsView/SystemsViewToggle';
 import { AccountStatContext } from '../Contexts';
 import { ImagesView } from '../components/InventoryViews/ImagesView/ImagesView';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 import './inventory.scss';
 
 interface InventoryViewsProps {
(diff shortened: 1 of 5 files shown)


src/routes/InventoryViews.tsx
```diff
@@ -21,6 +23,18 @@ export type View = 'systems' | 'images';
 const InventoryViews = ({ hasAccess }: InventoryViewsProps) => {
   const [view, setView] = useState<View>('systems');
   const { hasBootcImages } = useContext(AccountStatContext);
+  const chrome = useChrome();
+  const isSystemsViewEnabled = useSystemsViewFeatureFlag();
+
+  useLayoutEffect(() => {
```
computercamplove, line 27: @raswonders @johnsonm325 need your opinion on using `useLayoutEffect` instead of `useEffect` - the reason i used it here - it prevents global tag flicker on top on loading. It is not critical issue, in case i need to use `useEffect`. Just wonder if `useLayoutEffect` somehow not good in this case.

computercamplove: getting suggestion here too:

> The useLayoutEffect hook seems unnecessary here since there’s no layout measurement or synchronous DOM read/write; consider using useEffect instead to avoid blocking layout unnecessarily.

raswonders: useLayoutEffect is commonly used for flicker problems so it's ok. Problem I see though that we're checking for SystemsView here. InventoryViews won't render without useSystemsViewsFeatureFlag being on so I think its useless.

computercamplove: nice catch!

raswonders (approved):
lgtm

PR #3310, opened 2026-07-03
test(RHINENG-25382): unit test for useBulkSelect

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-25382

## What
unit test for `useBulkSelect`

## Why
require to release SystemsView to prod-stable

## How
all critical paths remain fully tested:
  - state calculations (empty, partial, full selection)
  - all `onBulkSelect` behaviors across different sources and values
  - edge cases and real-world scenarios
  - callback memoization

## Summary by Sourcery

Add comprehensive unit test coverage for the useBulkSelect hook in SystemsView, exercising selection state, bulk selection behaviors, and edge cases.

Tests:
- Add extensive useBulkSelect hook tests covering empty, partial, full, and cross-page selection states.
- Verify onBulkSelect handling for all BulkSelectValue options, sources, and unknown inputs.
- Test edge and real-world scenarios including empty data, total mismatches, mixed-page selections, and rapid successive calls.

### DIFF
--- a/src/components/SystemsView/hooks/useBulkSelect.test.ts
+++ b/src/components/SystemsView/hooks/useBulkSelect.test.ts
@@ -212,18 +212,6 @@ describe('useBulkSelect', () => {
       expect(result.current.isPartiallySelected).toBe(false);
     });
 
-    it('handles total being 0', () => {
-      const rows = createTestItems(5);
-      const selection = createMockSelection([]);
-
-      const { result } = renderHook(() =>
-        useBulkSelect({ selection, rows, total: 0 }),
-      );
-
-      expect(result.current.selectedCount).toBe(0);
-      expect(result.current.isPartiallySelected).toBe(false);
-    });
-
     it('handles selection count exceeding total', () => {
       const rows = createTestItems(5);
       const selectedItems = createTestItems(10);
@@ -277,44 +265,6 @@ describe('useBulkSelect', () => {
       expect(result.current.isPartiallySelected).toBe(true);
     });
 
-    it('correctly identifies page as selected even with items from other pages selected', () => {
-      const rows = createTestItems(5);
-      const otherItems = [
-        {
-          id: 'other-1',
-          name: 'Other 1',
-        },
-      ];
-      const selection = createMockSelection([...rows, ...otherItems]);
-
-      const { result } = renderHook(() =>
-        useBulkSelect({ selection, rows, total: 20 }),
-      );
-
-      expect(result.current.isPageSelected).toBe(true);
-      expect(result.current.selectedCount).toBe(6);
-    });
-
-    it('updates callback when dependencies change', () => {
-      const rows = createTestItems(5);
-      const selection = createMockSelection([]);
-
-      const { result, rerender } = renderHook(
-        ({ selection, rows, total }) =>
-          useBulkSelect({ selection, rows, total }),
-        {
-          initialProps: { selection, rows, total: 10 },
-        },
-      );
-
-      const firstCallback = result.current.onBulkSelect;
-
-      const newSelection = createMockSelection(rows.slice(0, 2));
-      rerender({ selection: newSelection, rows, total: 10 });
-
-      expect(result.current.onBulkSelect).not.toBe(firstCallback);
-    });
-
     it('handles rapid successive onBulkSelect calls', async () => {
       const rows = createTestItems(5);
       const selection = createMockSelection([]);
@@ -361,28 +311,6 @@ describe('useBulkSelect', () => {
       await result.current.onBulkSelect(BulkSelectValue.page, 'checkbox');
 
       expect(selection.setSelected).toHaveBeenCalledWith([]);
-    });
-
-    it('handles navigation between pages maintaining selections', () => {
-      const page1Rows = createTestItems(10);
-      const page1Selected = page1Rows.slice(0, 5);
-      const selection = createMockSelection(page1Selected);
-
-      const { result, rerender } = renderHook(
-        ({ rows }) => useBulkSelect({ selection, rows, total: 50 }),
-        { initialProps: { rows: page1Rows } },
-      );
-
-      expect(result.current.isPageSelected).toBe(false);
-
-      const page2Rows = createTestItems(10).map((item, i) => ({
-        ...item,
-        id: `page2-${i}`,
-      }));
-      rerender({ rows: page2Rows });
-
-      expect(result.current.selectedCount).toBe(5);
-      expect(result.current.isPageSelected).toBe(false);
     });
   });
 });
--- a/src/components/SystemsView/hooks/useBulkSelect.test.ts
+++ b/src/components/SystemsView/hooks/useBulkSelect.test.ts
@@ -178,23 +178,6 @@ describe('useBulkSelect', () => {
         expect(selection.onSelect).toHaveBeenCalledWith(true, rows);
         expect(selection.setSelected).not.toHaveBeenCalled();
       });
-
-      it('does nothing when source is not dropdown or checkbox', async () => {
-        const rows = createTestItems(5);
-        const selection = createMockSelection([]);
-
-        const { result } = renderHook(() =>
-          useBulkSelect({ selection, rows, total: 10 }),
-        );
-
-        await result.current.onBulkSelect(
-          BulkSelectValue.page,
-          'unknown' as BulkSelectSource,
-        );
-
-        expect(selection.onSelect).not.toHaveBeenCalled();
-        expect(selection.setSelected).not.toHaveBeenCalled();
-      });
     });
   });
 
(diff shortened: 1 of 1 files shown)


src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+};
+
+const createTestItems = (count: number): TestItem[] =>
+  Array.from({ length: count }, (_, i) => ({
+    id: `item-${i}`,
+    name: `Item ${i}`,
+  }));
+
+describe('useBulkSelect', () => {
+  describe('fundamental behaviors', () => {
```
raswonders, line 36: we are checking isPageSelected and isPartiallySelected state combinations, but we  never check isFullySelected in here.

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+          useBulkSelect({ selection, rows, total: 5 }),
+        );
+
+        await result.current.onBulkSelect(BulkSelectValue.page, 'checkbox');
+
+        expect(selection.onSelect).toHaveBeenCalledWith(true, rows);
+        expect(selection.setSelected).not.toHaveBeenCalled();
+      });
+
+      it('does nothing when source is not dropdown or checkbox', async () => {
```
raswonders: What guarantees this test gives us? To me it doesn't seem very helpful.

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+          BulkSelectValue.page,
+          'unknown' as BulkSelectSource,
+        );
+
+        expect(selection.onSelect).not.toHaveBeenCalled();
+        expect(selection.setSelected).not.toHaveBeenCalled();
+      });
+    });
+
+    describe('other/unknown values', () => {
```
raswonders: We don't need this as TS makes sure call sites only ever use BulkSelectValue 

Image

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+        await result.current.onBulkSelect('unknown' as BulkSelectValue);
+
+        expect(selection.onSelect).not.toHaveBeenCalled();
+        expect(selection.setSelected).not.toHaveBeenCalled();
+      });
+    });
+  });
+
+  describe('edge cases', () => {
+    it('handles empty rows array', () => {
```
raswonders: this comments could be more specific. We could say how it handles the situation not just that it handles it. 

For instance it could say something: "selects no rows when selection is empty"

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+      const { result } = renderHook(() =>
+        useBulkSelect({ selection, rows, total: 0 }),
+      );
+
+      expect(result.current.selectedCount).toBe(0);
+      expect(result.current.isPageSelected).toBe(false);
+      expect(result.current.isPartiallySelected).toBe(false);
+    });
+
+    it('handles total being 0', () => {
```
raswonders: this is more of a test of empty selection than the total === 0. Empty selection is also tried in test before.

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+
+      const { result } = renderHook(() =>
+        useBulkSelect({ selection, rows, total: 0 }),
+      );
+
+      expect(result.current.selectedCount).toBe(0);
+      expect(result.current.isPartiallySelected).toBe(false);
+    });
+
+    it('handles selection count exceeding total', () => {
```
raswonders, line 198: Isn't this just arbitrary test can this actually happen in the code?

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+      const { result } = renderHook(() =>
+        useBulkSelect({ selection, rows, total: 20 }),
+      );
+
+      expect(result.current.selectedCount).toBe(3);
+      expect(result.current.isPageSelected).toBe(false);
+      expect(result.current.isPartiallySelected).toBe(true);
+    });
+
+    it('correctly identifies page as selected even with items from other pages selected', () => {
```
raswonders: this one maybe extra too? I guess it would be enough If we only have test checking that isPageSelected is true if whole page is selected. That should be enough.

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+
+      const { result } = renderHook(() =>
+        useBulkSelect({ selection, rows, total: 20 }),
+      );
+
+      expect(result.current.isPageSelected).toBe(true);
+      expect(result.current.selectedCount).toBe(6);
+    });
+
+    it('updates callback when dependencies change', () => {
```
raswonders: Seems like extra as well ... useCallback with its dependency array protects against this, it feels like we're testing React library here.

src/components/SystemsView/hooks/useBulkSelect.test.ts
```diff
@@ -0,0 +1,404 @@
...
+      const { result } = renderHook(() =>
+        useBulkSelect({ selection, rows, total: 50 }),
+      );
+
+      await result.current.onBulkSelect(BulkSelectValue.page, 'checkbox');
+
+      expect(selection.setSelected).toHaveBeenCalledWith([]);
+    });
+
+    it('handles navigation between pages maintaining selections', () => {
```
raswonders: this hook doesn't own selection persistence in table, that makes this test quite out of place.

raswonders (commented):
Thank you for tests, coverage seem good, but I'd skip some of these as they feel redundant or not providing any real value. 

Please go over the notes let me know what you think and when you're ready for another review.

src/components/SystemsView/hooks/useBulkSelect.test.ts
computercamplove: @raswonders if we test `isPartiallySelected` when it `false`, why need to test `isFullySelected`? To test if all selected in the page we test `isPageSelected`.

src/components/SystemsView/hooks/useBulkSelect.test.ts
computercamplove: make sense:D thank you!

src/components/SystemsView/hooks/useBulkSelect.test.ts
computercamplove: @raswonders for example filter is applied or some selected systems are removed

src/components/SystemsView/hooks/useBulkSelect.test.ts
computercamplove: @raswonders gonna drop it here, but we don't have tests for `SystemsView` page chnages

src/components/SystemsView/hooks/useBulkSelect.test.ts
raswonders: Yes was wrong on this one, it struck me as weird we don't assert ever against this state but looking again I realized isFullySelected is only tracked internally. So no need.

Image

src/components/SystemsView/hooks/useBulkSelect.test.ts
raswonders: that's fine :) we're gonna have them

src/components/SystemsView/hooks/useBulkSelect.test.ts
computercamplove: @raswonders you are right - it is 2 typed. for some reason when i navigated cursor first time on `BulkSelectSource` it showed me "undefined" there too -_- but when i searched for `export type BulkSelectSource ` it started to show me right type in hints 
image

src/components/SystemsView/hooks/useBulkSelect.test.ts
raswonders: ah that's true! disregard then

src/components/SystemsView/hooks/useBulkSelect.test.ts
computercamplove: probably some visual code issue hint issue:)

computercamplove:
@raswonders i applied all suggestions expect 2 you agree not to do anything. Thank you for review!

src/components/SystemsView/hooks/useBulkSelect.test.ts
raswonders: yeah types are always our first line of defense :)

raswonders (approved):
looks very good now, thanks for considering the notes!

PR #3311, opened 2026-07-03
refactor: add useLegacyInventoryTableFeatureFlag

raswonders:
## What
- adds separate hook for LegacyIventoryTableFeatureFlag
- decouples two hooks
- makes override logic explicit in router rather than inside of a hook

## Why
- it keeps the established hook pattern
- makes hook influence clear and explicit at calling sites

## Summary by Sourcery

Introduce a dedicated legacy inventory table hook and move override logic into the router and tests.

Enhancements:
- Add a useLegacyInventoryTable hook to manage the legacy inventory table feature flag consistently.
- Simplify useSystemsViewFeatureFlag by removing legacy inventory override handling and keeping it focused on systems view state.
- Adjust route selection logic to explicitly prefer the legacy InventoryPage when the legacy inventory table flag is enabled, even if systems view is enabled.

Tests:
- Update Playwright test helpers and constants to use the new legacy inventory table flag naming and behavior, removing coupled overrides of other feature flags.

### DIFF
--- a/_playwright-tests/auth.setup.ts
+++ b/_playwright-tests/auth.setup.ts
@@ -16,7 +16,7 @@ import {
 import {
   isSystemsViewEnabled,
   isInventoryViewsEnabled,
-  forceLegacyInventoryTable,
+  isLegacyInventoryTableEnabled,
 } from './helpers/constants';
 
 async function authenticateUser(page: Page, user: UserConfig) {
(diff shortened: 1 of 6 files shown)


computercamplove (approved):
LGTM!

computercamplove:
@raswonders you can ignore this test 
image

PR #3320, opened 2026-07-07
test: add E2E playwright tests for ColumnManagementModal

raswonders:
RHINENG-26760

## What
It adds helpers for ColumnManagementModal which makes it easier to test modal itself and other parts of InventoryViews.

It adds e2e coverage for ColumnManagementModal described by JIRA.

## Testing
Command I was using locally

`INVENTORY_VIEWS="true" npx playwright test _playwright-tests/test_column_management_modal.test.ts --workers 4 --headed`

## Summary by Sourcery

Add Playwright end-to-end coverage and reusable helpers for the Inventory Systems Column Management modal.

New Features:
- Introduce a ColumnManagementModal helper API for interacting with the manage columns dialog in Playwright tests.
- Add dedicated E2E tests validating column enabling/disabling, persistence, reset, and drag-and-drop reordering in the Column Management modal.

Enhancements:
- Refactor existing inventory views column tests to reuse the ColumnManagementModal helper instead of direct locators.
- Add reusable helpers for asserting visible inventory columns, hidden columns, and column header ordering in the systems table.

### DIFF
--- a/_playwright-tests/helpers/columnManagementModal.ts
+++ b/_playwright-tests/helpers/columnManagementModal.ts
@@ -15,6 +15,9 @@ export type ColumnManagementModal = {
   selectedCount: Locator;
   readonly columns: Promise<ColumnListEntry[]>;
   open: (timeout?: number) => Promise<void>;
+  close: () => Promise<void>;
+  save: () => Promise<void>;
+  cancel: () => Promise<void>;
   enableColumn: (columnName: string) => Promise<void>;
   disableColumn: (columnName: string) => Promise<void>;
   dragColumnTo: (sourceColumn: string, targetColumn: string) => Promise<void>;
@@ -27,8 +30,7 @@ export type ColumnManagementModal = {
  * const modal = columnManagementModal(page);
  * await modal.open();
  * await modal.enableColumn('Status');
- * await modal.saveButton.click();
- * await expect(modal.root).toBeHidden();
+ * await modal.save();
  */
 export function columnManagementModal(
   page: Page,
@@ -38,21 +40,25 @@ export function columnManagementModal(
   const columnList = page.locator(
     `[data-ouia-component-id="${ouiaId}-column-list"]`,
   );
+  const saveButton = page.locator(
+    `[data-ouia-component-id="${ouiaId}-save-button"]`,
+  );
+  const cancelButton = page.locator(
+    `[data-ouia-component-id="${ouiaId}-cancel-button"]`,
+  );
+  const resetButton = page.locator(
+    `[data-ouia-component-id="${ouiaId}-reset-button"]`,
+  );
+  const closeButton = root.getByRole('button', { name: 'Close' });
   const columnCheckbox = (columnName: string) =>
     root.getByLabel(columnName, { exact: true });
 
   return {
     root,
-    saveButton: page.locator(
-      `[data-ouia-component-id="${ouiaId}-save-button"]`,
-    ),
-    cancelButton: page.locator(
-      `[data-ouia-component-id="${ouiaId}-cancel-button"]`,
-    ),
-    resetButton: page.locator(
-      `[data-ouia-component-id="${ouiaId}-reset-button"]`,
-    ),
-    closeButton: root.getByRole('button', { name: 'Close' }),
+    saveButton,
+    cancelButton,
+    resetButton,
+    closeButton,
     columnList,
     bulkSelectCheckbox: root.locator(
       '[data-ouia-component-id="BulkSelect-checkbox"]',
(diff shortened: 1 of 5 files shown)


raswonders:
@sourcery-ai review

_playwright-tests/helpers/columnManagementModal.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Locators for the ColumnManagementModal on the Systems view.
+ *
+ * @example
+ * const modal = columnManagementModal(page);
+ * await modal.open();
+ * await modal.enableColumn('Status');
+ * await modal.saveButton.click();
+ * await expect(modal.root).toBeHidden();
+ */
+export function columnManagementModal(
```
computercamplove, line 35: really like this helper!

_playwright-tests/test_column_management_modal.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+test.describe('Column Management Modal', { tag: ['@inventory-views'] }, () => {
+  test.beforeEach(async ({ page }) => {
+    await navigateToInventorySystemsFunc(page);
+
+    const visibleHeaders = page.locator('th').filter({ hasText: /.+/ });
+    await expect(visibleHeaders).toHaveCount(totalDefaultColumns);
+  });
+
+  test('open Manage Columns modal', async ({ page }) => {
```
computercamplove: we don't need to test it - it is already part of other test cases were we open the modal to apply collumns

if you want to test just the open button - it could be unit test.

computercamplove (dismissed):

_playwright-tests/test_column_management_modal.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    await modal.open();
+
+    await expect(modal.root).toBeVisible();
+    await expect(
+      modal.root.getByRole('heading', { name: 'Manage columns' }),
+    ).toBeVisible();
+  });
+
+  test('Name column is checked and cannot be toggled off', async ({ page }) => {
```
computercamplove: this easy can be unit test - we don't need to run whole app to verify it

_playwright-tests/test_column_management_modal.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    await modal.open();
+
+    const nameColumn = modal.root.getByLabel('Name', { exact: true });
+
+    await expect(nameColumn).toBeChecked();
+    await expect(nameColumn).toBeDisabled();
+  });
+
+  test('disable a default column and save', async ({ page }) => {
```
computercamplove: this should be unit test - default columns are toggable

_playwright-tests/test_column_management_modal.test.ts
raswonders: will remove it was wondering if its indeed needed myself

_playwright-tests/test_column_management_modal.test.ts
raswonders: in my opinion this shouldn't be unit test. Because we're testing here column configuration rather than component.

_playwright-tests/test_column_management_modal.test.ts
raswonders: Not quite sure if I agree. This particular test case is checking  whether we're able to disable defaultColumn and propagate those changes to table. So it scopes over (table + modal). Unit test would scope just modal

computercamplove: it is already covered in sorting test case

_playwright-tests/test_column_management_modal.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    await modal.open();
+
+    await modal.disableColumn(columnToDisable);
+
+    await modal.save();
+
+    await expectInventoryColumnHidden(page, columnToDisable);
+  });
+
+  test('cancel button disregards column changes', async ({ page }) => {
```
computercamplove: can be unit

_playwright-tests/test_column_management_modal.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    await modal.open();
+
+    await modal.disableColumn(columnToDisable);
+
+    await modal.cancel();
+
+    await expectDefaultInventoryColumnsVisible(page);
+  });
+
+  test('close button disregards column changes', async ({ page }) => {
```
computercamplove: can be unit

_playwright-tests/test_column_management_modal.test.ts
computercamplove: in this case i would add here another verification step - that user doesn't affect it (since it is in the name of the test "cannot be toggled off"):

- unselect all checkboxes and check that Name still is selected

Image

Image

_playwright-tests/test_column_management_modal.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    await expect(modal.saveButton).toBeEnabled();
+    await modal.save();
+
+    expect(await getVisibleInventoryColumnOrder(page)).toEqual(
+      reorderedColumnNames,
+    );
+  });
+
+  test('bulk select toggles all toggleable columns', async ({ page }) => {
```
computercamplove: can be removed - we already covered it here `Sticky columns remain visible during horizontal scroll`, step `Enable all columns via Bulk Select to create horizontal scroll `

_playwright-tests/test_column_management_modal.test.ts
raswonders: I decided remove also this one. Figured we can check toggleability in unit tests and assert untoggleable Name column over column definition.

raswonders:
removed all unnecessary tests as we agreed

computercamplove (approved):

PR #3324, opened 2026-07-08
fix: Image toggle and Global tag filter reappearing on navigation

computercamplove:
## What
Fix-up Image toggle and Global tag filter reappearing on navigation

## Why

- Bug: Images view toggle not appearing reliably (known issue)

 `useFeatureFlag` returns `undefined` while Unleash is loading. The `useEffect` in `Routes.js` ran once at mount with `isBifrostEnabled === undefined`, so `inventoryHasBootcImages()` was never called and `hasBootcImages` stayed `false` - the toggle never appeared.

- Bug: Global tag filter reappearing on navigation

`InventoryViews` used `useLayoutEffect` with `[chrome]` dependency. Any chrome reference change triggered cleanup (`hideGlobalFilter(false)`) followed by re-run (`hideGlobalFilter(true)`), causing the filter to flash visible. 

## How
- Feature flag removal (`hbi.ui.bifrost`) - the feature has been fully released to all users in production for ~2 years.
- Changed to `useEffect with [chrome] ` without cleanup (`hideGlobalFilter(false)`) 

## Testing
- Global tag filter is hidden on systems page when `SystemsView`/`InventoryViews`
- `ImageToggle` is always present when account has image based systems

### DIFF
--- a/src/routes/InventoryViews.tsx
+++ b/src/routes/InventoryViews.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useLayoutEffect, useState } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Flex, FlexItem, PageSection } from '@patternfly/react-core';
 import SystemsView from '../components/SystemsView';
 import {
@@ -24,13 +24,9 @@ const InventoryViews = ({ hasAccess }: InventoryViewsProps) => {
   const { hasBootcImages } = useContext(AccountStatContext);
   const chrome = useChrome();
 
-  useLayoutEffect(() => {
+  useEffect(() => {
     chrome?.hideGlobalFilter?.(true);
-
-    return () => {
-      chrome?.hideGlobalFilter?.(false);
-    };
-  }, [chrome]);
+  }, []);
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


adarshdubey-star (approved):

src/routes/InventoryViews.tsx
```diff
@@ -24,13 +24,9 @@ const InventoryViews = ({ hasAccess }: InventoryViewsProps) => {
...
 
-  useLayoutEffect(() => {
+  useEffect(() => {
     chrome?.hideGlobalFilter?.(true);
-
-    return () => {
-      chrome?.hideGlobalFilter?.(false);
-    };
-  }, [chrome]);
+  }, []);
```
raswonders: Could you please explain why chrome dependency shouldn't be there? I didn't quite get it from description.

computercamplove: oh, right - i didn't explain it, only removal of `hideGlobalFilter(false)` 

i thought it is enough to run it once on load just hide it, similar as Workspaces page. i read that it is more safe to use [chrome], in this case should i add it to `InventoryGroups.js` too?

raswonders: I'd not mind the InventoryGroups in this PR it doesn't seem relevant

computercamplove:
@raswonders i returned back [chrome], instead of []

computercamplove:
ignore the RBAC tests failing - gonna prepare separate PR for it
image

raswonders (approved):
lgtm, thanks!

PR #3326, opened 2026-07-08
refactor(RHINENG-28478): InventoryViews update all inventory columns to use CellValue handling defaults uniformly

raswonders:
RHINENG-28478

## What
This is follow up refactor on: #3300 

It refactors all Inventory based colums, so they all behave in unified way. (namely in rendering missing and default values).
Some columns didn't have a cells yet so they were created and appropriate unit tests as well.

## Why
We need consistency between cells in their API and how they handle missing values

## How
CellValues codifies handling of missing and default values. All column cells were refactored to use it.

## Testing
Go to Inventory / Systems with preview on
Enable all Inventory based columns via ColumnManagementModal.
Observe values are appropriately rendered in similar fashion to other third party columns

image

## Summary by Sourcery

Unify SystemsView inventory columns around the shared CellValue component for consistent handling of present and missing data, while narrowing cell inputs to focused value props instead of full System objects.

Enhancements:
- Refactor DisplayName, OperatingSystem, Status, LastSeen, Workspace, Workload, and Tags cells to accept specific value-based props and render via CellValue with standardized not-available and present states.
- Introduce dedicated Created, Infrastructure, and Vendor cell components that encapsulate formatting and CellValue usage for their respective inventory fields.
- Update inventory column definitions in SystemsView to wire the new value-based cell props and use the new cell components instead of inline render logic.

Tests:
- Expand and adapt unit tests for all affected inventory cells to cover the new value-focused APIs, CellValue-based missing-data behavior, and updated formatting rules.

### DIFF
--- a/src/components/SystemsView/columns/inventory/cells/Created.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/Created.tsx
@@ -1,13 +1,14 @@
 import React from 'react';
 import { DateFormat } from '@redhat-cloud-services/frontend-components/DateFormat';
 import CellValue from '../../CellValue';
+import { System } from '../../../hooks/useSystemsQuery';
 
 interface CreatedProps {
-  value: string | null | undefined;
+  value: System['created'];
 }
 
 const Created = ({ value }: CreatedProps) => {
-  if (value === null || value === undefined) {
+  if (value === undefined) {
     return (
       <CellValue
         type="notAvailable"
--- a/src/components/SystemsView/columns/inventory/cells/Created.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/Created.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { DateFormat } from '@redhat-cloud-services/frontend-components/DateFormat';
+import CellValue from '../../CellValue';
+
+interface CreatedProps {
+  value: string | null | undefined;
+}
+
+const Created = ({ value }: CreatedProps) => {
+  if (value === null || value === undefined) {
+    return (
+      <CellValue
+        type="notAvailable"
+        reason="Created date is not available for this system"
+      />
+    );
+  }
+
+  return <CellValue type="present" value={<DateFormat date={value} />} />;
+};
+
+export default Created;
--- a/src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
@@ -1,15 +1,38 @@
 import React from 'react';
-import OperatingSystemFormatter from '../../../../../Utilities/OperatingSystemFormatter';
-import { System } from '../../../hooks/useSystemsQuery';
+import type { SystemProfileOperatingSystem } from '@redhat-cloud-services/host-inventory-client';
+import CellValue from '../../CellValue';
 
 interface OperatingSystemProps {
-  system: System;
+  value: SystemProfileOperatingSystem | undefined;
 }
 
-const OperatingSystem = ({ system }: OperatingSystemProps) => (
-  <OperatingSystemFormatter
-    operatingSystem={system.system_profile?.operating_system}
-  />
-);
+const formatOperatingSystem = (
+  operatingSystem: SystemProfileOperatingSystem,
+): string => {
+  if (
+    operatingSystem.name === 'RHEL' ||
+    operatingSystem.name === 'CentOS Linux'
+  ) {
+    const { name, major, minor } = operatingSystem;
+    const hasVersion = typeof major === 'number' && typeof minor === 'number';
+
+    return hasVersion ? `${name} ${major}.${minor}` : name;
+  }
+
+  return operatingSystem.name;
+};
+
+const OperatingSystem = ({ value }: OperatingSystemProps) => {
+  if (value == null || value.name == null) {
+    return (
+      <CellValue
+        type="notAvailable"
+        reason="Operating system data is not available for this system"
+      />
+    );
+  }
+
+  return <CellValue type="present" value={formatOperatingSystem(value)} />;
+};
 
 export default OperatingSystem;
--- a/src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
@@ -23,7 +23,7 @@ const formatOperatingSystem = (
 };
 
 const OperatingSystem = ({ value }: OperatingSystemProps) => {
-  if (value == null || value.name == null) {
+  if (value === undefined) {
     return (
       <CellValue
         type="notAvailable"
--- a/src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
@@ -32,7 +32,16 @@ const OperatingSystem = ({ value }: OperatingSystemProps) => {
     );
   }
 
-  return <CellValue type="present" value={formatOperatingSystem(value)} />;
+  return (
+    <CellValue
+      type="present"
+      value={
+        <span aria-label="Formatted OS version">
+          {formatOperatingSystem(value)}
+        </span>
+      }
+    />
+  );
 };
 
 export default OperatingSystem;
--- a/src/components/SystemsView/columns/inventory/cells/Status.tsx
+++ b/src/components/SystemsView/columns/inventory/cells/Status.tsx
@@ -4,26 +4,24 @@ import {
   ExclamationCircleIcon,
   ExclamationTriangleIcon,
 } from '@patternfly/react-icons';
-import { NOT_AVAILABLE } from '../../CellValue';
+import CellValue from '../../CellValue';
 import { System } from '../../../hooks/useSystemsQuery';
 
-export type HostStalenessStatus =
-  | 'Fresh'
-  | 'Stale'
-  | 'Stale warning'
-  | typeof NOT_AVAILABLE;
+export type HostStalenessStatus = 'Fresh' | 'Stale' | 'Stale warning';
+
+export type StatusTimestamps = Pick<
+  System,
+  'stale_timestamp' | 'stale_warning_timestamp' | 'culled_timestamp'
+>;
 
 export const getHostStalenessStatus = (
-  system: Pick<
-    System,
-    'stale_timestamp' | 'stale_warning_timestamp' | 'culled_timestamp'
-  >,
+  value: StatusTimestamps,
   now: Date = new Date(),
-): HostStalenessStatus => {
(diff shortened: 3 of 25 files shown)


raswonders:
@sourcery-ai review

src/components/SystemsView/columns/inventory/cells/Workload.tsx
```diff
@@ -1,31 +1,38 @@
...
 
-const Workload = ({ system }: WorkloadProps) => {
-  const workloads = system.system_profile?.workloads as
-    | Record<string, unknown>
-    | undefined;
-
-  if (!workloads) {
-    return NOT_AVAILABLE;
+const Workload = ({ value }: WorkloadProps) => {
+  if (value === undefined) {
```
johnsonm325, line 11: other columns check for `(value == null)` as well. Should we do that here?

src/components/SystemsView/columns/inventory/cells/Vendor.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import CellValue from '../../CellValue';
+
+interface VendorProps {
+  value: string | null | undefined;
+}
+
+const Vendor = ({ value }: VendorProps) => {
+  if (value == null || value === '') {
```
johnsonm325: AI seems to think that this would handle `null`, `undefined`, and empty string. Should we go this route? Or is there a reason why each column has different checks?

johnsonm325: Ah, I see why now. `value == null` covers both `null` and `undefined` because it's not a strict equal.

src/components/SystemsView/columns/inventory/cells/Created.tsx
```diff
@@ -0,0 +1,22 @@
...
+  if (value === null || value === undefined) {
+    return (
+      <CellValue
+        type="notAvailable"
+        reason="Created date is not available for this system"
+      />
+    );
+  }
+
+  return <CellValue type="present" value={<DateFormat date={value} />} />;
```
johnsonm325, line 20: I think we should probably add a `key` here.

src/components/SystemsView/columns/inventory/cells/Status.tsx
```diff
@@ -51,43 +49,56 @@ export const getHostStalenessStatus = (
...
+const staleStatus = (
+  <span className="pf-v6-u-display-inline-flex pf-v6-u-align-items-center">
+    <Icon status="warning" className="pf-v6-u-mr-xs">
+      <ExclamationTriangleIcon />
+    </Icon>
+    <span className="pf-v6-u-text-color-status-warning pf-v6-u-font-weight-bold">
+      Stale
+    </span>
+  </span>
+);
```
johnsonm325, line 68: AI tells me that this is a bit of an anti-pattern because:

```
JSX constants are created once at module load, not per-render. This is fine for static content but could be problematic if:
  - Icons need context (theme, etc.)
  - Future changes add dynamic props
```

It says it's better to define it in the render. But I don't think this is a big deal.

src/components/SystemsView/columns/inventory/cells/Status.tsx
```diff
@@ -51,43 +49,56 @@ export const getHostStalenessStatus = (
...
+const staleWarningStatus = (
+  <span className="pf-v6-u-display-inline-flex pf-v6-u-align-items-center">
+    <Icon status="danger" className="pf-v6-u-mr-xs">
+      <ExclamationCircleIcon />
+    </Icon>
+    <span className="pf-v6-u-text-color-status-danger pf-v6-u-font-weight-bold">
+      Stale warning
+    </span>
+  </span>
+);
```
johnsonm325, line 79: Same here.

src/components/SystemsView/columns/inventory/columnDefinitions.tsx
```diff
@@ -91,8 +105,14 @@ const statusColumn = {
...
-  renderCell: (system: System) => (
-    <Status key={`status-${system.id}`} system={system} />
+  renderCell: (system: InventoryViewSystem) => (
+    <Status
+      value={{
+        stale_timestamp: system.stale_timestamp,
+        stale_warning_timestamp: system.stale_warning_timestamp,
+        culled_timestamp: system.culled_timestamp,
+      }}
+    />
```
johnsonm325, line 115: Not sure what to make of this comment from AI:

```
  Issue: Creates a new object on every render. Could cause unnecessary re-renders if Status uses React.memo.

  Severity: Low (minor perf issue, only matters if table has many rows)

  Better: Extract timestamps outside renderCell or use useMemo
```

We're not using `useMemo`, but I could see us adding it in the future without thinking about its effects.

johnsonm325: If this is, in fact, an issue, we'd need to update a few `renderCell`s here.

src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
```diff
@@ -1,15 +1,38 @@
 import React from 'react';
-import OperatingSystemFormatter from '../../../../../Utilities/OperatingSystemFormatter';
```
johnsonm325, line 2: There's a decent amount of logic repeated in this component that exists in `OperatingSystemFormatter`. Should we create a card to update `OperatingSystemFormatter`?

johnsonm325: Also AI tells me:

```
  Issue: The old component had aria-label="Formatted OS version" for accessibility. The new one doesn't!

  This is a potential accessibility regression - screen readers won't announce the purpose of this text.
```

johnsonm325 (approved):
I'll approve this, but definitely look over my comments and implement things that you think are good.

src/components/SystemsView/columns/inventory/cells/Workspace.tsx
```diff
@@ -1,21 +1,41 @@
...
+    return (
+      <CellValue
+        type="notAvailable"
+        reason="Workspace data is not available for this system"
+      />
+    );
+  }
+
+  if (firstGroup.ungrouped) {
+    return <CellValue type="notSet" value={UNGROUPED_HOSTS_LABEL} />;
```
computercamplove: I don't think we should display this label in the Workspace column.
"Ungrouped Hosts" is a workspace name, it doen't mean _no workspace_. We could clarify what is the reason we have `UNGROUPED_HOSTS_LABEL` with lower case _`hosts`_ in the Workspace filter.

computercamplove: `type="notSet" `

Ungrouped Hosts is a name of existing workspace (current requirement - there's no system without group - no empty `groups: []` in /hosts data), it shouldn't be grey, unless you test N/A `groups: []`, or `groups` without `name` filed. 

Or is it new requirement i missed?

computercamplove:
@raswonders while testing this PR i realized that you added back Remediation plans column https://github.com/RedHatInsights/insights-inventory-frontend/pull/3300/changes#top - i missed it. Was it by mistake or it was approved? I think it was left for Q3 based on doc "Inventory Views: Application business requirements & use cases" -i didn't follow any thread on this topic, so i am not aware what was decided after.

1. Disable Remediation plans columns if left for Q3
2. Left 2 comments for Workspace column. 

Otherwise, it looks great! Thank you for refactoring this!

src/components/SystemsView/columns/inventory/cells/Workspace.tsx
computercamplove: I think they want in the future allow users to rename this workspace - so we just display real name in the column. I think the way how we recognize if this renamed workspace is `ungouped: true` is future UX work (will it be icon in the Column or else and how Filter will look like - do we need the label etc.)

In my opinion we don't need to change it.

src/components/SystemsView/columns/inventory/cells/Workload.tsx
raswonders: So that depends on the passed in value and its type. Type of workloads is `SystemProfileWorkloads | undefined`. There is no null in its type so we don't have to guard against it

src/components/SystemsView/columns/inventory/cells/Vendor.tsx
raswonders: thanks for pointing this difference, there is no need to check for null here either. I'll review rest of the checks and make sure they adhere to value type.

src/components/SystemsView/columns/inventory/cells/Workspace.tsx
raswonders: You're correct I didn't realize we could use name directly if its ungrouped. I applied your suggestion with some other refactors for Workspace cell. Thanks!

raswonders: That's a valid point, we can bring it up with UX. My decision to render it Gray comes from: Ungroupped Hosts is value used when no workspace is set for system. That makes it fall into unSet (default) value category. We render unset in gray for all the other columns. Design doc has use case for unset values and its displaying them in gray. I'd keep it for now and have it reviewed by susan once its all in and uniform with rest of the columns.

src/components/SystemsView/columns/inventory/cells/Created.tsx
raswonders: I thought the same for a while but it turns out we don't need it. Data View handles keys of cells by itself.

src/components/SystemsView/columns/inventory/cells/Status.tsx
raswonders: good you brought this up I've verified and this is non-issue here. Its static content and styling is done through patternfly classes and they will consume different tokens if for instance dark theme changes.

So not causing any issues but it makes component slightly more performant as these don't get recreated and don't have to be memoized either. 

So I'm keeping this as is.

src/components/SystemsView/columns/inventory/columnDefinitions.tsx
raswonders: I think this is non-issue at least for now. We don't use React.memo() and have quite small scale.(max 100rows). If somebody introduces React.memo he should clim b react render tree and verify something ain't passing unstable props anyway so this not breaking future proofness. Also its important to note this is not talking about useMemo.

src/components/SystemsView/columns/inventory/cells/OperatingSystem.tsx
raswonders: I decided to abandon the original formatter cause it has its its own Not availble handling which conflict with our table. I agree there is some transformation logic which could be extracted / reused and I'll do it in follow up PR. 

I'll add back the label though

computercamplove (approved):
LGTM!

PR #3328, opened 2026-07-08
fix(RHINENG-22668): Add primary button to all modals for closing

johnsonm325:
## Jira
[RHINENG-22668](https://redhat.atlassian.net/browse/RHINENG-22668)

## What
By PF guidelines, all modals must have a primary button that performs a primary action. When there is no action for a modal, it still must have a primary button to close the modal.

The modals in this PR have been given a "Close" button. Also, because the modal was exactly the same in both locations, I have combined them into 1 component and shared them. If you think this is overkill, I can easily revert.

image

[RHINENG-22668]: https://redhat.atlassian.net/browse/RHINENG-22668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce a shared SystemDetailsModal component for system detail views and ensure its modal dialogs include a primary Close action.

New Features:
- Add a reusable SystemDetailsModal component used by both the Details and Overview tabs to display system information in a modal table.

Bug Fixes:
- Ensure system detail modals include a primary Close button and close correctly when it is clicked.
- Update the entity table toolbar conditional filter label text to match the current "Filter by text" menu option.

Enhancements:
- Add tests covering SystemDetailsModal behavior and its integration with the Details and Overview screens, including modal visibility and close interactions.

Tests:
- Add comprehensive unit tests for SystemDetailsModal, including rendering, variants, real-world data scenarios, and close behavior.
- Extend Details and Overview tests to verify presence and correct behavior of the new modal and Close button.

### DIFF
--- a/src/components/SystemDetails/Details.js
+++ b/src/components/SystemDetails/Details.js
@@ -1,14 +1,8 @@
 import React, { Fragment, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import {
-  Grid,
-  GridItem,
-  Modal,
-  ModalHeader,
-  ModalBody,
-} from '@patternfly/react-core';
+import { Grid, GridItem } from '@patternfly/react-core';
 import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
-import InfoTable from '../GeneralInfo/InfoTable';
+import SystemDetailsModal from './SystemDetailsModal';
 import '../GeneralInfo/system-details.scss';
 import { OperatingSystemCard } from '../GeneralInfo/OperatingSystemCard';
 import { BiosCard } from '../GeneralInfo/BiosCard';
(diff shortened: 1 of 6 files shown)


src/components/SystemDetails/SystemDetailsModal.js
raswonders, line 1: Can we re-create this together with tests in TypeScript?

raswonders (changes requested):
Thanks for improving modals, It would be really helpful if we create new components and their tests in TS though. Considering this is new leaf component component it should be straight forward. 

Please also re-create the snapshots for updated modals, they seem to fail currently.

raswonders (approved):
thanks for TS conversion lgtm

PR #3329, opened 2026-07-09
fix(RHINENG-28481): Fix column storage when switching between stable and preview

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-28481

## What
Fix column storage when switching between stable and preview

## Why
In stable mode (inventory views disabled), the column management modal was showing and persisting columns that require the inventory views API

## How
`usePersistedColumns` now reads `useInventoryViewsFeatureFlag` internally. In stable mode (`isInventoryViewsEnabled=false)`:
  - Sets columns directly to defaults, bypassing `localStorage` entirely
  - Clears `loadedStorageKeyRef` so switching back to preview forces a fresh load and correctly restores saved preview `prefs`
  - Skips saving so preview `prefs` in `localStorage` are never overwritten

## Testing
1. navigate to systems page
2. in switch to preview and apply new columns
3. reload page and check that columns still applied 
4. switch to stable version -> only default columns are displayd
5. switch back to preview  -> previously applied new columns are back since in preview InventoryViews is enabled

## Summary by Sourcery

Ensure systems view column persistence is only applied when inventory views are enabled, resetting to defaults in stable mode.

Bug Fixes:
- Prevent stable mode from loading and persisting inventory-views-dependent column preferences so default columns are correctly shown when inventory views are disabled.

Tests:
- Update usePersistedColumns tests to mock the inventory views feature flag and validate behavior when column persistence is enabled.

### DIFF
--- a/src/components/SystemsView/hooks/useColumns.tsx
+++ b/src/components/SystemsView/hooks/useColumns.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { useCallback, useEffect, useMemo } from 'react';
-
 import type { OnSort, SortDirection } from '../SystemsView';
 import {
   getColumnMinWidthStyle,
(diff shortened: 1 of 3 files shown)


raswonders (approved):
works and code looks good thanks!

PR #3332, opened 2026-07-10
fix: ignore InventoryViews sort keys when switching to legacy table

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-28346

## What
Ignore `InventoryViews` sort keys when switching to legacy table. 

## Why
Cross-app `InventoryViews` sort keys persisted in the URL after sorting in Preview mode (`SystemsView` +` ui.inventory-views`) were being read by the legacy table on switch to Stable, causing an invalid order_by sent to the /hosts API and a broken sort indicator.

## Testing
This bug can be reproduced when Legacy table is enabled in stable and `SystemsView`+`InventoryViews` in preview:

1. modify in stage unleash SystemsView flag to be enabled only in preview
2. Visit Systems page and make sure stable version displays Legacy table and preview is InventoryViews
3. In preview apply Total CVEs column
4. Sort by Total CVEs > make sure URL is changed - cves key is present in URL
5. switch to stable version > no Something went wrong is displayed, URL doesn't contain cves key

## Summary by Sourcery

Prevent invalid cross-app sort keys from breaking systems table sorting when switching between InventoryViews and the legacy table.

Bug Fixes:
- Ensure PatternFly does not receive an out-of-bounds sort column index when no matching column is shown.
- Ignore cross-app InventoryViews sort keys in URL and systems queries so the legacy /hosts table does not send invalid order_by parameters.

### DIFF
--- a/src/components/SystemsView/hooks/useColumns.tsx
+++ b/src/components/SystemsView/hooks/useColumns.tsx
@@ -42,10 +42,14 @@ export const useColumns = ({
   const { columns, setColumns } = usePersistedColumns(defaultColumns);
 
   const fromSortByToIndex = useCallback(
-    (sortBy?: Column['sortBy']) =>
-      columns
+    (sortBy?: Column['sortBy']): number | undefined => {
+      const index = columns
         .filter((col) => col.isShown)
-        .findIndex((col) => col.sortBy === sortBy),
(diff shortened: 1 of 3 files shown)


raswonders (approved):
Thanks for fixing this!

PR #3336, opened 2026-07-13
fix(RHINENG-28664): Status column sort param crashes legacy inventory table

raswonders:
RHINENG-28664

## What
sort param of newly added Inventory column was causing crash of InventoryTable. 
This PR expands on mechanism to filter out sort params of Columns which InventoryTable cannot handle. We're now going to filter out url sort of "Status" column as well 

## Testing
1. Go to Inventory Table
2. Enable Status column
3. Set sorting according this Column
4. Now switch on rendering of legacy table by localStorage.setItem("ui.legacy-inventory-table", "true")
5. Turn preview off

InventoryTable should render with default sort which should be acc. to lastSeen column

## Summary by Sourcery

Prevent legacy inventory table and SystemsView queries from using unsupported sort keys that crash the /hosts API.

Bug Fixes:
- Ignore SystemsView-only status sort key when deriving legacy inventory sortBy/order_by parameters to avoid crashes.
- Filter out cross-app and other invalid sort keys in SystemsView queries before calling the /hosts API.

Enhancements:
- Introduce a shared helper to validate and normalize sort keys for legacy inventory usage.

Tests:
- Add unit tests covering the new sort key validation helper for legacy inventory behavior.

### DIFF
--- a/src/components/SystemsView/hooks/useSystemsQuery.ts
+++ b/src/components/SystemsView/hooks/useSystemsQuery.ts
@@ -1,5 +1,6 @@
 import { keepPreviousData, useQuery } from '@tanstack/react-query';
 import { getHostList, getHostTags } from '../../../api/hostInventoryApiTyped';
+import { getLegacyInventorySortKey } from '../../../constants';
 import { InventoryFilters } from '../filters/SystemsViewFilters';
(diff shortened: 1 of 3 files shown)


computercamplove (approved):
LGTM

PR #3338, opened 2026-07-13
test: disable e2e test in prod - workspace test case only for kessel

computercamplove:
## What
disable test case intended only for Kessel enable feature for Prod CI

## Summary by Sourcery

Tests:
- Update workspace system management e2e test to be skipped when PROD environment is enabled.


raswonders (approved):
lgtm

PR #3349, opened 2026-07-16
feat(RHINENG-28774): gate Inventory Views columns by per-service RBAC

jkrocil:
## Jira
[RHINENG-28774](https://redhat.atlassian.net/browse/RHINENG-28774)

## What
Check per-service RBAC permissions before showing app_data columns in the Inventory Views table. Unauthorized columns render a centered lock icon with a tooltip. Sorting is disabled on locked columns. The column management modal shows a lock icon next to locked column titles.

## Why
The Inventory Views table displays data from 6 downstream services without checking whether the user has permissions for each service. The backend (RHINENG-28773) filters `app_data` by permission and returns a `denied_services` list in the response. The frontend reads this to render appropriate placeholders.

## How
The `/hosts-view` response includes a `denied_services` array listing services the user lacks permission for. `useInventoryViewsQuery` extracts this field and `useColumns` uses it to annotate each column with `isPermissionLocked`. `mapSystemsToRows` intercepts locked columns to render a `noPermission` CellValue instead of calling `renderCell`. The vendored `ListManager` renders a lock icon with accessible label next to locked column titles.

No independent RBAC/Kessel calls from the frontend — the backend is the single source of truth for permission state.

## Testing
- 7 unit tests for `CellValue` (all 5 cell types including `noPermission`)
- 2 unit tests for `mapSystemsToRows` (`isPermissionLocked` branch)
- All 16 new tests passing, existing tests unaffected

See screenshots on the Jira card: [RHINENG-28774](https://redhat.atlassian.net/browse/RHINENG-28774)

[RHINENG-28774]: https://redhat.atlassian.net/browse/RHINENG-28774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHINENG-28774]: https://redhat.atlassian.net/browse/RHINENG-28774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Gate Inventory Views app_data columns based on per-service RBAC and reflect denied permissions in both the systems table and column management UI.

New Features:
- Add per-service RBAC gating for Inventory Views columns using denied_services from the hosts-view API.
- Show lock icons with permission tooltips in table cells and column management modal for columns whose backing service is denied.
- Introduce an Inventory Views RBAC feature flag that controls whether denied services are honored in the frontend.

Enhancements:
- Propagate denied_services from the backend through the Inventory Views query and SystemsView into column definitions.
- Prevent sorting on permission-locked columns and automatically fall back to a visible sortable column when the current sort becomes denied.
- Extend column management context and ListManager to support permission-locked column metadata and default configurations.

Tests:
- Add unit tests for the new noPermission CellValue variant and its accessibility label behavior.
- Extend mapSystemsToRows tests to cover permission-locked columns rendering behavior.
- Add Playwright end-to-end tests validating lock icon rendering, sorting behavior, state transitions, and feature-flag-off behavior for Inventory Views RBAC.
- Introduce helpers and mocks for inventory views RBAC in Playwright setup, including a denied_services response interceptor.


jkrocil:
@sourcery-ai review

src/components/SystemsView/columns/allColumnDefinitions.ts
```diff
@@ -1,4 +1,4 @@
-import { ColumnManagementModalColumn } from '@patternfly/react-component-groups';
+import { ColumnManagementModalColumn } from '../../ColumnManagementModal/ColumnManagementModal';
```
johnsonm325: Let's revert this back.  Instead of switching the import source to the local vendored ColumnManagementModal, could we keep the upstream PF import and add isPermissionLocked here at the Column type level? This is where all column capabilities are composed, so it feels like the natural home for it.

Add this in to the file:

```
type PermissionGatedColumn = {
  isPermissionLocked?: boolean;
};
```

and then update type `Column` to:

```
export type Column = Resolve<
  ColumnManagementModalColumn &
    RenderableColumn &
    SortableColumn &
    ConsumerAppColumn &
    LayoutColumn &
    PermissionGatedColumn
>;
```

src/components/SystemsView/hooks/useInventoryViewsQuery.ts
```diff
@@ -45,9 +45,13 @@ const fetchInventoryViews = async ({
     direction,
   });
 
-  const { results: hosts, total } = await getHostViews(params);
+  const response = await getHostViews(params);
+  const { results: hosts, total } = response;
+  const deniedServices: string[] =
```
johnsonm325: would be nice to add a todo above `const deniedServices: string[] =`

`// TODO: Remove cast when API spec includes denied_services in HostViewsResponse`

src/components/SystemsView/hooks/useColumns.tsx
```diff
@@ -31,15 +31,38 @@ interface UseColumnParams {
...
+
+  const annotatePermissions = useCallback(
+    (cols: readonly Column[]): readonly Column[] =>
+      cols.map((col) => ({
+        ...col,
+        isPermissionLocked:
+          col.appName !== 'inventory' && deniedServices.includes(col.appName),
+      })),
+    [deniedServices],
+  );
```
johnsonm325, line 64: We can update this function like so and remove our changes to `mapSystemsToRows`:

```
  const annotatePermissions = useCallback(
    (cols: readonly Column[]): readonly Column[] =>
      cols.map((col) => {
        const isLocked = col.appName !== 'inventory' && deniedServices.includes(col.appName);
        return {
          ...col,
          isPermissionLocked: isLocked,
          renderCell: isLocked
            ? () => <CellValue type="noPermission" serviceName={col.appName} />
            : col.renderCell,
        };
      }),
    [deniedServices],
  );
```

src/components/SystemsView/utils/mapSystemsToRows.tsx
```diff
@@ -36,7 +37,11 @@ export const mapSystemsToRows = ({
     const selectableColumnCells = columns
       .filter((col) => col.isShown)
       .map((col) => {
-        const cell = col.renderCell(system);
+        const cell = col.isPermissionLocked ? (
+          <CellValue type="noPermission" serviceName={col.appName} />
+        ) : (
+          col.renderCell(system)
+        );
```
johnsonm325: With the update I suggest to `annotatePermissions` in `useColumns`, I think we can remove the changes to this file altogether.

johnsonm325 (changes requested):
@jkrocil I think this looks good, overall. I think we'd do best to wait until the backend change comes through and we can pull in the api spec via host-inventory-client. That will allow us to trim out a bit of the code in this PR.

But, I think there are some minor changes you can make in the meantime as well.

Nice job.

src/Utilities/useInventoryViewsRbacFeatureFlag.ts
```diff
@@ -0,0 +1,11 @@
+import useFeatureFlag from './useFeatureFlag';
+
+const useInventoryViewsRbacFeatureFlag = () => {
+  const hasUnleashFlag = useFeatureFlag('hbi.inventory-views-rbac');
+  const hasLocalFlag =
+    localStorage.getItem('hbi.inventory-views-rbac') === 'true';
```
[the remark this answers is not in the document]
jkrocil: Irrelevant - I'm following existing patterns here, see `useInventoryViewsFeatureFlag` or `useSystemsViewFeatureFlag` or `useLegacyInventoryTableFeatureFlag`...

_playwright-tests/test_inventory_views_rbac.test.ts
```diff
@@ -0,0 +1,452 @@
...
+  test('denied columns show lock icons in table cells', async ({
+    page,
+    systems,
+  }) => {
+    await page.addInitScript(() => {
+      localStorage.setItem('hbi.inventory-views-rbac', 'true');
+      localStorage.setItem('ui.systems-view', 'true');
+      localStorage.setItem('ui.inventory-views', 'true');
+    });
+    await installDeniedServicesMock(page, ['vulnerability', 'compliance']);
```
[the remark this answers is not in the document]
jkrocil: Irrelevant - again, following existing patterns.

_playwright-tests/test_inventory_views_rbac.test.ts
```diff
@@ -0,0 +1,452 @@
...
+  test.describe.configure({ retries: 3 });
+
+  test('denied columns show lock icons in table cells', async ({
+    page,
+    systems,
+  }) => {
+    await page.addInitScript(() => {
+      localStorage.setItem('hbi.inventory-views-rbac', 'true');
+      localStorage.setItem('ui.systems-view', 'true');
+      localStorage.setItem('ui.inventory-views', 'true');
```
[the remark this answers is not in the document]
jkrocil: `enableInventoryViewsRbac` only sets the `hbi.inventory-views-rbac` flag, not the other two, i.e. this is not a valid suggestion.

src/components/SystemsView/hooks/useColumns.tsx
```diff
@@ -31,15 +32,44 @@ interface UseColumnParams {
...
   onSort,
   direction,
   isInventoryViewsEnabled,
+  deniedServices,
 }: UseColumnParams) => {
-  const { columns, setColumns } = usePersistedColumns(defaultColumns);
+  const { columns: persistedColumns, setColumns } =
+    usePersistedColumns(defaultColumns);
+
+  const annotatePermissions = useCallback(
```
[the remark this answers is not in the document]
jkrocil: This is the opposite of what Michael requested in his review so I'm not doing that. :)

src/components/SystemsView/columns/allColumnDefinitions.ts
jkrocil: Done.

src/components/SystemsView/hooks/useColumns.tsx
jkrocil: Done.

src/components/SystemsView/utils/mapSystemsToRows.tsx
jkrocil: Done, reverted.

src/Utilities/hooks/useAppServicePermissions.ts
```diff
@@ -0,0 +1,206 @@
...
+      }
+      return { permissions, isLoading: false };
+    }
+
+    const permissions = {} as AppServicePermissions;
+    for (const appName of APP_NAMES) {
+      permissions[appName] = rbacResults[appName].hasAccess;
+    }
+    return { permissions, isLoading: false };
+  }, [
```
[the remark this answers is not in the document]
jkrocil: Not relevant anymore - we switched to reading `denied_services` from the backend response.

src/components/SystemsView/hooks/useInventoryViewsQuery.ts
jkrocil: Done and I actually removed the cast entirely, instead I switched it to typed denied_services which depends on the backend PR and an update to javascript-clients being merged but no more TODO.

jkrocil:
**Depends on backend PR #4613 merging + javascript-clients regen for `denied_services` typing on `HostViewQueryOutput`.**

Michael's feedback addressed, Sourcery's latest follow up seems mostly irrelevant except for the typing suggestion,
  which is addressed in f12acdb6

  9eba3b31 feat: gate per-service RBAC behind hbi.inventory-views-rbac flag
  - Flag hook with localStorage override; when off, deniedServices always []
  - Maps backend 'patch' to frontend 'content' at ingestion

  e4c1e94a refactor: address PR review feedback
  - PermissionGatedColumn type at Column level, reverted PF import
  - CellValue rendering moved into annotatePermissions, mapSystemsToRows reverted
  - Stable empty array ref, tooltip dedup

  98a1bb34 test: Playwright E2E tests
  - Route interception mock for denied_services
  - 10 tests: lock icons, sort suppression/fallback, flag gating, permission transitions, column persistence

  f12acdb6 refactor: use typed denied_services from host-inventory-client
  - Removed raw cast, uses proper types from updated client package

_playwright-tests/helpers/inventoryViewsRbacMock.ts
raswonders, line 1: What do you think of this mock in e2e @computercamplove? It feels like we isolate UI here and don't do proper e2e.

Is it ok or should we be steering this into role based tests something like we do with RBAC.

src/components/ColumnManagementModal/ColumnManagementModal.tsx
```diff
@@ -35,6 +35,8 @@ export interface ColumnManagementModalColumn {
   isUntoggleable?: boolean;
   /** Optional app identifier displayed alongside the column title. */
   appName?: string;
+  /** When true, the column is gated by a per-service RBAC permission the user lacks. */
+  isPermissionLocked?: boolean;
```
raswonders, line 39: Hmm we introduce new changes to ColumnManagementModal so we will need to push this upstream into Patternfly. I can trye. Feature is kinda generic already so it shouldn't require much changes for upstream.

src/components/InventoryViews/hooks/useInventoryViewsQuery.ts
```diff
@@ -46,9 +53,13 @@ const fetchInventoryViews = async ({
     direction,
   });
 
-  const { results: hosts, total } = await getHostViews(params);
+  const response = await getHostViews(params);
+  const { results: hosts, total } = response;
+  const deniedServices: string[] = (response.denied_services ?? []).map(
```
raswonders, line 58: this type error should disappear once we release backend parts and publish updated js-client

src/Utilities/useInventoryViewsRbacFeatureFlag.ts
```diff
@@ -0,0 +1,11 @@
+import useFeatureFlag from './useFeatureFlag';
+
+const useInventoryViewsRbacFeatureFlag = () => {
```
raswonders, line 3: We could use a more precise name like `useInventoryViewsColumnsRbacFeatureFlag`. Reason why, rbac is used all over the app and the given name can be little confusing in what it exactly scopes in

src/components/SystemsView/columns/CellValue.test.tsx
raswonders, line 1: Thanks for adding tests here ❤️

src/components/SystemsView/columns/CellValue.tsx
```diff
@@ -35,11 +39,19 @@ type NotApplicableCellValueProps = {
   reason: string;
 };
 
+/** User lacks the RBAC permission required for this service's data. */
+type NoPermissionCellValueProps = {
+  type: 'noPermission';
+  /** Lowercase service name (e.g. "vulnerability"). */
+  serviceName: string;
```
raswonders, line 46: nit pick for consistency. We use appName instead of serviceName in col definitions we should stick to this naming.

src/components/SystemsView/columns/CellValue.tsx
raswonders, line 1: I like the extension of CellValue

raswonders (commented):
Thank you for implementing this. 

I have one major concern and thats with e2e test, which is looking more like browser scoped ui test. I'd like to hear opinion of others whether we're ok with it or should we create a seperate card for e2e test. Once this is cleared I'll approve.

I've also left some notes pointing out small things.

_playwright-tests/helpers/inventoryViewsRbacMock.ts
computercamplove: @jkrocil For RBAC E2E tests we use 3 non-admin accounts - they are precreated on `stage` with pre-setup permissions - no need to do any setup or mock permissions, just pick the right `storageState` file and you're good:

1. Drop the mock - we test against real `stage`, no mocks. Viewer account already has only inventory access (no `app_data` access), so just use `test.use({ storageState: '.auth/viewer_user.json' });` and it'll work as-is (see `test_viewer_role_access.test.ts` for the pattern - same structure, just without the mock imports)

2. Place the file under `_playwright-tests/rbac/ `- same as `test_viewer_role_access.test.ts`, but use `{ tag: ['@inventory-views'] }` instead of `{ tag: ['@rbac'] }` -  so your tests run in **Inventory Views CI** (the one with Inventory Views enabled)

3. You don't need to enable `hbi.inventory-views-rbac` in E2E since we run tests against stage env and this flag will be always enabled there, if you want you can keep it.

this recommendations are for E2E tests

src/components/SystemsView/columns/CellValue.tsx
jkrocil: Done.

src/Utilities/useInventoryViewsRbacFeatureFlag.ts
jkrocil: Done.

_playwright-tests/helpers/inventoryViewsRbacMock.ts
jkrocil: Done. Thank you.

src/components/InventoryViews/hooks/useInventoryViewsQuery.ts
jkrocil: Done - updated.

src/components/ListManager/ListManager.tsx
```diff
@@ -166,23 +174,44 @@ const ListManager: FunctionComponent<ListManagerProps> = ({
   const renderColumnLabel = (
```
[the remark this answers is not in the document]
jkrocil: I don't believe this is worth extracting as suggested.

raswonders (approved):
Thanks for updating the PR. I've re-tested with user with granular permissions and feature seems to be working in UI as prescribed.

lgtm

johnsonm325 (approved):

PR #3359, opened 2026-07-19
refactor:  allow custom fetching for SystemsView 

raswonders:
## Jira
RHINENG-28482

## What

This PR is first step in making SystemsView more configuragle. It allows consumers to inject their own useQuery and use whatever backend they please with it as long as passed in query conforms to API contract defined in SystemsView.  

Above change helped us declutter SystemsView and allowed us to create two new separate components on top of it `InventoryViews` the new upcoming table, and `InventoryHosts` which provides similar UX as did vanilla SystemsView table before. 

Having separate components for each backend endpoint, also made the routing in APP more straight forward.

From JIRA scope intentionally left out redux integration, columns, filters and actions, these will be tackled separately via their own cards. 

## Testing

This PR touched alot of things from routing to fetching of both InventoryViews table and SystemsViews. 

Please during review do some UI testing as well namely:
- Make sure feature flags fro conditional rendering table work as before.
- Sorting, filtering and actions work in both `InventoryViews` and `InventoryHosts` tables.

## Summary by Sourcery

Extract data fetching and cache invalidation responsibilities out of SystemsView into shared hooks and wrapper components so the view can be reused with different inventory data sources.

New Features:
- Introduce generic SystemsView props and types to allow pluggable data query functions and external cache invalidation handling.
- Add InventoryViews and InventoryHosts wrapper components that wire SystemsView to host-views and host-list queries with appropriate react-query invalidation.
- Add a deriveActiveState helper to centralize computation of the SystemsView active state from query status.

Enhancements:
- Unify host list and host-views query parameter types and utilities under InventoryViews, reusing them across systems-related components.
- Move access control handling from SystemsView into the Systems page route component.
- Simplify mutation hooks to delegate cache invalidation through an injected callback instead of directly depending on react-query clients.

Tests:
- Update affected SystemsView tests and utilities to align with the new query hooks locations, types, and route naming.


computercamplove (approved):

src/components/InventoryViews/InventoryHosts.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const InventoryHosts = () => {
+  const queryClient = useQueryClient();
+
+  return (
+    <SystemsView
+      useDataQuery={useHostsQuery}
+      onInvalidate={() =>
+        queryClient.invalidateQueries({ queryKey: [HOSTS_QUERY_KEY] })
+      }
```
johnsonm325, line 14: The `onInvalidate` callbacks are created inline on every render. Since `SystemsView` passes `onInvalidate` down to `SystemActionModalsProvider` as a prop, this means the provider re-renders on every parent render. Not a real perf issue in practice (these components don't render frequently), but wrapping in `useCallback` would be trivially cleaner:

```
  const onInvalidate = useCallback(
    () => queryClient.invalidateQueries({ queryKey: [HOSTS_QUERY_KEY] }),
    [queryClient]
  );
```

johnsonm325 (approved):
Looks good. Just one trivial suggestion, but not a big deal as these pages don't re-render very often.

src/components/InventoryViews/InventoryHosts.tsx
raswonders: @johnsonm325 this is great point and worthy of discussion because these patterns will come up again and again. 

I double checked and only price we are paying with this inline declaration is re-creation small function on each render of InventoryViews -> this is tiny price not impacting anything.

The function itself isn't stable but it doesn't have to be because its not used as dep in any useEffect or prop in React.memo component.

SystemActionModalsProvider doesn't re-render because it props changed. 
It rerenders: 
- on rerender of parent
- on change of its own state (props arent't state)
- on change of context its using

So for this reason I'd just keep it as is cause useCallback would behave exact same way and made just code noisy.

If we will need stability of that function later on I'd add it then bacause only then there is real reason to have it. To add these things speculatively feels wrong.

PR #3365, opened 2026-07-20
feat(RHINENG-28360): Add Save As modal with validation and Views toolbar

johnsonm325:
## Jira
[RHINENG-28360](https://redhat.atlassian.net/browse/RHINENG-28360)

## What
- Add ViewSaveAsModal component for creating new inventory views
- Add duplicate view name validation with useViewNameValidation hook
- Add dummy listViewsApi and useViewsQuery hook for fetching views list
- Add toast notifications for create success/error via useAddNotification
- Add ViewsToolbar component with ManageViewButton dropdown
- Integrate toolbar into InventoryViews page below PageHeaderTitle
- Wire up modal state management and success handlers
- Add SCSS module type declaration (scss.d.ts)
- Feature flag gated behind ui.inventory-views.views-toolbar

This PR adds the UI layer for saving custom views.

[RHINENG-28360]: https://redhat.atlassian.net/browse/RHINENG-28360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce a feature-flagged UI for managing and saving custom inventory views, including a Save As modal, views toolbar, and supporting hooks and APIs for view creation and name validation.

New Features:
- Add a Save As modal for creating new inventory views with duplicate name validation and current table configuration capture
- Introduce a Views toolbar with a Manage view dropdown entry wired to the Save As flow
- Provide a dummy views listing API and a React Query hook for fetching available views

Enhancements:
- Show success and error toast notifications when creating inventory views via the create view mutation
- Gate the new inventory views UI behind a feature flag and integrate the toolbar and modal into the InventoryViews page

Tests:
- Add unit tests for the Save As modal behavior, including validation, callbacks, and form reset
- Add unit tests for the view name validation hook covering duplicates, case/whitespace handling, and exclusion options

Chores:
- Add a TypeScript declaration for SCSS modules to support new toolbar styling

### DIFF
--- a/src/components/InventoryViews/ViewsToolbar/ManageViewButton.tsx
+++ b/src/components/InventoryViews/ViewsToolbar/ManageViewButton.tsx
@@ -1,0 +1,59 @@
+import React, { useState } from 'react';
+import {
+  Dropdown,
+  DropdownItem,
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
+} from '@patternfly/react-core';
+
+export interface ManageViewButtonProps {
+  /** Current view ID (null for system views) */
+  currentViewId?: string | null;
+  /** Whether current view is a system view (non-editable) */
+  isSystemView?: boolean;
+  /** Callback when Save As is clicked */
+  onSaveAs?: () => void;
+}
+
+export const ManageViewButton = ({
+  currentViewId,
+  isSystemView = true,
+  onSaveAs,
+}: ManageViewButtonProps) => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  const onToggle = () => {
+    setIsOpen((prev) => !prev);
+  };
+
+  const onSelect = () => {
+    setIsOpen(false);
+  };
+
+  return (
+    <Dropdown
+      isOpen={isOpen}
+      onSelect={onSelect}
+      onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
+      toggle={(toggleRef: React.Ref<MenuToggleElement>) => (
+        <MenuToggle
+          ref={toggleRef}
+          onClick={onToggle}
+          isExpanded={isOpen}
+          aria-label="Manage view actions"
+        >
+          Manage view
+        </MenuToggle>
+      )}
+    >
+      <DropdownList>
+        <DropdownItem key="save-as" onClick={onSaveAs}>
+          Save as
+        </DropdownItem>
+      </DropdownList>
+    </Dropdown>
+  );
+};
+
+export default ManageViewButton;
--- a/src/components/InventoryViews/ViewsToolbar/ViewsToolbar.tsx
+++ b/src/components/InventoryViews/ViewsToolbar/ViewsToolbar.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import { Flex, FlexItem } from '@patternfly/react-core';
+import type { ViewOut } from '../../../api/inventoryViewsApi';
+import { ManageViewButton } from './ManageViewButton';
+import './ViewsToolbar.scss';
+
+export interface ViewsToolbarProps {
+  className?: string;
+  currentViewId?: string | null;
+  isSystemView?: boolean;
+  viewsList?: ViewOut[];
+  onSaveAs?: () => void;
+}
+
+export const ViewsToolbar = ({
+  className,
+  currentViewId,
+  isSystemView = true,
+  viewsList,
+  onSaveAs,
+}: ViewsToolbarProps) => {
+  return (
+    <Flex
+      className={`ins-c-views-toolbar ${className || ''}`}
+      spaceItems={{ default: 'spaceItemsMd' }}
+      alignItems={{ default: 'alignItemsCenter' }}
+    >
+      <FlexItem>
+        <div>View: </div>
+      </FlexItem>
+      <FlexItem>
+        <ManageViewButton
+          currentViewId={currentViewId}
+          isSystemView={isSystemView}
+          onSaveAs={onSaveAs}
+        />
+      </FlexItem>
+    </Flex>
+  );
+};
+
+export default ViewsToolbar;
--- a/src/components/InventoryViews/hooks/useViewNameValidation.ts
+++ b/src/components/InventoryViews/hooks/useViewNameValidation.ts
@@ -1,0 +1,20 @@
+import type { ViewOut } from '../../../api/inventoryViewsApi';
+
+export const validateViewName = (
+  viewsList: ViewOut[],
+  name: string,
+  options?: { excludeViewId?: string },
+) => {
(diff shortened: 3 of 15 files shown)


src/components/InventoryViews/ViewsToolbar/ManageViewButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+  isSystemView?: boolean;
+  /** Callback when Save As is clicked */
+  onSaveAs?: () => void;
+}
+
+export const ManageViewButton = ({
+  currentViewId,
+  isSystemView = true,
+  onSaveAs,
+}: ManageViewButtonProps) => {
```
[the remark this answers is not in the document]
johnsonm325: We'll wire them up later. We're not hitting any api right now, so I'm leaving as-is.

src/components/InventoryViews/hooks/useViewNameValidation.ts
raswonders, line 1: This views state seems to be only used in what i understand to be a helper hook. This layering is odd because we encapsulate views ... which is the master state here into something secondary. I feel like this shouldn't be hook but a function. and Views state should be consumed on InventoryViews level

src/components/InventoryViews/Modals/SaveAsModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  isOpen: boolean;
+  /** Callback to close the modal */
+  onClose: () => void;
+  /** Current table configuration to save */
+  currentConfiguration: ViewConfiguration;
+  /** Optional callback when view is successfully created */
+  onSuccess?: (viewId: string, viewName: string) => void;
+}
+
+const SaveAsModal = ({
```
raswonders: SaveAsModal could be anything, I'd propose let's name this ViewSaveModal or something which signifies what's the modal about

src/components/InventoryViews/ViewsToolbar/ViewsToolbar.tsx
raswonders, line 1: Could we build this with Patternfly primitives in similar fashion like DataViewToolbar is build. This will help us to stay consistent with look and feel of our InventoryViews table. 

https://github.com/patternfly/react-data-view/blob/93fc8b0da910fe64c653e99e46446a54155c19e3/packages/module/src/DataViewToolbar/DataViewToolbar.tsx#L6

src/routes/SystemsPage.tsx
raswonders, line 1: SystemsPage is now shared for both InventoryHosts the legacy-compatible table and InventoryViews table. I feel InventoryViews component should hold all Views related stuff if allowed by layout.

src/routes/SystemsPage.tsx
```diff
@@ -81,6 +111,14 @@ const InventoryViewsPage = ({ hasAccess = true }: SystemsPageProps) => {
           <ImagesView />
         )}
       </PageSection>
+      {isInventoryViewsEnabled && (
```
raswonders: we're using the same FeatureFlag which we currently have on review in production. 
We should create some for inventory-views in-progress features in particular cause we probably don't want to publish half baked features.

src/components/InventoryViews/hooks/useViewNameValidation.ts
johnsonm325: I'm not sure what this sentence means, "This layering is odd because we encapsulate views ... which is the master state here into something secondary."

My intention here is to use this in both "Save as" and "Rename" modals.

And this sentence, "and Views state should be consumed on InventoryViews level" I'm also not sure what this means. What do you mean by Views state? Do you mean the `viewsList` fetched here?

src/components/InventoryViews/ViewsToolbar/ViewsToolbar.tsx
johnsonm325: Good call. Thoughts on creating a Jira card related to this story to fully implement? https://redhat.atlassian.net/browse/RHINENG-28291

src/routes/SystemsPage.tsx
johnsonm325: This has been implemented now

src/routes/SystemsPage.tsx
johnsonm325: I've moved that logic into InventoryViews! Nice catch.

src/components/InventoryViews/hooks/useViewNameValidation.ts
raswonders: My comment is all about hierarchy of components and state which they should own (via useState).

When we load up InventoryViews component It feels like good idea to have views listing data available already. Why? Because it seems like it's centric state to whole feature and will be used in few places. (Toolbar/Modals) So it makes sense to me that is available at their common parent which should be InventoryViews. What I see now is that we are keeping it in some Modal via nested hook. 
This is odd because whenever we decide to reuse it, we'll be creating its own separate instance instead of common shared one.

`My intention here is to use this in both "Save as" and "Rename" modals.`

I'd say this is fine, if hierarchy of components is correct, inventoryViews owns both modals you can pass it into both of them via props, when InventoryViews owns it.

I tried my best but do let me know if it's still unclear

src/components/InventoryViews/ViewsToolbar/ViewsToolbar.tsx
raswonders: agree, this might be handled more gracefully in separate task do you want me to create it?

raswonders:
image

Also some UI things are still pending

- asterisk should be probably removed I think this is short workflow with single field everyone is clear name is required you cannot even send the modal without it
- View name should be Name acc. to mocks
- The description text seems to be styled differently than in our other modals it seems smaller. I'd make it same as our other modals in SystemsView actions

raswonders (approved):
lgtm, left few notes which deserve tackling but otherwise gj

src/components/InventoryViews/ViewsToolbar/ViewsToolbar.tsx
johnsonm325: Nope, I'll create the card, just didn't want to do it until I verified with you first. :)

src/components/InventoryViews/hooks/useViewNameValidation.ts
johnsonm325: AH! you specifically mean `viewsList` that is fetched via `useViewsQuery()`. I wasn't connecting that "Views state" was referring to `viewsList`, but now I'm following.

Yes, that makes complete sense and I agree 100%. I thought you were talking about `useViewNameValidation` as an entire hook.

johnsonm325:
Susan approved this updated view, but I changed a couple of the things you suggested. I kept "View name" though.

PR #3405, opened 2026-07-30
fix(deps): Upgrade unleash react SDK to v6

kruai:
## Jira
None

## What
Upgrade unleash react SDK to v6

## Why
Segments are not supported in earlier versions

## How
Update `package.json` and run `npm install`

## Testing
N/A

N/A

## Summary by Sourcery

Build:
- Bump @unleash/proxy-client-react from v5 to v6 in package.json and lockfile.


_playwright-tests/helpers/filterHelpers.ts
```diff
@@ -151,7 +151,7 @@ export const expectAllRowsHaveText = async (
  */
 export const searchByName = async (page: Page, name: string): Promise<void> => {
   const searchInput = page.locator('input[placeholder="Filter by name"]');
-  await page.reload({ waitUntil: 'networkidle' });
```
computercamplove, line 154: @kruai this change make Legacy table to fail

maybe we can add here condition if isLegacyInventoryTableEnabled=true then 'networkidle'  else 'load'

kruai: Updated it - let's see

computercamplove (approved):

PR #3406, opened 2026-07-31
test(RHINENG-25387): add unit tests for SystemsView 

raswonders:
RHINENG-25387

## What
Some unit tests covering SystemsView newly add colums property and table states, like error and loading.

SystemsView is complex component and most of its internal functionality is or will be covered by unit tests of its parts. This is why only handful of tests are in here.

## Summary by Sourcery

Add unit coverage for SystemsView column selection and table loading/error states.

Tests:
- Introduce SystemsView.test.tsx to verify column rendering based on the columns selector.
- Add tests ensuring SystemsView correctly displays loading states for isLoading and isFetching.
- Add a test validating SystemsView error messaging and absence of data when isError is true.

### DIFF
--- a/src/components/SystemsView/SystemsView.test.tsx
+++ b/src/components/SystemsView/SystemsView.test.tsx
@@ -1,0 +1,150 @@
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import { expect, jest } from '@jest/globals';
+import React from 'react';
+import { SystemsView, type UseSystemsViewDataQuery } from './SystemsView';
+import type { OnInvalidate } from './SystemActionModalsContext';
+import type { ColumnSelector } from './columns/resolveColumnSelector';
+import type { System } from '../InventoryViews/hooks/useHostsQuery';
+import { TestWrapper } from '../../Utilities/TestingUtilities';
+
+const mockSystem = {
+  id: 'host-1',
+  display_name: 'Test Host',
+} as System;
+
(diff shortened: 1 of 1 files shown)


computercamplove:
LGTM!

computercamplove (approved):

src/components/SystemsView/SystemsView.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }));
+  });
+
+  it('renders a column when the columns selector includes it', async () => {
+    renderSystemsView();
+
+    expect(
+      await screen.findByRole('columnheader', { name: 'Name' }),
+    ).toBeInTheDocument();
+  });
```
johnsonm325, line 93: Doesn't verify data renders in cells. Maybe asserting 'Test Host' appears as a cell? Not a blocker, just a consideration if you care.

johnsonm325 (approved):
These tests also don't cover the empty state `NoEntitiesFound`. Neither of these are blockers, but just wanted to throw them out there if you wanted to beef these tests up at all.

src/components/SystemsView/SystemsView.test.tsx
raswonders: So the intent here is to cover the columns narrowly basically verify if column turns up down the line. For that header check is sufficient. 

There are further checks for cell values per Cell component. And also checks for row generation in mapSystemsToRows.test.tsx

PR #3409, opened 2026-07-31
test(RHINENG-25383): add unit tests for action modals context 

raswonders:
RHINENG-25383

## What
Add unit tests for modal control functions. 
Small refactor of Context component. I've cleaned some duplicit kessel conditional logic. The kessel routing is done already in consumers and was not needed in Context.

## Testing
To make sure refactored code works, try to use AddToWorkspace & MoveToWorkspace row actions. They both should open appropriate modals.

## Summary by Sourcery

Add dedicated modal openers for moving systems to workspaces and update systems view actions to use them, while introducing comprehensive unit tests for the SystemActionModals context and its consumers.

New Features:
- Expose a separate openMoveSystemsToWorkspaceModal action in the SystemActionModals context for workspace move workflows.

Bug Fixes:
- Align systems view row and bulk actions to use the correct move-to-workspace modal, ensuring consistent behavior for move actions.

Enhancements:
- Simplify SystemActionModalsContext by removing kessel migration feature flag routing for workspace-related modals.

Tests:
- Add a new SystemActionModalsContext test suite covering all modal open/close behaviors, including workspace and tags modals.
- Extend existing SystemsView row, bulk actions, and tags-related tests to support the new move-to-workspace modal action and updated context API.

### DIFF
--- a/src/components/SystemsView/SystemActionModalsContext.test.tsx
+++ b/src/components/SystemsView/SystemActionModalsContext.test.tsx
@@ -23,6 +23,13 @@ jest.mock('./hooks/usePatchSystemsMutation', () => ({
   })),
 }));
 
+jest.mock('@redhat-cloud-services/frontend-components/useChrome', () => ({
+  __esModule: true,
+  default: () => ({
+    getUserPermissions: jest.fn(),
+  }),
+}));
+
 const testSystem = {
   id: 'host-1',
   display_name: 'Test Host',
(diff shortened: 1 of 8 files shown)


src/components/SystemsView/SystemActionModalsContext.test.tsx
```diff
@@ -0,0 +1,362 @@
...
+    await userEvent.click(
+      screen.getByRole('button', { name: /open delete modal/i }),
+    );
+
+    expect(
+      screen.getByText(/Delete system from inventory\?/i),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        /Test Host will be removed from all localhost:5000 applications and services/i,
```
computercamplove: if you try to delete more than 1 system the content of modal will change from system name "Test Hosts" to  "2 systems"

computercamplove: if this test case for deleting 1 system you would need to update the test title

src/components/SystemsView/SystemActionModalsContext.test.tsx
```diff
@@ -0,0 +1,362 @@
...
+    renderWithProvider(<OpenTagsModalButton systems={[]} />, {
+      withFilters: true,
+    });
+
+    expect(
+      screen.queryByRole('heading', { name: /all tags in inventory/i }),
+    ).not.toBeInTheDocument();
+  });
+
+  it('opens the all-tags modal when openTagsModal is called with an empty selection', async () => {
```
computercamplove, line 344: unit tests for hosts with 0 tags, do we need to test with at least 1 tag?

src/components/SystemsView/SystemActionModalsContext.test.tsx
raswonders: I don't think we need this exact phrase anyway. We just want to check if correct modal opens. Check for /Delete system from inventory\?/i) should be sufficient

src/components/SystemsView/SystemActionModalsContext.test.tsx
raswonders: I don't think so. This isn't suppose to be exhaustive tests for TagsModal API. We just making sure that correct modal opens. Tests you speaking of should be scoped to modal itself.

computercamplove (approved):

johnsonm325 (approved):

PR #3417, opened 2026-08-03
test(RHINENG-25386): expands unit tests for Systemsview actions

raswonders:
RHINENG-25386
RHINENG-25384

## What
Expands & refactors tests for Row Actions. I've created sets for both kessel enabled and disabled. Also incorporated RBAC states.

## Summary by Sourcery

Expand and refactor SystemsView action tests to cover both row-level and bulk actions under Kessel and non-Kessel scenarios with RBAC-aware behavior.

Tests:
- Add comprehensive tests for SystemsView row actions covering Kessel migration on/off, workspace membership, and RBAC-driven enable/disable states.
- Introduce a new test suite for SystemsView bulk actions (move, delete, add/remove workspace) including selection and table-state preconditions.
- Extract shared SystemsView test helpers and mocks into reusable fixtures for systems, permissions, feature flags, and modal contexts.

### DIFF
--- a/src/components/SystemsView/SystemsViewBulkActions.test.tsx
+++ b/src/components/SystemsView/SystemsViewBulkActions.test.tsx
@@ -1,0 +1,359 @@
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { expect, jest } from '@jest/globals';
+import { SystemActionModalsContext } from './SystemActionModalsContext';
+import {
+  GENERAL_GROUPS_WRITE_PERMISSION,
+  GENERAL_HOSTS_WRITE_PERMISSIONS,
+} from '../../constants';
+import type { System } from '../InventoryViews/hooks/useHostsQuery';
+
+const mockOpenDeleteModal = jest.fn();
+const mockOpenAddToWorkspaceModal = jest.fn();
+const mockOpenMoveSystemsToWorkspaceModal = jest.fn();
+const mockOpenRemoveFromWorkspaceModal = jest.fn();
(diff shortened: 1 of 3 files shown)


johnsonm325 (approved):
A few minor things I and AI noticed during the review:
 
- These tests in `SystemsViewRowActions` are missing Kessel test cases. The Kessel-enabled block doesn't test Edit or Delete under Kessel.
- In bulk selecting, the Kessel path tests both `activeState: 'loading'` and `activeState: 'empty'` disabling Delete, but the non-Kessel path only tests RBAC and empty selection. If the `activeState` guard applies in non-Kessel too, it's untested there.
- When bulk-selecting, some hosts could be in a workspace and some not. The test only checks single-system selections for the workspace-state guards. A test with `[createSystem(), createSystem({ groups: [testWorkspaceGroup] })]` would verify the "any" logic (e.g., "disabled when any selected host is already in a workspace").

raswonders:
> A few minor things I and AI noticed during the review:
> 
> * These tests in `SystemsViewRowActions` are missing Kessel test cases. The Kessel-enabled block doesn't test Edit or Delete under Kessel.
> * In bulk selecting, the Kessel path tests both `activeState: 'loading'` and `activeState: 'empty'` disabling Delete, but the non-Kessel path only tests RBAC and empty selection. If the `activeState` guard applies in non-Kessel too, it's untested there.
> * When bulk-selecting, some hosts could be in a workspace and some not. The test only checks single-system selections for the workspace-state guards. A test with `[createSystem(), createSystem({ groups: [testWorkspaceGroup] })]` would verify the "any" logic (e.g., "disabled when any selected host is already in a workspace").

I've applied changes for all suggestions.

PR #3426, opened 2026-08-04
feat(RHINENG-28357): View selector dropdown

johnsonm325:
## Jira
[RHINENG-28357](https://redhat.atlassian.net/browse/RHINENG-28357)

## What
This PR creates the "View" dropdown that shows the current view name and allows users to select from available views (user-created + system views).

- Dropdown shows current view name (e.g., "Production view" or "All systems")
- Lists available views in order: "All systems" → divider → user views → divider → system views
- Active view indicated with checkmark
- System views visually distinguished with badge/icon
- Clicking view switches to it (triggers active view change)

Note: the current implemention doesn't allow us to fetch views, so this is a mocked list of views. The modal actions can be performed, but they don't affect any of mocks. Also, the views don't have associated filters or columns in this implementation. This simply creates the selector and allows switching between views.

## Screenshots
image

[RHINENG-28357]: https://redhat.atlassian.net/browse/RHINENG-28357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce a selectable views dropdown to the inventory toolbar and hook it into inventory view state, with tests validating its behavior.

New Features:
- Add a view selector dropdown to the inventory views toolbar that displays the current view and allows switching between available views

Enhancements:
- Wire inventory views state so the active view ID is tracked in component state instead of a static constant and passed through the toolbar

Tests:
- Add unit tests for the view selector dropdown covering rendering, interaction, keyboard support, and system view labeling


src/components/InventoryViews/ViewsToolbar/ViewSelector.tsx
```diff
@@ -0,0 +1,118 @@
...
+        )}
+        {userViews.length > 0 && (
+          <>
+            <Divider />
+            {userViews.map((view) => (
+              <SelectOption key={view.id} value={view.id}>
+                {view.name}
+              </SelectOption>
+            ))}
+          </>
```
raswonders, line 93: Patternfly has also SelectGroup which seems to natively divide content could we use it here? 

https://www.patternfly.org/components/menus/select/

Image

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  it('should show dropdown options when clicked', async () => {
+    const user = userEvent.setup();
+    renderViewSelector();
+
+    await user.click(screen.getByRole('button', { name: 'Select a view' }));
+
+    expect(screen.getByText('Production view')).toBeInTheDocument();
+    expect(screen.getByText('RHEL 9 staging')).toBeInTheDocument();
+    expect(screen.getByText('Security overview')).toBeInTheDocument();
```
raswonders: This one seems fragile. Could we check for something else here than hardcoded dropdown values to see if opens?

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const toggle = screen.getByRole('button', { name: 'Select a view' });
+    await user.click(toggle);
+    expect(toggle).toHaveAttribute('aria-expanded', 'true');
+
+    await user.click(screen.getByText('Production view'));
+
+    expect(toggle).toHaveAttribute('aria-expanded', 'false');
+  });
+
+  it('should show the selected view name in the toggle', () => {
```
raswonders, line 62: I don't quite get how this one differs from first one. The first one might be extra?

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    renderViewSelector({ activeViewId: 'view-production' });
+
+    expect(screen.getByText('Production view')).toBeInTheDocument();
+  });
+
+  it('should render dividers between sections', async () => {
```
raswonders, line 68: When we use SelectGroups we can remove this one

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const user = userEvent.setup();
+    renderViewSelector();
+
+    await user.click(screen.getByRole('button', { name: 'Select a view' }));
+
+    const dividers = screen.getAllByRole('separator');
+    expect(dividers.length).toBeGreaterThanOrEqual(2);
+  });
+
+  it('should close the dropdown when pressing Escape', async () => {
```
raswonders: Is this custom behavior we testing? This one feels like we testing Patternfly's Select behavior, in which case I'd delete it

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const toggle = screen.getByRole('button', { name: 'Select a view' });
+    await user.click(toggle);
+    expect(toggle).toHaveAttribute('aria-expanded', 'true');
+
+    await user.keyboard('{Escape}');
+
+    expect(toggle).toHaveAttribute('aria-expanded', 'false');
+  });
+
+  it('should select a view with keyboard Enter', async () => {
```
raswonders: same as Esc one

raswonders (commented):
Thanks for the feature. I tested it and it seems to work. I left some notes regarding unit tests and markup.

raswonders (approved):

src/components/InventoryViews/ViewsToolbar/ViewSelector.tsx
johnsonm325: It doesn't natively have it. Notice the `<Divider />` component in between the `SelectGroup`s. :)

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
johnsonm325: This one checks that when an active view is selected that it shows up in a collapsed menu:

image

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
johnsonm325: We're setting the views from the `MOCK_VIEWS` and then checking there, so I don't think it's fragile. I'm going to keep the check but I'm going to move `MOCK_VIEWS` to a testing file so it will stick around. And, I'll loop through the `MOCK_VIEWS` instead of calling out each individual line item.

src/components/InventoryViews/ViewsToolbar/ViewSelector.test.tsx
johnsonm325: keeping this test based on my comment on the Divider.

PR #3434, opened 2026-08-06
feat(RHINENG-28461): Generate V2 TypeScript Client for Views API

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-28461

## What
Generate TypeScript client for V2 Inventory Views API

## Why
Use real API

## How
Use the existing @redhat-cloud-services/host-inventory-client package instead of dummy/mocks

## Testing
- visit Systems page stage
- try to create/rename/delete view
image
image

## Summary by Sourcery

Replace the mock inventory views API with a real TypeScript client generated from the host-inventory V2 Views API and wire it into the existing views operations.

New Features:
- Introduce a typed client wrapper around the V2 Inventory Views API using the shared JavaScript client factory.

Enhancements:
- Switch list, get, create, update, delete, and clone inventory view operations from local dummy implementations to calls against the real backend Views API.
- Export additional typed helpers for view-related requests and responses to align with the generated client types.

Tests:
- Update modal unit tests to mock the new API client functions for creating, renaming, and deleting views.

### DIFF
--- a/src/api/inventoryViewsApi.ts
+++ b/src/api/inventoryViewsApi.ts
@@ -1,7 +1,18 @@
-// ============================================================================
-// TYPE IMPORTS (from @redhat-cloud-services/host-inventory-client)
-// ============================================================================
-
+import { APIFactory } from '@redhat-cloud-services/javascript-clients-shared';
+import axiosInstance from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+import * as clientExports from '@redhat-cloud-services/host-inventory-client';
+import type { AxiosInstance } from 'axios';
+import type {
+  ApiViewsGetViewsListParams,
+  ApiViewsGetViewsListReturnType,
+} from '@redhat-cloud-services/host-inventory-client/ApiViewsGetViewsList';
+import type {
+  ApiViewsGetViewByIdParams,
+  ApiViewsGetViewByIdReturnType,
+} from '@redhat-cloud-services/host-inventory-client/ApiViewsGetViewById';
+import type { ApiViewsCreateViewReturnType } from '@redhat-cloud-services/host-inventory-client/ApiViewsCreateView';
+import type { ApiViewsUpdateViewReturnType } from '@redhat-cloud-services/host-inventory-client/ApiViewsUpdateView';
+import type { ApiViewsCloneViewReturnType } from '@redhat-cloud-services/host-inventory-client/ApiViewsCloneView';
 import type {
   ViewConfiguration,
   ViewIn,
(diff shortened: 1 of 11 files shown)


src/api/inventoryViewsApi.ts
```diff
@@ -10,187 +21,83 @@ import type {
...
-    count: MOCK_VIEWS.length,
-    page: 1,
-    per_page: 50,
-    total: MOCK_VIEWS.length,
-    results: MOCK_VIEWS,
+export const listViewsApi =
+  async (): Promise<ApiViewsGetViewsListReturnType> => {
+    return (await inventoryApi().apiViewsGetViewsList(
+      {},
+    )) as unknown as ApiViewsGetViewsListReturnType;
```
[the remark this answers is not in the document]
computercamplove: The as unknown as is the established pattern across the entire project (hostInventoryApiTyped.ts does the same on every method). Changing it here without fixing the upstream APIFactory generics would either break compilation or create inconsistency with the rest of the codebase.

src/api/inventoryViewsApi.ts
```diff
@@ -10,187 +28,70 @@ import type {
...
-    org_id: 'org-123',
-    org_wide: true,
-    configuration: { columns: [] },
-    created_by: 'system',
-    created_at: '2026-01-01T00:00:00Z',
-    updated_at: '2026-01-01T00:00:00Z',
-    is_owner: false,
-  },
-];
+const INVENTORY_API_BASE = '/api/inventory/v1';
```
johnsonm325: `INVENTORY_API_BASE` exists in `src/config.js` as well.

src/api/inventoryViewsApi.ts
```diff
@@ -10,187 +28,70 @@ import type {
...
+  await inventoryApi().apiViewsDeleteView({ viewId: id });
+};
 
-  // No return value for DELETE
+export const cloneViewApi = async (
+  id: string,
+): Promise<ApiViewsCloneViewReturnType> => {
+  return (await inventoryApi().apiViewsCloneView({
+    viewId: id,
+  })) as unknown as ApiViewsCloneViewReturnType;
```
johnsonm325: We can leave this in in case we ever may need it, but I don't see an instance where we'd ever need to use it. Our use case is a user will choose a view that they want to replicate with a series of changes and they will simply make the changes and save it as a new view.

src/api/inventoryViewsApi.ts
```diff
@@ -10,187 +28,70 @@ import type {
...
-// Type aliases mapping to our domain terminology
 export type CreateViewRequest = ViewIn;
 export type UpdateViewRequest = ViewPatch;
 export type InventoryView = ViewOut;
 
-// ============================================================================
-// MOCK DATA (Replace with real API when available)
-// ============================================================================
-
 export const ALL_SYSTEMS_VIEW_ID = 'all-systems';
```
johnsonm325, line 36: we can remove this. `export const ALL_SYSTEMS_VIEW_ID = 'all-systems';`

src/api/inventoryViewsApi.ts
```diff
@@ -10,187 +28,70 @@ import type {
...
+  APIFactory(INVENTORY_API_BASE, endpoints, { axios });
 
-  // Simulate API delay
-  await new Promise((resolve) => setTimeout(resolve, 300));
+export const listViewsApi = async (
+  params: ApiViewsGetViewsListParams = {},
+): Promise<ApiViewsGetViewsListReturnType> => {
+  return (await inventoryApi().apiViewsGetViewsList(
+    params,
+  )) as unknown as ApiViewsGetViewsListReturnType;
```
johnsonm325, line 54: These `as unknown as` casts are sus. I asked AI how we might fix this, and it seems the best fix would be on the `javascript-clients` side, thought I'm not quite sure how it would be done:

```
This pattern is used across all HCC apps that consume generated clients — it's not unique to this PR. The proper fix would be in @redhat-cloud-services/javascript-clients-shared making APIFactory preserve the generic return types.

For this PR specifically, I'd say the as unknown as pattern is acceptable since it matches the established codebase convention. Fixing it here but nowhere else would create inconsistency.
```

johnsonm325: I've proposed a solution in slack

johnsonm325 (commented):
I don't want to approve or request changes yet on this, but I do have some suggestions.

Also, I tried to rename a view that I had just created and it didn't work. Not sure if it's user error or what, because I see that you did it.

Image

computercamplove:
> Also, I tried to rename a view that I had just created and it didn't work. Not sure if it's user error or what, because I see that you did it.

interesting, i did not have this issue

1. i want to clarify if you tried to remove ALL_SYSTEMS_VIEW_ID locally before you started to test my change - it messed up active view - "not allowed method" 405 usually happening when there's wrong type id, like empty list/string instead of "id" - maybe the id you sent to update was not of "new one" view?

2. did you run npm ci to  update the client?

src/api/inventoryViewsApi.ts
computercamplove: @johnsonm325 but backend doesn't provide it - when we delete custom view we get set to default "All systems" view, did you mean move it from API.ts file to `InventoryViews.ts` where we set active view?

src/api/inventoryViewsApi.ts
computercamplove: removed

src/api/inventoryViewsApi.ts
computercamplove: asked in slack

raswonders (approved):
lgtm ... Listing works, Rename, Save as and Delete mutations work as well. Code looks fine too.

raswonders (approved):
lgtm ... Listing works, Rename, Save as and Delete mutations work as well. Code looks fine too.

src/api/inventoryViewsApi.ts
johnsonm325: Let's keep it then as a fallback. thanks!

johnsonm325 (approved):

PR #3449, opened 2026-08-11
feat(RHINENG-29494): add Ansible workload hook for /ansible/inventory route

computercamplove:
## Jira
[RHINENG-29494](https://redhat.atlassian.net/browse/RHINENG-29494)

## What
add Ansible workload hook for /ansible/inventory route

## Why
As part of the global tag filter decommission, this PR replicates the Ansible workload pre-selection that insights-chrome previously injected via `GLOBAL_FILTER_UPDATE` for `/ansible/* `routes. The implementation is per-app (in inventory), not in chrome - as agreed by the Frontend CoP.

## How

- New `useAnsibleWorkloadDefault` hook in `InventoryViews/hooks/` sets `?workloads=ansible` in the URL on initial page load for the /ansible/ bundle and provides default filter config
- `SystemsView` accepts a new generic `defaultFilters` prop that controls "Reset filters" behavior - no ansible-specific logic in `SystemsView`. Other teams can reuse this prop if they need default filters for their routes in the future.
- `DataViewFiltersProvider` uses `defaultFilters` to restore defaults on "Reset filters" instead of clearing to empty
- Toolbar shows "Reset filters" (instead of "Clear filters") when defaults are active
- Removed `useSearchParamsWithFragment` hook - it existed solely for the global filter's hash-fragment merging, which is no longer needed

## Testing
Before testing - make sure flag is enabled to hide global tag filter https://insights-stage.unleash.devshift.net/projects/default/features/platform.chrome.hide.global-filter

1. Visit /ansible/inventory - verify `?workloads=ansible` in URL and "Ansible Automation Platform" chip visible
2. Verify table only shows Ansible systems
3. Add other filters (name, status), click "Reset filters" - verify ansible workload stays, other filters cleared
4. Click X on the ansible chip - verify filter is removed, all systems shown
5. Refresh page - verify ansible filter is re-applied
6. Visit /insights/inventory - verify no workload filter is applied, button says "Clear filters"

image

[RHINENG-29494]: https://redhat.atlassian.net/browse/RHINENG-29494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce an Ansible-specific default workload filter for inventory routes and integrate default filter handling into the shared systems view.

New Features:
- Add an Ansible workload default hook that pre-selects the ansible workload for /ansible inventory routes and exposes corresponding default filters.
- Allow SystemsView to accept configurable default filters that persist across page refreshes and can be reused by other routes.

Enhancements:
- Update the filters context and toolbar to distinguish between having default filters and not, showing a Reset filters action that restores defaults instead of clearing everything.
- Remove the legacy search-params-with-fragment helper now that global tag filter hash-fragment merging is no longer needed.

Tests:
- Adjust SystemsView-related tests to account for the new filters context property indicating whether default filters are active.


raswonders:
I'm sporadically not able to clear the ansible default filter on first click:

https://github.com/user-attachments/assets/0065c7d0-0c0b-44ab-997a-d99539770c73

computercamplove:
> I'm sporadically not able to clear the ansible default filter on first click:
> 
>  recording_20260811_170415.mp4

@raswonders i pushed second commit, i couldn't find solution that could be made only in ansible workload hook:

# Problem

Clicking the chip X to remove the Ansible workload default filter intermittently required two clicks.

# Root cause

When a filter chip is removed, `useResetPage` fires to reset pagination to page 1. It called PF's `pagination.onSetPage`, which internally rebuilds the full URL from a `searchParams` React-state closure. If React Router hasn't flushed the filter removal into that state yet, `onSetPage` resurrects the removed `workloads=ansible` param in the URL.

# Fix

Replaced `pagination.onSetPage` with a direct `setSearchParams` call that reads from `window.location.search` (always up-to-date since `history.pushState` is synchronous) and only sets `page=1`. This way the hook never touches filter params, so it can't resurrect removed ones. The pagination hook's own `useEffect` on `searchParams.toString()` syncs its internal page state from the URL on the next render.

  Changes

  - useResetPage.ts — removed pagination param and onSetPage call; directly writes page=1 to the URL
  - SystemsView.tsx — dropped pagination argument from useResetPage call
  - useResetPage.test.tsx — removed pagination stub and onSetPage assertions

src/components/SystemsView/SystemsView.tsx
```diff
@@ -260,6 +265,20 @@ const SystemsViewInner = ({
             <DataViewToolbar
               ouiaId="systems-view-header"
               clearAllFilters={clearAllFilters}
+              customLabelGroupContent={
+                hasDefaultFilters ? (
```
raswonders, line 269: Do we really want to display Clear filters & Reset filters on different occassion?

I think it would be completely fine if we render Reset filters all the time whether we have some default or not.. make it simple for customers and ourselves

raswonders (approved):
Approve gj! I left a note as food for thought I'd love to simplify reseting and one button seems like right answer to me. What do you think?

In regards to fix:

Fix by nature is little hacky as it bypasses state updates via lib but does them raw via browser API. But because React Router is limited in this area it's one of the best options we currently have. We will probably refactor this later on into group update of url params.

src/components/SystemsView/SystemsView.tsx
computercamplove: @raswonders Clear filter when default filters are not applied on the table
Reset filter - we reset filter to default filters if table has it

The reason why inventory legacy table still has Reset filter because nobody fixed it after we removed default filters. I am fine to make it same as Legacy table.

PR #3453, opened 2026-08-12
feat(RHINENG-29807): Load view configuration into SystemsView

johnsonm325:
## Jira
[RHINENG-29807](https://redhat.atlassian.net/browse/RHINENG-29807)

## What
Apply saved view configuration (columns, sort, filters) to the table when a view is selected. Replace localStorage column persistence with server-side view configuration, add sortBy keys to column definitions, and introduce view config filter serialization/deserialization.

[RHINENG-29807]: https://redhat.atlassian.net/browse/RHINENG-29807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Apply inventory view configurations to SystemsView so that column visibility, sort order, and supported filters are restored when a view is selected and synchronized with URL search parameters.

New Features:
- Load saved view configurations into SystemsView, applying configured columns, initial sort, and supported filters when a view is selected.
- Synchronize view-based filter state with URL search parameters when switching or creating views.

Enhancements:
- Replace local column persistence with configuration-driven column selection based on backend view definitions.
- Add sortBy metadata to several SystemsView columns to support server-side ordering and view-config-driven sorting.

Tests:
- Add unit tests for view configuration filter serialization/deserialization between toolbar state and backend format.
- Add unit tests for translating view column configurations into column selectors that control visibility and order in SystemsView.


src/components/InventoryViews/InventoryViews.tsx
```diff
@@ -14,31 +16,93 @@ import ViewDeleteModal from './Modals/ViewDeleteModal';
...
   const [activeViewId, setActiveViewId] = useState(ALL_SYSTEMS_VIEW_ID);
   const queryClient = useQueryClient();
   const isInventoryViewsPrivateEnabled = useInventoryViewsPrivateFeatureFlag();
   const { data: viewsData } = useViewsQuery();
   const viewsList = viewsData?.results ?? [];
   const activeView = viewsList.find((v) => v.id === activeViewId);
   const isSystemView = activeView?.is_system_view ?? true;
+  const viewsLoaded = !!viewsData;
+
+  const columnSelector = useMemo(
```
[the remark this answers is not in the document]
johnsonm325: Won't expose view identity to SystemsView. It's a reusable table component and doesn't need to be concerned with the view id.

johnsonm325:
@sourcery-ai regarding your overall comments, `useSearchParamsWithFragment was removed in commit 01eebdd2. and sortBy is the correct configuration mapping based on the backend api identifiers. There are currently columns that don't appear in the saved columns but a fix is being implemented for those.

src/components/InventoryViews/InventoryViews.tsx
```diff
@@ -16,39 +18,89 @@ import {
...
   const queryClient = useQueryClient();
   const isInventoryViewsPrivateEnabled = useInventoryViewsPrivateFeatureFlag();
   const { data: viewsData } = useViewsQuery();
   const viewsList = viewsData?.results ?? [];
-  const activeView = viewsList.find((v) => v.id === activeViewId);
+  const activeView = useMemo(
+    () => viewsList.find((v) => v.id === activeViewId),
+    [viewsList, activeViewId],
+  );
   const isSystemView = activeView?.is_system_view ?? true;
```
raswonders: only consumer of isSystemView variable is ViewToolbar and it already defines default value. I'd refactor this and pass `activeView?.is_system_view` directly to ViewToolbar component via its prop.

src/components/InventoryViews/InventoryViews.tsx
```diff
@@ -16,39 +18,89 @@ import {
...
+  const viewsLoaded = !!viewsData;
```
raswonders: this is fragile and could be considered as anti-pattern. Query's result has bunch of lifecycle related state in it which we can can safely consume. Consuming `isSuccess` flag seems more appropriate here than !!viewsData 
https://tanstack.com/query/latest/docs/framework/react/guides/queries

src/components/InventoryViews/createViewColumnSelector.ts
raswonders, line 1: It's little confusing why we're selecting based on sortBys here. 

I think it would be more future proof if we sync keys we have in UI for columns with the ones backend holds. Backend should be source of truth for our key= properties in column definitions. Once thats is done we could filter catalogue in selector base on  the key= properties.

raswonders (commented):
Thanks for the PR before my approval I'd like to clarify these:

1. What we're going to do about difference between keys we keep localy in columnDefs and ones which are defined in the backend. I believe we should unify them and use those from Backend.
2. I see some Views are send back with empty columns. Are we fine with it or backend should not allow this to happen? 
My take on this is that we shouldn't allow states wich make no sense for the APP. View without columns makes no sense to me. Currently we display InventoryColumns for such views. Which seems like a an odd choice. I'd expect empty table for such cases. Which is unusable but reflects true reality and state of the view.

johnsonm325:
@raswonders good call on this:

> I see some Views are send back with empty columns. Are we fine with it or backend should not allow this to happen?
My take on this is that we shouldn't allow states wich make no sense for the APP. View without columns makes no sense to me. Currently we display InventoryColumns for such views. Which seems like a an odd choice. I'd expect empty table for such cases. Which is unusable but reflects true reality and state of the view.

We're currently falling back to a hard-coded All systems view until we actually get that view from the backend. Then we can update the code to always only show the columns provided by the config.

raswonders:
> We're currently falling back to a hard-coded All systems view until we actually get that view from the backend. Then we can update the code to always only show the columns provided by the config.

Curious if we talk about same or different things. What I meant was.. we have Views in select dropmenu which have no columns in them (in response body columns={}). Those are still selectable and when we select them we display Inventory columns instead of what I'd expect no columns.

johnsonm325:
> Curious if we talk about same or different things. What I meant was.. we have Views in select dropmenu which have no columns in them (in response body columns={}). Those are still selectable and when we select them we display Inventory columns instead of what I'd expect no columns.

@raswonders Yes, we are talking about the same thing. I've updated the code to display no columns when the views api returns a view with no columns in the configuration. Currently, the api does not have a All systems fallback view (for when a user has no views saved, or they delete the current view they are on), so we have a hard-coded All systems view. I have this view coded with the inventory columns. So once we remove it, everything should work as expected.

raswonders (approved):
thanks for adding those updates lgtm

raswonders (approved):
thanks for adding those updates lgtm

PR #3488, opened 2026-08-20
fix(RHINENG-29643): Add scroll bar and Show more link to View selector

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-29643

## What
View selector dropdown should have a maximum height with scroll, and a "Show more" link at the bottom.

## Why
UX

## Testing (insights-qa account)
1. Navigate to Systems page in stage and switch to preview mode
2. open Views selector menu
3. Verify if more than 50 views "show more" button is displayed
4. verify scroll bar is present

## Summary by Sourcery

Enable users to browse large view lists with scrolling and incremental loading.

New Features:
- Add a scrollable view selector with a Show more option for loading additional views.
- Support paginated loading of inventory views in batches of 50.

Enhancements:
- Expose pagination state through the inventory views toolbar and selector.

### DIFF
--- a/src/components/InventoryViews/InventoryViews.tsx
+++ b/src/components/InventoryViews/InventoryViews.tsx
@@ -100,8 +100,16 @@ const InventoryViews = () => {
   const [activeViewId, setActiveViewId] = useState(ALL_SYSTEMS_VIEW_ID);
   const queryClient = useQueryClient();
   const isInventoryViewsPrivateEnabled = useInventoryViewsPrivateFeatureFlag();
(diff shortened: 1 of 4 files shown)


raswonders (approved):
I can confirm height is same as Workspace filter dropwdown, and load more works same way. Code looks good too. I don't see many opportunities to reuse code here from Workspace filter.

Only thing which seems odd is the placement of default views and that it's fighting load more pattern. I've opened discussion for this but this pr lgtm. Thanks

PR #3517, opened 2026-08-25
feat(RHINENG-30319): Show 4 advisories in installable advisories

johnsonm325:
## Jira
[RHINENG-30319](https://redhat.atlassian.net/browse/RHINENG-30319)

## What
The Installable advisories column should show all 4 values, even if one of the values is 0. If the value is 0, there is no link. The links have also been updated to link to the system details page, content table, and filter applied for given advisory. But if there are no advisories at all, it still shows "No installable advisories".

image

image

[RHINENG-30319]: https://redhat.atlassian.net/browse/RHINENG-30319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Show all installable advisory categories while linking only available counts to their corresponding filtered system views.

New Features:
- Display all four installable advisory categories, including zero-valued counts.
- Link non-zero advisory counts to the system’s filtered inventory patch view by advisory type.

Bug Fixes:
- Preserve the no-installable-advisories state when all advisory counts are zero and the unavailable state when advisory data is missing.

Tests:
- Update installable advisory coverage for zero-count visibility, filtered links, and non-linking zero values.

### DIFF
--- a/src/components/SystemsView/columns/content/cells/InstallableAdvisories.test.tsx
+++ b/src/components/SystemsView/columns/content/cells/InstallableAdvisories.test.tsx
@@ -7,7 +7,8 @@ import { TestWrapper } from '../../../../../Utilities/TestingUtilities';
 import type { PatchAppData } from '@redhat-cloud-services/host-inventory-client';
 
 const systemId = 'test-system-uuid';
-const PATCH_SYSTEM_PATH = `//patch/systems/${systemId}`;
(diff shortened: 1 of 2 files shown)


computercamplove (approved):
- expected filter is applied when redirecting to Content tab
- all values are displayed even one of the value is 0

PR #3518, opened 2026-08-25
feat(RHINENG-28363-2): Column changes not recognized as dirty state

computercamplove:
**Issues:**
  - Column changes not recognized as dirty state
  - "Save" button sometimes triggers "Save as" modal (race condition)
  - "Reset to default" in Manage Columns restores wrong defaults for newly created views

  **Fixes:**
  1. Remove `hasColumnConfig` gate in `useViewDirtyState` - column edits now always count as dirty
  2. Add `event.stopPropagation()` to primary split-button action - prevents click bubbling
  3. Change `columnSelector` memo dependency from `[activeViewId, viewsLoaded]` to `[activeView?.configuration]` — ensures defaults update when view configuration loads

## Testing
1. not owned view: modified columns triggers dirty state - "Save as" split button
2. owned view : modified columns triggers dirty state - "Save" split button

https://github.com/user-attachments/assets/e973da17-31a2-4587-b66f-3a4d0cec896e

## Summary by Sourcery

Ensure column edits reliably drive view dirty state and save behavior while preserving the correct defaults across view changes.

Bug Fixes:
- Recognize column edits as dirty state for both saved and system views, including views without persisted column configuration.
- Prevent the primary Save action from opening the split-button menu through event bubbling.
- Restore and apply the correct column defaults when switching views or loading view configuration.

Enhancements:
- Derive column dirty-state baselines from the active view configuration and reset local edits after view changes or successful saves.

Tests:
- Add coverage for selecting current versus baseline columns and recalculating column selectors when views change.

Chores:
- Update manage-views end-to-end expectations to preserve existing filters during column changes.

### DIFF
--- a/_playwright-tests/test_manage_views.test.ts
+++ b/_playwright-tests/test_manage_views.test.ts
@@ -116,12 +116,6 @@ test.describe(
     const configurationUpdatedC = {
       columns: [...vulnerabilityColumns, ...malwareColumns],
       columnsCount: vulnerabilityColumns.length + malwareColumns.length + 3, // +3 for checkbox, name and per-row actions
-      filters: [
-        {
-          filter: 'System type',
-          value: ['Image-based system', 'Package-based system'],
-        },
-      ],
     };
 
     test('User creates custom views with own configuration', async ({
@@ -304,13 +298,6 @@ test.describe(
         }
         await manageColumnsModal.save();
 
-        // Applies new filter
-        await filterSystemsWithConditionalFilter(
-          page,
-          'System type',
-          'Package-based system',
-        );
-
         await manageView.save(viewC);
         await manageView.verifyActiveView(viewC);
       });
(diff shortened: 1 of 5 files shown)


johnsonm325 (commented):
This PR doesn't allow you to "Save as" with the Tags column. Currently, in stage, we are able to do that, but this PR I think includes the Tags column in the request.

computercamplove:
> This PR doesn't allow you to "Save as" with the Tags column. Currently, in stage, we are able to do that, but this PR I think includes the Tags column in the request.

@johnsonm325 i was not able to reproduce it with Tags - did you mean "Status" column? - we still have issue with "Status" column 400.

Tags shouldn't cause 400 error, but this column currently still can't be saved to custom view config. The only fix made for Tags column is that Tags was added to core-columns for "All systems" view (when you create new view with tags columns - it will not be in the custom view config, but if you switch back to "All systems" Tags column will be in config).

johnsonm325 (approved):

raswonders (approved):
tried it and it works, code looks fine although comments are quite verbose which can be a smell.

PR #3523, opened 2026-08-27
feat(RHINENG-30387): Support saving all host-level filters in view configurations

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-30387

## What changed
- Dependency/API: host-inventory-client 6>7; switched `apiViewsUpdateView` > `apiViewsPatchView`.
- Save: new `buildHostFilters.ts` maps toolbar state > the nested host block (hostname, staleness, source, tags, workspace, system_type, and the last_check_in date range).
- Restore: `parseViewConfigFilters` restores every host field; new `resolveLastSeenKeyFromBounds` classifies saved dates back to the right Last seen key (configs store only dates, not the preset key), and `parseViewConfigLastSeenCustomRange` seeds the custom date pickers.
- Wiring: custom range surfaced up via `onLastSeenCustomRangeChange`.

## Testing
- Create view with all filters applied (switch between views to verify : UI displays applied filters)

https://github.com/user-attachments/assets/4605a07e-d1d1-4d89-87f0-fa6e91f455e8

## Summary by Sourcery

Persist and restore the complete set of host-level filters in inventory view configurations.

New Features:
- Persist all host-level inventory filters in view configurations, including hostname, staleness, source, tags, workspace, system type, and last-seen ranges.
- Restore saved host filters and custom last-seen date ranges when switching between views.

Bug Fixes:
- Preserve open-ended custom last-seen ranges instead of requiring both date bounds.

Enhancements:
- Track custom last-seen range changes as part of view dirty-state detection.
- Classify saved last-check-in bounds back to the corresponding preset or custom last-seen selection.

Build:
- Upgrade the host inventory client dependency to version 7 and use the view patch API.

Tests:
- Add coverage for host-filter serialization, restoration, last-seen range handling, and dirty-state detection.


computercamplove:
@raswonders @johnsonm325 copying it here:

 i'd really appreciate extra attention on Last seen :skull: reason is everything else relies on the URL filter state, but the custom Last seen range isn't stored in the URL, so it needed extra work to save, restore, and detect as dirty and that's the part most likely to have issues :grimacing: i provided the screenrecording that everything is working but still - need your attention. if you think Custom last seen better be fixed in separate jira, please let me know.

src/components/InventoryViews/utils/viewConfigFilters.ts
```diff
@@ -75,38 +87,101 @@ export const parseViewConfigFilters = (
...
+ * with parseViewConfigFilters, which restores `last_seen: 'custom'`.
+ *  @param viewFilters - ViewConfiguration filters to parse
+ *  @returns           Start/end range, or null when no host date bounds are saved
+ */
+export const parseViewConfigLastSeenCustomRange = (
+  viewFilters: ViewFilters | undefined,
+): LastSeenCustomRange => {
+  if (!viewFilters) return null;
+
+  const raw = viewFilters as unknown as Record<string, unknown>;
```
raswonders: same type is being inferred so this type cast is unnecessary

src/components/InventoryViews/utils/viewConfigFilters.ts
```diff
@@ -75,38 +87,101 @@ export const parseViewConfigFilters = (
...
+  const hostFilters = raw.host as HostFilters | undefined;
```
raswonders: here as well, same type is being inferred so this type cast is unnecessary

raswonders (commented):
found a bug when we Select lastSeen => Custom, table is dirty it shouldn't be and it shouldn't be possible to save this state cause we're not filtering yet. Also when we actually click save in this screen state never gets cleaned and we can click forever.

Image

raswonders (approved):
/lgtm feature seems to be working now thanks for fixes :)

PR #3536, opened 2026-08-31
fix(InventoryViews):set All systems on top of menu

computercamplove:
1. Fix All systems view: always present, pinned, and selected by default (_default selection currently falls back to the All systems view. Once the "Set as default" feature (per-user default view API) is introduced, this will be updated to select the user's chosen default instead_)
2. fix menu option order in Manage view

## Summary by Sourcery

Keep the All systems view pinned as the default inventory view and align Manage view actions with the intended order.

Bug Fixes:
- Ensure the All systems view is always available, selected by default, and restored after deleting the active view.
- Correct the Manage view menu option order.

Enhancements:
- Load all inventory view pages eagerly so system views can be identified regardless of pagination.

Tests:
- Update the systems filter test to reflect the revised view ordering.

### DIFF
--- a/_playwright-tests/test_systems_filter.test.ts
+++ b/_playwright-tests/test_systems_filter.test.ts
@@ -100,7 +100,7 @@ test.describe('Filtering Systems Tests', { tag: ['@systems-table'] }, () => {
       })
       .or(
         page.locator(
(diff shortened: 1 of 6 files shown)


raswonders (approved):
lgtm thanks for the fix :)

PR #3558, opened 2026-09-03
fix(RHINENG-30483): Prevent duplicate view saves requets from the Save split button

computercamplove:
## Jira
https://redhat.atlassian.net/browse/RHINENG-30483

## What

- Prevent duplicate view saves requets from the Save split button
- disable Save split button when `isSaving` is pending

## Why
When dirty state is true user able to click Save multiple times (same multiple PATCH requests are send). The root cause is timing between success save/PATCH request and dirty state verification (the Save is done and dirty state is still `true` -  refetch hasn't landed yet - that re-enables Save button ). 

## How
the Save button now stays disabled until the mutation and its refetch complete, driven by `updateView.isPending`

## Summary by Sourcery

Prevent duplicate view saves by tracking the update request state across the view toolbar.

Bug Fixes:
- Prevent duplicate view updates by blocking repeated Save actions while a save request is in progress.

Enhancements:
- Disable view management controls during saving and wait for view data refreshes to complete after a successful update.

### DIFF
--- a/src/components/InventoryViews/InventoryViews.tsx
+++ b/src/components/InventoryViews/InventoryViews.tsx
@@ -226,10 +226,6 @@ const InventoryViews = () => {
     currentLastSeenCustomRange,
   });
 
-  useEffect(() => {
-    if (isSaving && !isViewDirty) setIsSaving(false);
-  }, [isSaving, isViewDirty]);
-
   const handleSelectView = useCallback(
     (viewId: string) => {
       setActiveViewId(viewId);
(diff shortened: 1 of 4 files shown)


src/components/InventoryViews/InventoryViews.tsx
```diff
@@ -155,6 +155,7 @@ const InventoryViews = () => {
   const activeView = viewsList.find((v) => v.id === activeViewId);
   const isSystemView = activeView?.is_system_view ?? true;
   const viewsLoaded = !!viewsData;
+  const [isSaving, setIsSaving] = useState(false);
```
raswonders: Is this state necessary it seems to me it tracks similar state which updateView.isPending is tracking. Could we use that instaed?

raswonders (commented):
It seems like we solve the problems but there are some quirks.

When mutation is pending we still allow clicks on "Save" Primary action. Meanwhile dropdown "Save" action is disabled. This is slightly confusing because button does nothing but we can still click it. It would be best if we could disable Primary action as well while mutation is ongoing. 

I've also left note about isSaving which seems like extra state which we might not need.

src/components/InventoryViews/InventoryViews.tsx
computercamplove: make sense , thank you!

raswonders (approved):
Button works now thank you! lgtm

PR #3559, opened 2026-09-03
fix(InventoryViews): Stop query all pages for views on page load

computercamplove:
## What
Stop query all pages for views on page load

## Why
Backend is fixed: Prioritize system views ahead of user views in the views listing  https://github.com/RedHatInsights/insights-host-inventory/pull/4887

## Testing
- When fisrt time Systems page is visited in network tab only 1 request for views is visble
image
- The "All systems" option is on top

Before qe-u-inventory account:
image

## Summary by Sourcery

Load the initial inventory views page only and fetch additional pages when requested.

Bug Fixes:
- Prevent inventory views from eagerly fetching every page when the view selector loads.

Enhancements:
- Keep view pagination on demand while preserving the ability to load additional views from the selector.


raswonders (approved):
lgtm thanks for the fix :)

PR #3561, opened 2026-09-03
feat(RHINENG-30530): Clear url params when loading a view

johnsonm325:
## Jira
[RHINENG-30530](https://redhat.atlassian.net/browse/RHINENG-30530)

## What
This PR clears url parameters when switching to a different view, whether via the View selector or by refreshing the page. Params will populate in the url when applying them on a view, but will disappear when switching.

[RHINENG-30530]: https://redhat.atlassian.net/browse/RHINENG-30530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Clear stale Inventory Views URL parameters when loading or switching views while preserving saved configuration and bundle-level defaults.

New Features:
- Normalize Inventory Views URLs on page load and view changes so they contain only the selected view configuration and applicable bundle defaults.

Bug Fixes:
- Prevent stale filters, sorting, and pagination parameters from persisting across refreshes or view switches.
- Prevent SystemsView from hydrating before the URL has been normalized.

Tests:
- Add coverage for URL parameter generation, default filter preservation, URL normalization, view switching, and lifecycle scenarios.

### DIFF
--- a/src/components/InventoryViews/InventoryViews.test.tsx
+++ b/src/components/InventoryViews/InventoryViews.test.tsx
@@ -2,6 +2,9 @@
  * Tests for InventoryViews.tsx
  */
 
+import type { ViewConfiguration } from '../../api/inventoryViewsApi';
+import { SORT_URL_PARAM, SORT_DIR_URL_PARAM } from '../SystemsView/constants';
+
 describe('InventoryViews component logic', () => {
   describe('getCurrentConfiguration behavior', () => {
     it('should prefer currentColumns when defined', () => {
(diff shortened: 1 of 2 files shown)


johnsonm325:
This PR will cause prod-preview to reset url parameters on a refresh. After talking with Jan, we're going to let prod-preview temporarily have these scenarios because we will be releasing Inventory Views within weeks.

raswonders:
State should survive page refreshes, our other table works in same way. Its common practice. So I'm curious why we go different route.  

I've tried to experiment with UI and Url is cleared when we switch Views. even without this pr. Could you help me understand which app issues we're trying to address by this?

johnsonm325:
@raswonders This is more of an MVP for now. The main reason we're addressing it like this is because of the weird situation where you would load in a dirty view and that's not something I think we're wanting to do from a UX perspective.

I believe our goal is to implement some version of state persistence between refreshes, but for Q3 we're focusing on just making it consistent by ensuring users load into a clean state and defaulting to loading the default view.

raswonders:
> @raswonders This is more of an MVP for now. The main reason we're addressing it like this is because of the weird situation where you would load in a dirty view and that's not something I think we're wanting to do from a UX perspective.
> 
> I believe our goal is to implement some version of state persistence between refreshes, but for Q3 we're focusing on just making it consistent by ensuring users load into a clean state and defaulting to loading the default view.

Please help me understand the reason, why we want prevent users load into draft/dirty state. Are there any acute problems we can't ignore and trying to work around by this. 

In my eyes loading potentially into draft state is OK and desired. I thought we shared common understanding regarding of View URL state management and its future but I might be missing something. Let's sync up tomorrow.

johnsonm325:
@raswonders We did agree, but then after talking with Jan, Susan, and I believe Talia (possibly Jerome too), we agreed to land on this solution for now. They are all aware of the scenario. It's a temporary solution to manage how we handle loading into a dirty state or not.

In my opinion, loading into a dirty state is not the route we want to go. It can cause confusion for the user.

raswonders:
@johnsonm325 

I realize managing state is one of the hardest problems in frontend. It's also highly influential on whole stack so its only understandable we can have disagreements about it if our understanding differs.   

Let me lay out few reasons why I think this approach might not be the best:

1. Breaks web standard and what is normally expected. Sometimes page reloads for unexpected reasons (not invoked by user) this will reset state and move user to different View than he intented
1. Different pattern than seen in all LightSpeed tables (They all persist state across refresh) 
2. Users won't be able to use bookmarks anymore
3. Clashes with changes of filter revamp PR. The PR changed how default filters are applied to SystemsView.

Looking into current UI behavior, I've noticed that the page refresh currently carries over filter state only and not the knowledge which View is selected. As we talked last time this will be resolved once we implement view_id as URL param. After that whole bag of state gets carried over and nothing needs to be nullified. 

I see Views as already pretty working MVP. We're not getting much benefit by applying this change but introducing quite bit of tech debt. As you've said we would need actually walk back from this at some point. So you can see why I'm not totally persuaded it's worth it.

PR #3578, opened 2026-09-08
feat(RHINENG-30488): expose filters as configurable prop on SystemsView

raswonders:
RHINENG-30488

## What
Complete revamp of filters inside SystemsView + additional wiring in consumers InventoryViews and InventoryHosts

## Why
We have two rendering sites with growing needs for filter configurability and more are comming from third party apps.

## How
- moves filter configuration from being hard-coded in components into catalog of filter specs  
- removes static InventoryFilters state and introduces dynamic one instead which is based on filters chosen
- adds configurable prop `filters` for consumers internal and external
- shares catalog of predefined filter specs which consumers can select and modify during runtime
- we allow use of custom filters via `catalog.custom()` factory
- `Reset filters` behavior changed. Button now appears when filter values are different from defaults. 

## Testing
There were significant changes to filters and Views as well. Please perform careful manual UI tests  of filters in InventoryViews. 

BONUS points: try to configure filters via new prop and let me know how difficult it seems.  

Example of using filters prop inline:
image


src/components/SystemsView/SystemsView.tsx
```diff
@@ -69,13 +68,22 @@ import {
   resolveColumnSelector,
   type ColumnSelector,
 } from './columns/resolveColumnSelector';
+import {
+  resolveFilterSelector,
+  type FilterSelector,
+} from './filters/resolveFilterSelector';
+import { buildFilterParams as buildFilterParams } from './filters/buildFilterParams';
```
johnsonm325: Probably don't need the `as buildFilterParams` here.

src/components/SystemsView/filters/inventory/bindInventoryFilters.ts
```diff
@@ -0,0 +1,193 @@
...
+ *  @param catalog - Shared filter catalog of named factories
+ *  @returns       Bound host-list filters in toolbar order
+ */
+export const bindInventoryHostListFilters = (
+  catalog: FilterCatalog,
+): BoundFilter<ApiHostGetHostListParams>[] => [
+  catalog.hostname({
+    updateFilterParams: (params, value) => ({
+      ...params,
+      ...(value && { hostnameOrId: value }),
```
johnsonm325, line 33: Can `value` be an array here? If so, should we update this to `...(value.length > 0 && { systemType: buildSystemType(value, ...) }),`

johnsonm325 (commented):
I had to pause my review, but I'll pick it back up tomorrow (Thursday) morning.

src/components/SystemsView/filters/inventory/bindInventoryFilters.ts
raswonders: It cannot be array in this case. Hostname is `Text` type filter. As such it deals with strings only. The value is typed as well so you would get an compiler Error if you would use array here.

computercamplove (approved):
LGTM!

src/components/SystemsView/filters/inventory/filterDefinitions.tsx
```diff
@@ -0,0 +1,242 @@
...
+  type TextFilterSpec,
+} from '../types';
+
+const WorkspaceChip = ({ id }: { id: string }) => {
+  const { filters } = useDataViewFiltersContext();
+  const groupId = filters[SYSTEMS_VIEW_WORKSPACE_FILTER_PARAM];
+  const { names, isFetching, ids, pendingLabelFetchIds } =
+    useWorkspaceDisplayNames(
+      Array.isArray(groupId) ? (groupId as string[]) : undefined,
+    );
```
johnsonm325, line 34: > `WorkspaceChip` calls `useWorkspaceDisplayNames` per chip instead of once in the parent. React Query dedupes the network call, but it's N hook instances/re-renders vs the old single call.

Do we think this is worth a fix? The re-render isn't ideal, but also not a heavy load.

src/components/SystemsView/filters/buildFilterParams.ts
```diff
@@ -0,0 +1,25 @@
...
+export const buildFilterParams = <TFilterParams>(
+  specs: readonly BoundFilter<TFilterParams>[],
+  filters: SystemsViewFilterState,
+  ctx: FilterSelectContext,
+): TFilterParams =>
+  specs.reduce(
+    (params, filter) =>
+      filter.updateFilterParams(
+        params,
+        filter.getValue?.(filters, ctx) ?? filters[filter.filterId],
```
johnsonm325: > `getValue?.(...) ?? filters[filterId]` means a `getValue` that intentionally returns undefined gets overridden by the raw URL value. Benign today, latent trap for future custom filters.

Do we see this as being a potential issue? Maybe we just call out in documentation to ensure that `undefined` is not returned intentionally?

src/components/SystemsView/filters/inventory/filterDefinitions.tsx
```diff
@@ -0,0 +1,242 @@
...
+  getValue: (filters, ctx): LastSeenSelectValue => ({
+    key:
+      typeof filters.last_seen === 'string'
+        ? (filters.last_seen as LastSeenKey | '')
+        : '',
+    range: ctx.lastSeenCustomRange,
+  }),
+} satisfies CustomFilterSpec<LastSeenKey | ''>;
+
+export const tagsSpec = {
```
johnsonm325, line 185: > tags lost its custom onChange - the one multi-group filter now relies on the generic handler. Probably fine, but ask for confirmation there's a multi-group tag select+chip test so it's not a silent regression.

The AI review seemed to think that the generic onChange handler would suffice, but I wanted to make sure that it was removed intentionally and we haven't regressed by accident.

johnsonm325 (approved):
I think there are some important verifications on your end in these comments I've added. I'm going to approve, but if anything raises a red flag for you, please address.

Nice job on this PR! Getting excited about seeing this implemented.

src/components/SystemsView/filters/inventory/filterDefinitions.tsx
raswonders: i've checked it. it's not a problem while there is few workspace chips might be an issue when there is many. Definitelly want to fix this but probably will create seperate card for this optimization. It's not straight forward I think and I'd also want to install Query devtools so we can analyze state live for other things as well.

src/components/SystemsView/filters/buildFilterParams.ts
raswonders: This is minor but I'm gonna cover it. I think we really want use result of getValue() even if its undefined. I removed ?? and use ternary instead.

src/components/SystemsView/filters/inventory/filterDefinitions.tsx
raswonders: Wasn't my intent to simplify this but apparently AI did right thing. Generic onChange seems work just fine I verified it.

PR #3596, opened 2026-09-10
feat(RHINENG-30443): active view id in url

johnsonm325:
## Jira
[RHINENG-30443](https://redhat.atlassian.net/browse/RHINENG-30443)

## What
Adds default view resolution and URL stamping to the Inventory Views feature. When navigating to /insights/inventory with no view_id query parameter (fresh landing or old bookmark), the app now:

1. Resolves the default view (currently the "All systems" system view)
2. Stamps view_id into the URL so a refresh restores the correct view
3. Preserves any existing filter params (dirty default view survives refresh)
4. Handles stale/deleted view IDs gracefully (falls back to default, keeps filters)

## Why
User pain addressed:

Before this change, navigating to `/insights/inventory` without a `view_id` would load the All Systems view correctly, but refreshing the page with filters applied would:
- Restore the filters (persisted in URL)
- Show the wrong view ID in state (all-systems sentinel instead of the real UUID)
- Create a mismatch between what's shown and what's active

Forward compatibility:

Backend PR #4903 introduces user-pinned default views via `get_default_view_id(org_id, user_id)`. This PR structures the frontend to slot in that resolution later — one clear swap point (`resolveDefaultViewId`) that today falls back to the system view and can later prefer a user's pinned default.

## Testing
Manual verification:

1. `/insights/inventory` → stamps `view_id=<system-view-uuid>`
2. `/insights/inventory?status=stale` → stamps view_id + keeps `status=stale`
3. Refresh with `view_id` + filters → preserves both
4. Deleted view ID in URL → falls back to default, keeps filters

[RHINENG-30443]: https://redhat.atlassian.net/browse/RHINENG-30443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Synchronize Inventory View selection with the URL so default and saved views reliably survive navigation and refreshes.

New Features:
- Persist the active Inventory View ID in the URL and restore it across refreshes and bookmarked navigations.
- Resolve missing or stale view IDs to the default system view while preserving existing filters.
- Keep view selection synchronized when switching, creating, or deleting views, and remove view IDs when returning to the stable inventory table.

Bug Fixes:
- Prevent mismatches between the displayed Inventory View and the active view state after refreshes or deleted-view links.
- Prevent an incorrect view from flashing while Inventory Views are loading.

Enhancements:
- Centralize default view resolution to support future user-pinned default views.

Tests:
- Add coverage for default, stale, valid, and loading view ID URL reconciliation, including filter preservation.
- Add unit tests for default system view resolution.


computercamplove:
when i switch from preview to stable version, the URL keeps the views_id - it doesn't break the stable version, yet i am not sure it should be in the URL

src/components/InventoryViews/InventoryViews.tsx
```diff
@@ -133,24 +139,49 @@ const InventoryViews = () => {
...
   const activeView = viewsList.find((v) => v.id === activeViewId);
   const isSystemView = activeView?.is_system_view ?? true;
   const viewsLoaded = !!viewsData;
 
+  // Reconcile activeViewId + URL against loaded views:
+  //  - No view_id in the URL: adopt the default view and stamp it in, keeping any
+  //    filter params already present (a dirty default view survives a refresh).
+  //  - A stale view_id (deleted view / old bookmark): fall back to the default,
+  //    dropping just the stale param while keeping any filters.
+  useEffect(() => {
```
raswonders, line 159: This reconciliation should happen in render step rather than effect. Cause we want to reconcile before component is shown.

src/components/InventoryViews/InventoryViews.tsx
raswonders, line 364: When I set View to custom one and refresh page, we get flash of ALL_SYSTEMS view before component correctly settles on view_id from URL. 

This could be avoided by holding off from rendering UI, until useViewsQuery is no longer pending. The query provides isPending boolean which we can use.

raswonders (commented):
I've tested it mostly works well. I've left two notes one is regarding view reconciliation timing which should run before browser paints and other one should solve flash of content we experience when we load in custom view_id.

src/components/InventoryViews/InventoryViews.tsx
```diff
@@ -128,29 +136,73 @@ const InventoryViews = () => {
...
   const isSystemView = activeView?.is_system_view ?? true;
   const viewsLoaded = !!viewsData;
 
+  // Reconcile activeViewId + URL against loaded views:
+  //  - No view_id in the URL: adopt the default view and stamp it in, keeping any
+  //    filter params already present (a dirty default view survives a refresh).
+  //  - A stale view_id (deleted view / old bookmark): fall back to the default,
+  //    dropping just the stale param while keeping any filters.
+  // Uses useLayoutEffect to reconcile before browser paint, preventing flash of wrong view.
+  useLayoutEffect(() => {
```
bastilian: I would consider using `useLayoutEffect` a bad practice and the multiple `returns`/guards are also a signal that something's not right here.

Most of what is happening within this `useLayoutEffect` can and should probably be done without it.

src/components/InventoryViews/InventoryViews.test.tsx
```diff
@@ -62,4 +62,151 @@ describe('InventoryViews component logic', () => {
       // Memo would recompute when switching views
     });
   });
+
+  describe('default view loading (reconcile effect)', () => {
```
raswonders: Sourcery is right here. I'd erase thes generated tests as they don't provide any value. We basically play with locally created copies of URLparams. Our production URL handling code could be totally wrong but these tests would be still passing.

raswonders (approved):
I tested the changes and it works nicely. You've also solved the flash of ALL_SYSTEMS problem so thanks for that! Only thing I'd do is to remove those view_id tests or substitute them with something of substance. Otherwise lgtm.

PR #3627, opened 2026-09-21
feat(RHINENG-29867): Add Set as default action to Manage view dropdown

johnsonm325:
## Jira
[RHINENG-29867](https://redhat.atlassian.net/browse/RHINENG-29867)

## What
Adds a **"Set as default"** action to Inventory Views (Gen 3), letting users pin a saved view as their default, and makes the app respect that default on load.

- New "Set as default" item in the **Manage view** dropdown (disabled with a tooltip when the active view is already the default, and while the request is in flight).
- On a fresh `/insights/inventory` visit, loads the user's pinned default from `GET /views` (`default_view_id`). A `?view_id=` in the URL always overrides it.
- Success/error toasts on setting a default.
- Deleting the active view navigates to the correct next view with a **clean URL** (no stale filters/sort carried over): the user's pinned default when a non-default view is deleted, or **All systems** when the default itself is deleted (mirrors the backend's read-time fallback).

## Why
Users want their preferred view to persist across sessions instead of always landing on "All systems." Pairs with backend RedHatInsights/insights-host-inventory#4969.

## How
- `setDefaultViewApi` wrapper + `useSetDefaultViewMutation` (toasts + `['views']` invalidation).
- `resolveDefaultViewId` now trusts the backend `default_view_id`, falling back to the system view only when it's absent.
- Delete fallback extracted to a pure `resolveViewIdAfterDelete` helper; the delete handler stamps the target `view_id` directly to avoid briefly flashing the just deleted view before the views query refetches.

## Testing
- `useSetDefaultViewMutation`: success toast + cache invalidation; error toast + no invalidation.
- `resolveDefaultViewId`: honors backend default, falls back correctly.
- `resolveViewIdAfterDelete`: non-default → pinned default; default → All systems; no-default and no-system-view edge cases.
- `ManageViewButton`: "Set as default" enabled/disabled states.
- ✅ 288/288 InventoryViews tests pass, `type-check` and `eslint` clean.

image

[RHINENG-29867]: https://redhat.atlassian.net/browse/RHINENG-29867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Enable persistent default Inventory Views and apply them consistently during navigation and view deletion.

New Features:
- Allow users to set a saved Inventory view as their default and identify default views in the selector.
- Load the backend-configured default view on inventory entry while preserving explicit URL view selections.
- Show success or error notifications when changing the default view.

Bug Fixes:
- Navigate to the correct fallback view with a clean URL when deleting the active view.
- Ensure configured filter defaults are stamped into the URL without overwriting existing filter values.

Enhancements:
- Centralize default-view resolution and post-deletion fallback behavior.
- Add unit and end-to-end coverage for default view selection, landing behavior, deletion fallback, and filter default handling.

Tests:
- Expand Inventory Views tests to cover setting defaults, default indicators, mutation outcomes, and deletion fallback scenarios.

Chores:
- Add the default-view API integration and React Query mutation support.


computercamplove:
@johnsonm325 E2E test is missing for Set default, it can be added to` 'Inventory Views: Manage views CRUD operations'` test suite

_playwright-tests/test_manage_views.test.ts
```diff
@@ -70,6 +71,30 @@ test.describe(
...
+      await test.step(`Reloading the inventory lands on the default view`, async () => {
+        // A fresh navigation carries no view_id, so the app must resolve the
+        // landing view from the backend default_view_id rather than falling back
+        // to "All systems". This also exercises fetching the pinned default by id
+        // when it isn't on the first loaded page of the views list.
+        await navigateToInventorySystemsFunc(page);
+        await manageView.verifyActiveView(defaultViewName, { timeout: 15000 });
+      });
+    });
   },
```
computercamplove, line 113: i would add here step:
delete default view -> verify All systems view now is default view

computercamplove:
@johnsonm325 i noticed that we have no indication in Select view menu what is view is default (blue check is indicator of an active view)
image

_playwright-tests/test_manage_views.test.ts
computercamplove: i would even add another step before my suggestion

after you set view to default 
1. create new view - select it/make it active
2 delete it -> verify we fallback to default view `defaultViewName` (`defaultViewName` is active)

computercamplove:
@johnsonm325 i found strange bug - dirty state gets enabled even if view configuration is correct and no changes is applied to this view, it is reproducible when you switch between views:
for example:
1. on custom view i set new filter by Name
2. i verify - the configuration is right
image

3. i switch to different view or chnage the page

4. i navigate back and select the view. configuration is right, but state is dirty - Save button is enabled 
image

raswonders (commented):
I've tested it and it worked for me. I wasn't able reproduce Zhamilya's dirty state problem. 

Code looks good. I'd maybe consider curbing the comments in InventoryViews, they are becoming quite verbose. I'm not approving yet because we still have suspicion of a bug.

johnsonm325:
I also cannot reproduce @computercamplove 's bug.

computercamplove (approved):

